PR #13, opened 2016-06-03
Support `.tsx`

rozzzly:
The extension `.tsx` is the typescript equivalent to `.jsx`. It should be supported as well! 

Right here I added it as an alias to `.jsx` but it would be better to give  `.tsx` an alternate color from `.jsx`

One thing to note:
`.js` icon is yellow.
`.ts` icon is **blue**.
`.jsx` icon is **blue**.
`.tsx` icon is ???

Perhaps `.jsx` become yellow, `.tsx` becomes blue? Seems more idiomatic to me.

### DIFF
--- a/src/build/supportedExtensions.js
+++ b/src/build/supportedExtensions.js
@@ -57,7 +57,7 @@ exports.extensions = {
     { icon: "python", extensions: ["py"] },
     { icon: "r", extensions: ["r"] },
     { icon: "rails", extensions: [] },
-    { icon: "react", extensions: ["jsx"] },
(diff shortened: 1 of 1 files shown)


robertohuertasm:
Hi @rozzzly I've created a couple of new icons. One yellow and another dark blue. For the moment, I've chosen the dark blue one but maybe we can change it later depending on feedback.

I'm closing this pull request as I've already modified the code to use two different icons. Thanks anyway for the PR.

robertohuertasm:
Finally, I change it to yellow... much better on dark theme! ;D

PR #27, opened 2016-06-16
Implemented auto-reload vscode

code1line:
Now that updates coming from insider build daily, that's probably a good idea.


robertohuertasm:
Good Job! 👍

PR #58, opened 2016-07-08
Add more file extensions to C++, Font, Binary, Shell and Zip

be5invis:
This PR includes:

- `.o, .a, .exe, .lib, .obj, .dll, .pyc` for Binary
- `.hpp, .cc, .cxx` for C++
- `.otf, .eot, .pfa, .pfb, .wfd` for Fonts
- `.htm` for HTML (correction)
- `.zsh, .fish` for Shell
- `.tar, .gz, .bzip2, .bz2, .xz` for Zips

Another problem is that `.m` is used for **both** Objective-C and MATLAB. It is not a simple task to distinguish them. Perhaps we need an option item to specify how to classify them?


fernandosqa:
very nice

PR #61, opened 2016-07-08
Convert icons to grayscale

### LINKED ISSUE #60: Please support grayscale icons
Please support a grayscale version of the icons.

jedmao:
Resolves #60


jedmao:
Oops! Didn't meant to make the PR against master.

jedmao:
If you were willing to make a grayscale branch I'd be happy to submit a PR against that.

robertohuertasm:
@jedmao I'm not sure of how to proceed with this. I don't think having separate branches would be a good idea for the moment. I mean, by doing that, I somehow would assume that I would have to maintain both icon collections. My first idea was to just reference the url of the different icon packages from the readme and then, as a second phase, insert a list in the Command Palette that would automatically change the icons.zip url according to an internal reference. What do you think?

jedmao:
That's fine or you could just add a list of community provided icons to a readme. That way, people can submit PRs against master that just adds an item to the list.

Even better if we could see a preview or sprite sheet of the icons before we download them.

robertohuertasm:
@jedmao I'm ok with providing a list on the readme!! I will create an issue for this 😄

PR #73, opened 2016-07-10
correct coffeescript and include .htpasswd

jens1o:
There was a typo... I fixed it.

This PR also include an update for apache files, because I pushed that previous PR direct in the master branch. Sorry about that :(


robertohuertasm:
@jens1o I think you zip didn't contain a coffeescript logo with the corrected name. I've added it just in case. :+1:

jens1o:
Oh, I thought I oonly had change the name in the .zip File. I think I just uploaded the wrong file...

PR #115, opened 2016-07-29
Fix copy paste failure in changelog

jens1o:


robertohuertasm:
No need for this. Fixed while merging #116

PR #116, opened 2016-07-29
Fix protocol in changelog

jens1o:


robertohuertasm:
Thanks for seeing this! 😉

PR #148, opened 2016-08-06
fix json syntax

jens1o:
***Fixes e0fad191d85e4295705af5444fa0f0adf827f216***

**Changes proposed:**
* [ ] Add
* [ ] Prepare
* [ ] Delete
* [X] Fix

**Things I've done:**
* [X] I've pulled the latest master branch
* [ ] I've run `npm install` to install all the dependenies
* [X] I've checked my code, it is clean.
* [ ] I've run ESLint.
* [X] My pull request is not too big, otherwise I've already squashed it.
* [ ] I've run `npm run build` to build the extension. (If I had done something with the extension.)

fixes the "bad syntax" issue. Silly you, GitHub.


robertohuertasm:
@jens1o I explicitly put it in one line to make it clear where they can find the extensions as this is the format used in the supportedextensions.js file.

jens1o:
But in this way, GitHub makes the background color red. So I unminified it, but I put a link under there, with the exact same line number.

robertohuertasm:
I see, I only thought about Visual Studio Marketplace rendering. There's looking perfect... Why is Github even coloring this red? Maybe if we just ignore json syntax it will work.

jens1o:
Okey, I'll cross out the json specifics.

robertohuertasm:
I've already changed it directly. Thanks!

jens1o:
Okey!

robertohuertasm:
@jens1o I didn't see your last line. It's also referenced in a previous paragraph but I think it's ok to be redundant in this case. If you leave this change alone I will merge it. :grinning:

jens1o:
done in #149

PR #178, opened 2016-08-23
Use the haxe icon for Haxe source files

Gama11:
`.hxml` are just config files for the Haxe compiler, source files use `.hx`.


robertohuertasm:
It will be published in next release.

PR #180, opened 2016-08-24
Use the haxe icon for .hxp files

Gama11:
Sort of a follow-up to #178. Lime, a very popular Haxe framework, has a project format that uses the `.hxp` file ending (most often used for the OpenFL framework that is built on top of it). Files are basically just Haxe source files with a different ending.

See http://www.openfl.org/learn/docs/command-line-tools/project-files/hxp-format/

It guess it would be even nicer if the Lime icon was used (https://github.com/openfl/lime/blob/develop/templates/default/icon.svg). But I'm not sure if you'd like to add that or not? Do you have any rules regarding adding new icons?

### DIFF
--- a/src/build/supportedExtensions.js
+++ b/src/build/supportedExtensions.js
@@ -55,7 +55,7 @@ exports.extensions = {
     { icon: 'haml', extensions: ['haml'] },
     { icon: 'handlebars', extensions: ['hbs', 'handlebars'] },
     { icon: 'haskell', extensions: ['has', 'hs', 'lhs', 'lit', 'gf'] },
-    { icon: 'haxe', extensions: ['hx', 'hxml'] },
+    { icon: 'haxe', extensions: ['hx', 'hxml', 'hxp'] },
     { icon: 'html', extensions: ['htm', 'html'] },
     { icon: 'image', extensions: ['jpeg', 'jpg', 'gif', 'png', 'bmp'] },
     { icon: 'ionic', extensions: ['ionic'], special: 'project' },


robertohuertasm:
@Gama11 Feel free to change the PR in order to include Lime icon and the proper association. If someone doesn't like to link lime icon with .hxp he will be able to associate to normal haxe logo by using file associations.

I'll wait for your new PR in order to do the merge.

Gama11:
What's the process of adding a new icon (couldn't find any docs on it in the Readme)? Does `icons.zip` need to be updated manually?

robertohuertasm:
Yes, sorry for that. You must update the icons.zip manually and add the icon to the icons folder. It must be 32x32 transparent and follow the same naming convention of all the current icons. Take a look at the folder and you will get the idea. You can also use some of the icons as a template for the image. Make it png 24 to avoid white borders and you're done :grin:

Gama11:
What about `replacements.js`? Is it generated?

robertohuertasm:
Yes, this is generated by doing npm run build once you've added the new extension to supportedExtensions.js file. But don't worry, I always run this command before publishing, just in case. :wink:

Gama11:
Yeah, it's just that I'd like to test it locally before making the PR to verify it's working.. Because right now it's not, .hxp just has a blank icon. Probably some stupid mistake on my side.



The changes: https://github.com/Gama11/vscode-icons/commit/b36cedfb4d5c4a529f5dc17f44a9947c22c196de

robertohuertasm:
supportedExtensions is properly set? Be sure to not use .(dots) in the extensions.

Gama11:
See https://github.com/Gama11/vscode-icons/commit/b36cedfb4d5c4a529f5dc17f44a9947c22c196de

PR #193, opened 2016-08-27
add photoshop

jens1o:
***Fixes #191***

**Changes proposed:**
* [X] Add
* [ ] Prepare
* [ ] Delete
* [ ] Fix

**Things I've done:**
* [X] I've pulled the latest master branch
* [ ] I've run `npm install` to install all the dependenies
* [ ] I've checked my code, it is clean.
* [ ] I've run ESLint.
* [X] My pull request fixes an issue, I referenced the issue.
* [X] My pull request is not too big, otherwise I've already squashed it.
* [ ] I've run `npm run build` to build the extension. (If I had done something with the extension.)

add psd files

PR #182, opened 2016-08-24
Add support for .hxp / Lime project files

Gama11:
Followup for #180.

PR #200, opened 2016-08-27
Add the Flash icon for .swf, .swc and .sol

Gama11:
.sol is a local shared object, aka flash cookie

Didn't update icons.zip to avoid merge conflicts with the other PRs.


robertohuertasm:
It's ok but I would rather make the icon smaller. Let's put some space between the border and the icon. Let's standarize it in 2 pixels of margin as per some other icons in the extension. [#193](https://github.com/robertohuertasm/vscode-icons/pull/193)

jens1o:
@robertohuertasm Okey.

Gama11:
The HaxeDevelop icon I added also didn't have any margins I think. Neither did Lime really (#182), but I don't think it's necessary there?

robertohuertasm:
I don't think so, don't worry about them. In fact, your lime icon had more than 2px margin although only vertical. It's not a common rule because most of the icons that we currently have are not equal but just to make the icons feel more or less right with the others that we have. If we're making a square icon full 32px it's going to be seen as bigger. It's more a perceptive issue than numeric. @jens1o @Gama11

Gama11:
I have actually noticed that the HaxeDevelop icon appears bigger than the others, so I'd like to change that one.

Gama11:
Done!

jens1o:
You need to also add them to the icons.zip.

Gama11:
As mentioned previously:

>Didn't update icons.zip to avoid merge conflicts with the other PRs.

jens1o:
Oh, I overlooked this notice. But you could create a new branch, in which you update the zip file.

Gama11:
What would that do?

jens1o:
A branch is a split of your fileset, in which you can change files, because it is a copy of the master branch. When you change a file in there, it does not affect the other branches. You can create for every branch one PR, and if you commit something to this branch, it does not affect others.

jens1o:
You can create a new branch, which does not affect the other files.

Gama11:
I'm well aware of that. I'm actually trying to avoid creating merge conflicts with your PRs, not my own.

jens1o:
I thought it just add or remove them? In a zip, it would add them, don't it?

Gama11:
Hm.. I thought Git might just treat .zip as a binary file. Maybe not!

robertohuertasm:
@Gama11, @jens1o I'm not really sure about this either so I've changed the code and now icons.zip is just a build artifact. Every time anyone runs build task icons.zip file will be created from icons folder. :grin:

jens1o:
Nice thing! :heart:

Gama11:
Awesome!

PR #216, opened 2016-08-28
Support "source" as a src folder

Gama11:


Gama11:
Btw, what is the source icon supposed to represent? A tree? Seems a little strange to me.

robertohuertasm:
Indeed, it's a tree. I didn't know what to use and thought of prime matter...  If you have another idea we can discuss it.

Gama11:
I often see angle brackets (`</>`)  used to represent "code", see [here](https://www.google.de/search?q=code&source=lnms&tbm=isch&sa=X&ved=0ahUKEwilu5zkuefOAhWJliwKHb59B8QQ_AUICCgB&biw=1856&bih=995#tbm=isch&q=code+logo).

image

image

Gama11:
Also here: https://www.iconfinder.com/search/?q=code

robertohuertasm:
I don't identify src folder with code as code is almost everything that you will have in your code. Indeed, dist folder also contains code so... I'm not saying that a tree is the best idea of all but for me it makes more sense an icon that implies the idea of something that is used to build something or to grow something new.

We can create a folder icon with these brackets and link it to "source" and let people choose whatever they feel is more suitable for them using associations.

Gama11:
>I don't identify src folder with code as code is almost everything that you will have in your code.

Huh?

robertohuertasm:
Sorry, there was a typo in my sentence. It's been corrected now.
I mean that the idea of using an icon that implies "code" (as per your own words) is not really the better idea to make a folder feel unique or different from the others because all the other folders will surely contain code as well...

Gama11:
Really? What kind of code would the other folders contain? For tests maybe, but that already has its own icon anyway.

robertohuertasm:
I've bought your argument. I've created the icon and looks ok. I'll publish it soon.

Gama11:
Nice! :+1:

robertohuertasm:
You should be seeing the new icon by doing "icons update".

Gama11:
Oh, didn't realize that worked without an update of the extension.

I like it! Contrast of the brackets against the green background is perhaps a bit low.

robertohuertasm:
Yes. I've added a little bit more contrast. I didn't want to make it super contrasted just subtle but noticeable.

Gama11:
Yeah, that's reasonable. It's a tough balance. :)

Gama11:
Speaking of, I don't think it's really possible to make out what the test icon is supposed to be at 16x16 (had to look at the 32x32 version).

robertohuertasm:
Yes, I know... I was thinking the same... I was leaving it for another moment... but definitely needs some refactoring.

robertohuertasm:
I've changed to red and white. More contrast and clearer icon. What do you think?

Gama11:
I still think it's a little hard to make out, but definitely better than before! :)

PR #224, opened 2016-08-29
Add support for .haxelib

Gama11:
Haxelib has a "newrepo" command that creates a local repo in a folder named ".haxelib", similar to node_modules.

I edited the .psd with GIMP, which seems to work well enough. However, you just made a tweak to the .psd too, and I didn't want to overwrite those changes... so I added a folder2.psd (not sure how to "merge" psd files..). If it's not needed I can just remove it.


robertohuertasm:
@Gama11 don't worry I'll do the psd merge. :+1:

Gama11:
Thanks!

Gama11:
Btw, how does the "special" option work if only one oone of the extensions should use it? There's a `haxelib.json`, which is basically like the `package.json` for NPM, but Haxelib doesn't have its own icon, so I'd like to just use the Haxe one. However, that one is already used for multiple other extensions.

robertohuertasm:
I guess this is similar to this:
```
{ icon: 'bower', extensions: ['bowerrc'] },
{ icon: 'bower', extensions: ['bower'], special: 'json' },
```

You should do this instead:
```
{ icon: 'haxe', extensions: ['hx', 'hxml'] },
{ icon: 'haxe', extensions: ['haxelib' ], special: 'json' },
```

Gama11:
Ok, cool. Feel free to just commit that, I won't get to it until tomorrow. :)

PR #254, opened 2016-09-01
Support license files with an .md extension

Gama11:

### DIFF
--- a/src/build/supportedExtensions.js
+++ b/src/build/supportedExtensions.js
@@ -84,6 +84,7 @@ exports.extensions = {
     { icon: 'log', extensions: ['log'] },
     { icon: 'less', extensions: ['less'] },
     { icon: 'license', extensions: ['license', 'enc'] },
+    { icon: 'license', extensions: ['license.md'], special: 'md' },
     { icon: 'lisp', extensions: ['bil'] },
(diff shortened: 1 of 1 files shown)


Gama11:
Btw, there seem to be a few `extensions` that start with a dot.. For instance the `eslint` ones. Didn't you mention that there shouldn't be a leading dot?

robertohuertasm:
@Gama11if you omit the  dot then you will allow: `eslintrc.yml`, note that no dot is placed at the beginning of the file name. By using the dot you're inforcing the user to, at least, start the filename with a dot. If you don't care about this then by omitting the dot you provide more options to associate files to a specific icon.

Gama11:
Any way to make this case-insensitive in order to support variations like `LICENSE` and `LICENSE.md`?

robertohuertasm:
As far as I know it should be case insensitive already? is it not working for you?

Gama11:
No. Uppercase:



Renamed to lowercase:

Gama11:
No:





Also, I guess different spellings should be considered (license vs licence), but that's a different issue.

robertohuertasm:
Fixed in #264

Gama11:
Awesome, thanks! :)

PR #255, opened 2016-09-01
Use the VSCode icon for tasks.json

Gama11:
There is also a `settings.json`, but I didn't add that because the name might be a little too generic, leading to files using the icon incorrectly... Perhaps it could check that these files (`launch.json`, `tasks.json`, `settings.json`) are within a `.vscode` folder somehow?


robertohuertasm:
@Gama11 we have no access to our siblings or parents in the scope where we inject the code. Anyway, the extension will have to migrate to the official API as Microsoft is thinking of removing the possibility of injecting code anymore by using checksums [sic].

Gama11:
Hm, I can see where they're coming from. The user experience of needing to start with admin rights for each update of the extension isn't great. :) Besides, there might be extension that aren't careful about it and mess up the VSCode install..

PR #381, opened 2016-10-12
Added yarn icon for yarn.lock files

wagerfield:
Added icon and file type matching for `yarn.lock` files.

[Yarn is a new dependency management client developed by Facebook](https://yarnpkg.com/).


robertohuertasm:
That was fast! We have now icons for yarn. I will release it probably tomorrow as today we have already published 3.0.4. Thanks @wagerfield!

PR #443, opened 2016-11-17
Add more webpack file names

schiehll:
***Fixes #[ISSUE]***

**Changes proposed:**
* [x] Add
* [ ] Prepare
* [ ] Delete
* [ ] Fix

**Things I've done:**
* [x] I've pulled the latest master branch.
* [x] I've run `npm install` to install all the dependencies.
* [ ] I've run ESLint.
* [x] I've run `npm run build` to build the extension.

Add some more suffixes to webpack file names when using babel.


schiehll:
Oh sorry about that, I initially thought in open a issue asking for it, but why not a PR. I should've looked further on how to change it hehe

I will make the changes to the right file.

robertohuertasm:
@schiehll no problem, it's not explained anywhere as the build process has change a little bit but esentially you just have to change the `supportedExtensions.js`file and then run `npm run build`. This process will generate the `icons.json`file for you. I guess you have figure it out. 😁 

Thanks for the PR!

schiehll:
Sure! Thank you for creating this awesome extension!

JimiC:
@robertohuertasm That `webpack` extension declaration is getting kinda big. What if we could support of `wildcard` for file extensions? (i.e `webpack.config.*`)

robertohuertasm:
@JimiC, that was possible when we did code patching. In fact, we implemented it but VSCode team worked with us and encouraged us to abandon this practise so I would say this is not really possible at the moment. 
There's an [issue I opened in the VSCode repo that asks for glob support](https://github.com/Microsoft/vscode/issues/12493#issuecomment-249117649).  You can say something there so they see this is something that would be awesome.

PR #446, opened 2016-11-20
Add ability to generate examples from icons.

JimiC:
I created this tool to help me preview how the new SVG icons will look in the editor and thought of sharing it with the rest.

**How it works:**
Read all about it in `README`.

### DIFF
--- a/src/build/supportedFolders.js
+++ b/src/build/supportedFolders.js
@@ -6,7 +6,7 @@ exports.extensions = {
     { icon: 'dist', extensions: ['dist', 'out', 'export', 'build'] },
     { icon: 'docs', extensions: ['docs'] },
     { icon: 'flow', extensions: ['flow'] },
-    { icon: 'git', extensions: ['git', 'github'], dot: true },
+    { icon: 'git', extensions: ['github', 'git'], dot: true },
     { icon: 'haxelib', extensions: ['haxelib'] },
     { icon: 'images', extensions: ['images', 'img'] },
     { icon: 'less', extensions: ['less'] },
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
 test
 dist
+examples
 npm-debug.log
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -1,5 +1,6 @@
 .github/**
 .vscode/**
+examples/**
 images/*.psd
 images/*.ai
 Dockerfile
--- a/README.md
+++ b/README.md
@@ -79,6 +79,37 @@ If you're willing to create an icon just follow this few conventions:
 4. Center the icon, but lower it a pixel.
 5. It must be transparent.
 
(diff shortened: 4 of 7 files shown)


JimiC:
@robertohuertasm Please be so kind to review the code. I don't if my coding style follows yours, but I'm used to writing IIFE (so not to pollute the global namespace).

robertohuertasm:
Hi @JimiC, what's the idea behind this PR? I'm not sure to get it 😞 
About coding style, being this a node module I wouldn't care about polluting the global namespace but it's ok for me if you like using IIFEs 😉

JimiC:
As I wrote in README 

>In order to help you preview how the icon you are contributing will look in the editor, we are providing you with a tool that generates dummy folders or files of the provided icon. 

As for the coding style, I can switch it to match your style, no big deal for me. It's all about consistency, and I have an O.C.D. with that. So say the word and I'll do the switch.

robertohuertasm:
I read the README but I didn't get the idea until now 😓 👍 
It seems a good idea to me 😉
Regarding coding style, I have no big issues with this as long as the code is readable and follows the eslint rules. But if you have such condition 😜 I would say I usually favour function declarations over function expressions and I'm not really into IIFEs unless really needed since working with modules provides such type of code encapsulation, but again, no sweat about this.

JimiC:
@robertohuertasm OK. Hold your horses and don't merge it yet, until I switch the code.

JimiC:
@robertohuertasm I'm done with the PR. Found some time at work to finish this.

src/build/supportedFolders.js
```diff
@@ -6,7 +6,7 @@ exports.extensions = {
     { icon: 'dist', extensions: ['dist', 'out', 'export', 'build'] },
     { icon: 'docs', extensions: ['docs'] },
     { icon: 'flow', extensions: ['flow'] },
-    { icon: 'git', extensions: ['git', 'github'], dot: true },
+    { icon: 'git', extensions: ['github', 'git'], dot: true },
```
JimiC: I switched places because `.git` is getting ignored by `.gitignore`.

robertohuertasm: What do you mean?

robertohuertasm: yes, but what's the relationship between that and the order in the supported folders json?

robertohuertasm: I already accepted the changes, it's just out of curiosity 😉

JimiC: >  The generator only takes the first element from the supported extensions (no meaning to generate all extensions).

robertohuertasm: Sorry, I missed that part. So it's related to the example generator. That makes sense.

JimiC: My fault, I usually fail to fully explain in my writings.

PR #454, opened 2016-11-21
Added VS Code icons to files.

davormas:
Added extensions.json and settings.json as VS Code files.

### DIFF
--- a/icons.json
+++ b/icons.json
@@ -1148,6 +1148,8 @@
     "jsconfig.json": "_f_vscode",
     "tsconfig.json": "_f_vscode",
     ".vscodeignore": "_f_vscode",
+    "extensions.json": "_f_vscode",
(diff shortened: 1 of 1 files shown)


jens1o:
Please do not modify this file. Go ahead into the build directory, there is a file called ```supportedExtensions.js``` Please add it there, thanks 😃

robertohuertasm:
settings.json is not only a vscode related file so I think it shouldn't be changed. 

extensions.json could follow the same rationale, but I'm not as certain.

robertohuertasm:
I would directly close this PR. What do you think @jens1o, @JimiC?

JimiC:
I agree with your reasoning. `settings.json` isn't in any way affiliated with `vscode`.

jens1o:
I agree, too. Closing this.

Issue #392, opened 2016-10-15
Modify the build functionality to accept light icons


robertohuertasm:
Now the project doesn't take care of light themes. I think it would be easy enough to modify the build process in order to add specific icons for light themes.

jens1o:
Hmm. nice comfort, but is it worth it?

robertohuertasm:
I think so. There are some people complaining about that. And if we have the option I will accept PR gladly. Some icons will only require a color change or some border applied. Plus, the official api allows to publish specific icons just for light themes. It wouldn't affect the usual flow that we have now.

jens1o:
Great! I'm agree with you!

JimiC:
I believe we are a step closer into support this. Looking into this as well.

PR #463, opened 2016-11-27
Add support for purescript.

### LINKED ISSUE #431: Icon Request: PureScript
- I'm using Visual Studio Code Version [1.7.1] [Stable]
- I'm using the extension version [4.0.0]
- [x] I've regarded to the [FAQ](https://github.com/robertohuertasm/vscode-icons/blob/master/README.md#faqs) and they did not help.
- My system is: [Windows] [x64]
- [x] I'm sure this issue is not a _duplicate_?
- [x] I want to create an icon request:
  - Type: extension
  - Icon Name: PureScript / PureScript Test/Spec
  - Extensions: .purs, .spec.purs, .test.purs
  - [x] I crossed out "Description", "Steps to reproduce", "Expected behaviour" and "Actual behaviour".

JimiC:
***Fixes #431***

**Changes proposed:**
* [x] Add
* [ ] Prepare
* [ ] Delete
* [ ] Fix

**Things I've done:**
* [x] I've pulled the latest master branch.
* [x] I've run `npm install` to install all the dependencies.
* [x] I've run ESLint.
* [x] I've run `npm run build` to build the extension.


JimiC:
Has the same colors as `ansible`.

jens1o:
not really on a high-contrast monitor...

JimiC:
What you mean?

jens1o:
I think the background of the logo is quite more dark than the Code background color.

JimiC:
So is the [logo](https://github.com/purescript/logo/blob/master/PS_Logo_Final.svg).

jens1o:
Okey.

robertohuertasm:
Hi @JimiC, I think this is a perfect case for the new light support functionality. I personally don't like the idea of having the squared background like this. Is too similar to the vscode background but different enough to notice the change in the tonality. I would remove it or sorround it with a 1px border of a more lighter tone so users could notice the difference. In this case I do think the logo background is important so I would opt for the second option. What do you think?

jens1o:
By the way, who's "Nooknok"? When I try to visit his profile, I see a 404 error page...

JimiC:
As I said to @jens1o the icon has the same background as the `ansible` icon and those colors are the official ones of the logo. On a light theme colors can be swapped. I don't we need to do any extra work for the time being.

JimiC:
@robertohuertasm I love the idea of having a `light` version and with the work we are doing converting all icons to SVG it can be easily done. But first we need to finish the first task.

robertohuertasm:
After having done a few tests and seen the [purescript github page](https://github.com/andyarvanitis/purescript-native/blob/master/README.md), the [official page](http://www.purescript.org/) and what [google has to say](https://www.google.es/search?q=purescript+logo&espv=2&source=lnms&tbm=isch&sa=X&ved=0ahUKEwi_oMmB1snQAhXMxYMKHcn-BLIQ_AUICCgB&biw=1920&bih=935) I would discard the background square. 

I will try to explain my statement:

1. I think that the logo with the background that we have in this PR doesn't respect the margins in the "official" logo. I know this is because we want to better see the icon inside the square but, in my opinion, this doesn't look good, it seems forced. The alternative, being making the icon smaller doesn't work in our favour either, because the icon is barely recognizable.

2. After seeing some of the Purescript resources I think that the icon is the main piece of the logo. They even use the icon on itself without any background. I think this background is just the contrasted version they use but given that VSCode dark theme has a similar background I think we could take advantadge of that. This would allow us to make the icon look bigger and be close to the "official" way. We're not always accurate  or loyal to the original implementation (see markdown, js, ts...)
image

3. @JimiC, you're right about the Ansible icon. First thing about this icon is that the A should be white. I will change it asap. The other thing to remark, in contrast with the Purescript logo is that in Ansible the Circle is a piece of the icon so it makes sense to incorporate it. It's not like that in the case of Purescript.

Finally, about the `light` version, I think it can be implemented no matter the state of the SVG conversion process. What I mean when I talk about a light version is not that we're going to provide a light version for all the icons at once. I'm talking about implementing a way to mark the icons with a light version flag in the `supportedExtensions`file and generate the proper `icons.json`file. There's already some people who is using the light version with our theme extension. In this case, I was thinking that maybe this could be a way to force us to implement this automatic process by the virtue of having this new icon amongst us. The real light icons implementation would come later. But indeed it's not my idea to force anything so never mind!

JimiC:
@robertohuertasm In the SVG version the A is white for the `ansible`logo. When I worked on the `purescript` logo what troubled me was that it's white and on the artboards white background it doesn't show. I CAN make the logo be only the white parts and amke it 30x30 on the 32x32 canvas.
The problem is that it will not show on a `light` them.

So until we have support for `light` theme we can go with this and revise it later.

JimiC:
I suggest you make your priority to implement the support of a `light` version of an icon and let me do the adjustments on the icons.

Now that I'm familiar with the tool I use to create the icons the process can speed up.

robertohuertasm:
@JimiC ok! You've made your point. Let's do it like you say. Once the light version functionality (#392) is ready this will be the first icon to convert 😉

JimiC:
@robertohuertasm And the `ansible` too.

PR #488, opened 2016-12-04
Refactored the build process.

JimiC:
I took the liberty to refactor the functions of the build process.

List of changes:
- Made the functions more generic.
- Added the ability to generate the icons file to any directory, passing it as an argument to the `generate` function.
- Added functions comments.
- Added ability to find the relative path to the icons folder (handy if we decide to move the `icons.json` file to dist).
- Added `npm` script entry for `eslint` (this will help if we adopt a CI service).
- Added some `eslint` rules to get rid of inline rule exceptions.
- Added ability to auto-update the `package.json` file, if the icon theme path changes.
- Added sorting to the folders and files list on generating (now we don't care of the supported icon positioning in the list).

@robertohuertasm Keep what you like. Through it away if you don't like it.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,7 +12,14 @@
         "code": 100
       }
     ],
-    "no-console": 0,
+    "no-console": [
+      1,
+      {
+        "allow": [
+          "error"
+        ]
+      }
+    ],
     "no-restricted-syntax": [
       0,
       "ForInStatement"
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,6 +6,13 @@
     "airbnb-base/legacy"
   ],
   "rules": {
+    "max-len": [
+      2,
+      {
+        "code": 100
+      }
+    ],
+    "no-console": 0,
     "no-restricted-syntax": [
       0,
       "ForInStatement"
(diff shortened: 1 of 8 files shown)


.eslintrc.json
```diff
@@ -6,6 +6,13 @@
     "airbnb-base/legacy"
   ],
   "rules": {
+    "max-len": [
+      2,
+      {
+        "code": 100
+      }
+    ],
+    "no-console": 0,
```
robertohuertasm: Why this?

robertohuertasm: I would suggest to allow console.error only and change all the current `console.log` calls that stand for an error message to `console.error` .

Legit `console.log` calls should keep the eslint comment.

robertohuertasm:
@JimiC I'm going to publish `4.0.5` today. In general I agree with all your contributions. 👍 The only point being in discussion is `no-console` rule set to 0. I'm thinking we should publish these changes in the next release.

.eslintrc.json
JimiC: The logic behind it was that we had inline disables for every use of `console`.

JimiC:
@robertohuertasm If I remove the `no-console` rule then all references to `console` are marked (even those for `console.error`). The only other alternative is to use a 3rd party logger.

robertohuertasm:
Yes, it is. That's why I suggested to allow `console.error`. 😉

JimiC:
@robertohuertasm The PR is again ready for reviewing. I have also updated the change list.

PR #490, opened 2016-12-05
Just making sure that the grammar is right.

Kitanga:
I just went through and made sure that the grammar was OK. Your message was kept intact. By the way, thank you for this extension!!!!

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,9 @@ Bring icons to your VS Code.
 
 As you all know, VSCode finally supports File and Folder icons in an official way. Read this great article by [@chrisdias](https://twitter.com/chrisdias): ["Rebellion. Mutiny. Revolt. Uprising."](https://code.visualstudio.com/blogs/2016/09/08/icon-themes) if you want to learn the history about this plugin and how finally icons landed in VSCode.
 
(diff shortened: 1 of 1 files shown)


robertohuertasm:
@Kitanga, thank you very much for these corrections. It's always a good thing to have a native speaker overviewing what we, non-English natives, write. 👍

Kitanga:
Your welcome! Btw, I'm technically a non-English native, according to the English Native haha.

Anyways, I was wondering what software you all use to create these icons?

robertohuertasm:
We're using Adobe Illustrator and/or Inkscape.

JimiC:
@Kitanga I'm in the process of writing a "How to" for Illustrator and Inkscape. It will be published soon(TM).

*soon(TM): For those who don't get it, it's a play on words, meaning "sooner than never".

jens1o:
@JimiC HERE. Inkscape, perfect. I'm too silly for it. 😆

Kitanga:
Well I've used Inkscape very little.
What about using GIMP?

JimiC:
AFAIK GIMP has no native support and requires a plugin. Not ideal. Correct me if I'm wrong.

PR #501, opened 2016-12-11
Fix declarations of supported file extensions and filenames.

JimiC:
Writing some more tests revealed some bugs we have in `supportedExtensions`.


src/build/supportedExtensions.js
```diff
@@ -96,9 +96,9 @@ exports.extensions = {
     { icon: 'image', extensions: ['jpeg', 'jpg', 'gif', 'png', 'bmp', 'ico'] },
     { icon: 'ionic', extensions: ['ionic.project'], filename: true },
     { icon: 'ionic', extensions: ['ionic.config.json'], filename: true },
-    { icon: 'jade', extensions: ['jade', 'pug', 'tag'] },
```
jens1o: why do you removed 'tag'?

jens1o: Okey.

JimiC:
@robertohuertasm Shouldn't the extensions marked with language support not be present in the `fileNames` or `fileExtensions`?

src/build/supportedExtensions.js
```diff
@@ -5,16 +5,16 @@ exports.extensions = {
...
-    { icon: 'ng2_component_js', extensions: ['.component.js'], svg: true },
-    { icon: 'ng2_directive_ts', extensions: ['.directive.ts'], svg: true },
-    { icon: 'ng2_directive_js', extensions: ['.directive.js'], svg: true },
-    { icon: 'ng2_pipe_ts', extensions: ['.pipe.ts'], svg: true },
-    { icon: 'ng2_pipe_js', extensions: ['.pipe.js'], svg: true },
-    { icon: 'ng2_service_ts', extensions: ['.service.ts'], svg: true },
-    { icon: 'ng2_service_js', extensions: ['.service.js'], svg: true },
-    { icon: 'ng2_module_ts', extensions: ['.module.ts'], svg: true },
-    { icon: 'ng2_module_js', extensions: ['.module.js'], svg: true },
+    { icon: 'ng2_component_ts', extensions: ['.component.ts'], filename: true, svg: true },
```
robertohuertasm: Why the filename property if this is just an extension?

src/build/supportedExtensions.js
robertohuertasm: Extensions that collide should be studied. Some can be solved using languageIds. @JimiC you removed `prolog` language in another PR, which is not right, as languageIds can be added via plugins. In the case of `Prolog` you can find the extension [here](https://marketplace.visualstudio.com/items?itemName=rebornix.prolog) and if you see [their contribution point ](https://github.com/rebornix/vscode-prolog/blob/master/package.json) you'll see that the id of prolog was correct.

JimiC: Right, restoring and commit again.

JimiC: Regarding the `tag` extension, I can't find anything to justify its presence in two places.

src/build/supportedExtensions.js
JimiC: Seems you are correct. The test was complaining... and head was buzzing. Will revise.

src/build/supportedExtensions.js
robertohuertasm: Cool. Yes, I only see a `Riot extension`. Any language extension for Jade/Pug... so let's removed as you originally intended. 👍

src/build/supportedExtensions.js
robertohuertasm: Ok! 👍

JimiC:
Extensions for `ng2` restablished. Will have to check the tests in the other PR regarding those extensions.

robertohuertasm:
@JimiC, regarding the question about language support and extensions, for me it makes sense to have both or at least support both cases, that way if the language support doesn't support (redundance here, sorry) some of the extensions, we don't have to wait or beg the author to implement them in their language extension.

JimiC:
@robertohuertasm Yeah, I figured that out after doing some tests, but forgot to comment on my comment. Sorry.

JimiC:
If this PR is OK you can merge it. I'm going to adjust #502 accordingly.

PR #502, opened 2016-12-11
Add more test cases.

JimiC:
This PR does the following:

- Adds more test cases to ensure the `icons.json` is generated correctly.
- Contains fixes of #501.
- Fixes an error with the `jar` extension declaration.
- Avoids `lodash` usage where possible.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,14 +6,20 @@
     "airbnb-base/legacy"
   ],
   "globals": {
+    "afterEach": true,
+    "beforeEach": true,
+    "afterAll": true,
+    "beforeAll": true,
     "describe": true,
+    "expect": true,
     "it": true,
     "fit": true,
-    "expect": true,
-    "xit": true,
+    "test": true,
     "xdescribe": true,
-    "fdescribe": true
-  }, 
+    "fdescribe": true,
+    "xit": true,
+    "xtest": true
+  },
   "rules": {
     "max-len": [
       2,
--- a/src/build/languages.js
+++ b/src/build/languages.js
@@ -1,4 +1,4 @@
-module.exports = {
+exports.languages = {
   javascript: 'javascript',
   perl: ['perl', 'perl6'],
   prolog: 'prolog'
--- a/src/build/supportedExtensions.js
+++ b/src/build/supportedExtensions.js
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-var languages = require('./languages');
+var languages = require('./languages').languages;
 exports.extensions = {
   supported: [
     { icon: 'actionscript', extensions: ['as'] },
--- a/package.json
+++ b/package.json
@@ -64,9 +64,9 @@
         "prebuild": "npm run lint && npm run test",
         "build": "node ./src/build/build.js",
         "example": "node ./src/build/example.js ",
-        "lint": "node ./node_modules/eslint/bin/eslint src",
+        "lint": "eslint src tests",
         "postinstall": "node ./node_modules/vscode/bin/install",
-        "test": "jest --config ./jest.config.json"
+        "test": "jest -c ./jest.config.json"
     },
     "devDependencies": {
         "eslint": "^3.12.0",
@@ -74,10 +74,10 @@
         "eslint-plugin-import": "^2.2.0",
         "eslint-plugin-jsx-a11y": "^3.0.1",
         "eslint-plugin-react": "^6.6.0",
-        "ncp": "^2.0.0",
+        "jest-cli": "^17.0.3",
         "lodash": "^4.17.2",
-        "vscode": "^1.0.0",
-        "jest-cli": "^17.0.3"
+        "ncp": "^2.0.0",
+        "vscode": "^1.0.0"
     },
     "dependencies": {
         "open": "0.0.5",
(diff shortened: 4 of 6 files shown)


.eslintrc.json
```diff
@@ -6,15 +6,22 @@
...
+    "test": true,
     "xdescribe": true,
-    "fdescribe": true
-  }, 
+    "fdescribe": true,
+    "xit": true,
+    "xtest": true
+  },
   "rules": {
+    "linebreak-style": "off",
```
robertohuertasm: `eol` should be set to `\n`. My bad for not setting it into the `.vscode/settings.json` or an `.editorConfig` file. I'll do it so this won't be an issue.

JimiC:
`Prolog` reestablished.

src/build/languages.js
```diff
@@ -1,5 +1,4 @@
-module.exports = {
+exports.languages = {
```
robertohuertasm: Why this?

src/build/supportedExtensions.js
```diff
@@ -1,20 +1,20 @@
...
-    { icon: 'ng2_component_js', extensions: ['.component.js'], svg: true },
-    { icon: 'ng2_directive_ts', extensions: ['.directive.ts'], svg: true },
-    { icon: 'ng2_directive_js', extensions: ['.directive.js'], svg: true },
-    { icon: 'ng2_pipe_ts', extensions: ['.pipe.ts'], svg: true },
-    { icon: 'ng2_pipe_js', extensions: ['.pipe.js'], svg: true },
-    { icon: 'ng2_service_ts', extensions: ['.service.ts'], svg: true },
-    { icon: 'ng2_service_js', extensions: ['.service.js'], svg: true },
-    { icon: 'ng2_module_ts', extensions: ['.module.ts'], svg: true },
-    { icon: 'ng2_module_js', extensions: ['.module.js'], svg: true },
+    { icon: 'ng2_component_ts', extensions: ['.component.ts'], filename: true, svg: true },
```
robertohuertasm: This was already commented in another PR. Why the filename property?

src/build/supportedExtensions.js
```diff
@@ -144,7 +144,7 @@ exports.extensions = {
     { icon: 'phpunit', extensions: ['phpunit.xml'], filename: true },
     { icon: 'plantuml', extensions: ['pu', 'plantuml', 'iuml', 'puml'] },
     { icon: 'procfile', extensions: ['procfile'], filename: true },
-    { icon: 'prolog', extensions: ['P', 'pro'], languages: [languages.prolog] },
```
robertohuertasm: LanguageIds can be added via plugins. In the case of `Prolog` you can find the extension [here](https://marketplace.visualstudio.com/items?itemName=rebornix.prolog) and if you see [their contribution point ](https://github.com/rebornix/vscode-prolog/blob/master/package.json) you'll see that the id of prolog was correct.

src/build/languages.js
JimiC: http://stackoverflow.com/questions/7137397/module-exports-vs-exports-in-node-js
https://blog.tableflip.io/the-difference-between-module-exports-and-exports/

src/build/supportedExtensions.js
JimiC: As mentioned in the other PR, will revise.

.eslintrc.json
JimiC: My git is set to change the `eol` to `CRLF` when `pull` and `LF` when `push`as I'm on `Windows`. The rule is colliding with that. There isn't going to be an issue if we disable the rule.

src/build/supportedExtensions.js
robertohuertasm: Cool! 👍

src/build/languages.js
robertohuertasm: Sorry @JimiC but I still don't get the rationale here. Could you ellaborate this change?

.eslintrc.json
robertohuertasm: I will upload a `.gitattributes` file so that won't be an issue for you anymore. I'm using `autocrlf = input`.

JimiC: Cool.

JimiC:
OK, I figured it out. It was the damn `dot` that was screwing with me.
I also removed the lint rule about the line breaks.

src/build/languages.js
JimiC: We aren't actually export a `module` per say, but a helper object, which falls into the category of the `exports` case.

robertohuertasm: To be honest, I still don't see the point. As far as I know this is a matter of preference and its use will depend greatly on the way we're going to compose our module. Being this just the only export, I don't see this syntactic change as if it's bringing some benefit to the table.

The examples you reference are composed like that (in the case of `supportedExtensions` and `supportedFolders`) because when I first built the extension I thought that they were going to export more items. Finally they remained like that. No other reason behind it. In fact, I've thought of refactoring these modules for a long time but I've been a little bit careless with style in this project.

The way I see it, if I know my module is going to only export one item I don't like creating extra properties. It's like a `default` for ES5, in this case.

You see I may have also something in my inheritance that leans me towars `bizantine discussions` 😜 

In order to close this, let's keep it this way for the sake of consistency as I myself haven't made it clear in the code.

JimiC: Do you want me to change every `exports` to `module.exports`? I can do that.

JimiC: Yeah. Once again I failed to express myself clearly. 😞

PR #510, opened 2016-12-12
Fix jest related extension.

JimiC:
How did that `dot` got there? Must had a bad weekend...


robertohuertasm:
I didn't see it myself during the review process so... we all must have had a bad weekend! 😉

JimiC:
I'll see to it to add another test case to catch those type of things.

JimiC:
Unfortunately it's not possible to add a test when the case is a filename.
What is possible is to check that the extension doesn't have a leading dot if it's a file extension.

PR #516, opened 2016-12-12
Add support for travis-ci.

### LINKED ISSUE #479: [Proposal] Add DevOps CI service
It would be very helpful, if we could automate the ESLint and build check via a CI (Continuous Integration) service like [`Travis`](https://travis-ci.org/). It can help check the PR before merge.

JimiC:
***Fixes #479***

@robertohuertasm I've noticed that `Travis-CI` is already set and awaiting. If `Travis` doesn't pick up you may need to `Activate` the repository (it has a related button).


JimiC:
Oooooooo. It works!

JimiC:
Yeah \o/.

.travis.yml
```diff
@@ -0,0 +1,3 @@
+language: node_js
+node_js: "6"
+script: npm run build
```
robertohuertasm: Just pointing out the newline warning.
Great work @JimiC !

JimiC: Cause it's not `js`. That's why. (facepalm)

JimiC:
Should we add [YamlLint](https://github.com/adrienverge/yamllint)?

robertohuertasm:
I don't see any problem with that.

JimiC:
Unfortunately [yaml-lint](https://www.npmjs.com/package/yaml-lint) is at a very premature state and doesn't support rules. There is no point in adding a `yaml linter` atm.

PR #518, opened 2016-12-12
Add a new test case for leading dots in file extension declaration.

JimiC:


JimiC:
@robertohuertasm Is there an issue with this PR or you want to do same more testing?

robertohuertasm:
I just want to find some time to review this once I get home 😉

robertohuertasm:
@JimiC I think we don't need this. I mean, `filenames` can have a `dot` at the beginning (`.httpaccess`, `.appveyor.yml`, `.babelrc` ...)

JimiC:
If you take a closer look filename cases are excluded. This test is on for those mark as extensions.

robertohuertasm:
Sorry @JimiC, I was misled by a previous comment from #510 and didn't pay proper attention:

> What is possible is to check that the extension doesn't have a leading dot if it's a file extension.
 
It's ok then :grinning:

tests/iconGenerator.test.js
```diff
@@ -20,6 +20,16 @@ describe('generating icons', function () {
     expect(iconGenerator.getPathToDirName(toDirName, fromDirPath)).toEqual(pathTo);
   });
 
+  it('file extension should not have a leading dot', function () {
+    files.supported
+      .filter(function (file) { return !file.filename; })
```
robertohuertasm: I guess that you've chosen to favor readability instead of performance by double looping over the `files.supported` collection, haven`t you?

JimiC: Yeap. The impact isn't noticeable.

JimiC:
# Glossary:
`filename`: A string that classifies the entire name as an extension.
`extension`: A string that classifies as an file extension.
`language`: A string that classifies as a `languageIds` supported extension.

Hope this helps us speaking a common `domain` language.

tests/iconGenerator.test.js
```diff
@@ -144,7 +154,7 @@ describe('generating icons', function () {
       var fileExtensions = iconGenerator.buildJsonStructure().files.names.fileExtensions;
 
       files.supported
-        .filter(function (file) { return !file.filename && !file.languages; })
```
JimiC: This removal is due to the decision to include the `language` supported extension in the test.
At first I excluded them but after you clarified it I forgot to change it back.

robertohuertasm: If everything is fine I'll proceed with the merge then.

JimiC: I just pointed this out in case you were wondering.
I'm a GO.

PR #521, opened 2016-12-13
Add support for "json" via languageIds.

### LINKED ISSUE #476: Use languageIds to provide generic icons for non-specifically-listed extensions
Instead of just listing all the extensions explicitly, we could list out `languageIds` so that, for example, all files associated with `json` get the generic `_f_node` icon unless a more specific icon is listed.

JimiC:
***Fixes #476***


src/build/languages.js
```diff
@@ -1,5 +1,6 @@
 exports.languages = {
   javascript: 'javascript',
   perl: ['perl', 'perl6'],
-  prolog: 'prolog'
+  prolog: 'prolog',
+  json: 'json'
```
JimiC: This should have been sorted. My bad. Will do in the next PR that adds a `language` support.

PR #530, opened 2016-12-13
Add support for light icon version.

### LINKED ISSUE #392: Modify the build functionality to accept light icons

JimiC:
***Fixes #392***

**Changes proposed:**
* [x] Add
* [ ] Prepare
* [ ] Delete
* [ ] Fix

**Things I've done:**

### DIFF
--- a/src/build/iconGenerator.js
+++ b/src/build/iconGenerator.js
@@ -36,12 +36,19 @@ function buildJsonStructure(iconsFolderBasePath) {
       .reduce(function (old, current) {
         var defs = old.defs;
         var names = old.names;
+        var light = old.light;
         var icon = current.icon;
+        var hasLightVersion = current.light;
         var iconFolderType = 'folder_type_' + icon;
+        var iconFolderLightType = 'folder_type_light_' + icon;
         var folderPath = iconsFolderBasePath + iconFolderType;
+        var folderLightPath = iconsFolderBasePath + iconFolderLightType;
         var openFolderPath = folderPath + '_opened';
+        var openFolderLightPath = openFolderPath + '_opened';
         var iconFolderDefinition = '_fd_' + icon;
+        var iconFolderLightDefinition = '_fd_light_' + icon;
         var iconOpenFolderDefinition = iconFolderDefinition + '_open';
+        var iconOpenFolderLightDefinition = iconFolderLightDefinition + '_open';
         var iconFileExtension = current.svg ? '.svg' : '.png';
 
         defs[iconFolderDefinition] = {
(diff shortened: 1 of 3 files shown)


src/build/iconGenerator.js
```diff
@@ -113,9 +163,30 @@ function buildJsonStructure(iconsFolderBasePath) {
...
+  // dark theme
+  defaultSchema.iconDefinitions._file
+    .iconPath = iconsFolderBasePath + 'file.svg';
+  defaultSchema.iconDefinitions._folder
+    .iconPath = iconsFolderBasePath + 'folder.svg';
+  defaultSchema.iconDefinitions._folder_open
+    .iconPath = iconsFolderBasePath + 'folder_opened.svg';
+
+  // light theme
+  // default file and folder related icon paths if not set,
```
jens1o: Use` /*[...]*/` instead for this long comment.

tests/iconGenerator.test.js
```diff
@@ -179,10 +352,25 @@ describe('generating icons', function () {
...
+            expect(fileNames[extension]).toEqual(definition);
+          });
+        });
+    });
+
   it('ensures each supported file extension that is supported by language ids ' +
     'has a language id referencing its definiton',
     function () {
-      var languageIds = iconGenerator.buildJsonStructure().files.names.languageIds;
+      var languageIds = iconGenerator.buildJsonStructure().files.languageIds;
```
jens1o: why this?

JimiC: It has been changed in line 153.

src/build/iconGenerator.js
JimiC: I'm used in using single-line comment style inside functions/methods even if they are multi-lined.
I only use comment-block styling if they are file header (usually license) or above functions/methods as descreptive comments.
If @robertohuertasm wished to follow the comment-block style inside functions I have no problem switching to that.

JimiC: For more references see:
http://www.hongkiat.com/blog/source-code-comment-styling-tips/
http://javascript.crockford.com/code.html

robertohuertasm: I'm ok with this style. VSCode itself uses this kind of style when you select several lines so no problem with it. Is it ok for you @jens1o ?

jens1o: Yeah, it's ok. I think it's the basic code style to use comment blocks.

tests/iconGenerator.test.js
jens1o: and why?

jens1o: Ah, okey.

src/build/iconGenerator.js
```diff
@@ -36,12 +36,19 @@ function buildJsonStructure(iconsFolderBasePath) {
...
         var names = old.names;
+        var light = old.light;
         var icon = current.icon;
+        var hasLightVersion = current.light;
         var iconFolderType = 'folder_type_' + icon;
+        var iconFolderLightType = 'folder_type_light_' + icon;
         var folderPath = iconsFolderBasePath + iconFolderType;
+        var folderLightPath = iconsFolderBasePath + iconFolderLightType;
         var openFolderPath = folderPath + '_opened';
+        var openFolderLightPath = openFolderPath + '_opened';
```
JimiC: Damn. There is a bug here.

JimiC:
@robertohuertasm Do you think it would be wise to add a case in code where if the `file_light`, `folder_light` and `folder_light_open` get assigned, the code populates the related properties with the dark theme definition when a `light` version is not provided?

src/build/iconGenerator.js
JimiC: Fixed.

robertohuertasm:
@JimiC, sorry but I'm not really sure to understand what you say. You first say that the light properties get assigned but then you later say `when the light version is not provided`?

I guess you're asking if we should provide light versions no matter if they're assigned or not, is that right?

JimiC:
> ___if___ the file_light, folder_light and folder_light_open get assigned

Emphases on `if`.
An with 
> when the light version is not provided

I mean for the other icons other than the default

I think I'll implement this guard just in case we decide to have other default icons for `light`. With the current implementation if we do so it will break the `light` icons.

robertohuertasm:
Better to discuss this via Slack or something because I'm not following you on this.

PR #525, opened 2016-12-13
Resolves the linebreak-style issue (for real this time).

JimiC:
Finally I figured this out. The previous commit/fix didn't fixed the issue for me.

The sorting in the `supportedExtensions` file although trivial has been done also for testing purposes.

PR #531, opened 2016-12-13
Fixes eol

robertohuertasm:
Related to #525 

@JimiC, let me know if this works for you.


JimiC:
According to Git, images are treated as binary by default.

`*.* text eol=lf` is equal to `* text eol=lf`

JimiC:
Fetching the changes didn't created any issues with the line-breaks for me.

robertohuertasm:
I was trying things out. Grad it also works for you. We can put it nicely in another PR.

JimiC:
So `.gitattributes` could be modified to:

```
###############################################################################
# Set default behavior to automatically normalize line endings.
###############################################################################
* text eol=lf

###############################################################################
# Denote all files that are truly binary and should not be modified.
###############################################################################
*.psd binary
*.ai binary
```

PR #532, opened 2016-12-14
Improve language ids support.

JimiC:
Makes the use of the extension declaration redundant and adds support for the example generator.


JimiC:
We also need to explore what other extensions are supported (besides the default) for those with `languageIds` so to remove those too.

src/build/supportedExtensions.js
```diff
@@ -48,10 +47,11 @@ exports.extensions = {
     { icon: 'config', extensions: ['makefile', '.env.example'], filename: true },
     { icon: 'compass', extensions: [] },
     { icon: 'composer', extensions: ['composer.json', 'composer.lock'], filename: true },
-    { icon: 'csharp', extensions: ['cs', 'csx'], languages: [languages.csharp] },
+    { icon: 'cpp', extensions: ['hpp', 'cc', 'cxx'], languages: [languages.cpp] },
+    { icon: 'csharp', extensions: ['csx'], languages: [languages.csharp] },
     { icon: 'cshtml', extensions: ['cshtml'] },
     { icon: 'csproj', extensions: ['csproj'] },
-    { icon: 'css', extensions: ['css'], languages: [languages.css] },
+    { icon: 'css', extensions: [], languages: [languages.css] },
```
jens1o: maybe we don't need to have the extensions array defined in them?

JimiC: It stays there for 2 reasons:
- in case there is another extension that has the same icon.
- schema consistency.

PR #541, opened 2016-12-17
png file added to binaries. linux complains

robertohuertasm:
Linux was recreating png files. The binary declaration in `.gitattributes` did the trick.


JimiC:
Same here. Looks like `.gitattributes` had to exclusively declare handling for the `.png`. Thanks for the fix.

Issue #537, opened 2016-12-15
Huge Javascript block icon


brenoc:
I just updated the extension and now the JavaScript icons changed to this huge block of solid bright yellow.
![screen shot 2016-12-15 at 10 42 05 am](https://cloud.githubusercontent.com/assets/284515/21224404/2c131548-c2b3-11e6-9ae1-7409a009fa08.png)

Previously was a much simple and clean one:
![](https://github.com/robertohuertasm/vscode-icons/blob/440e993372e96bd9b6717af8f306c95881c2ba24/icons/file_type_js@2x.png?raw=true)

I think this was introduced 3 days ago by this PR: #520

@robertohuertasm previously said in this issue (#354) that this icon wouldn't happen and I totally agree with him. This icon catches to much attention in the editor.

What made that change happen? I might use another set of icons, because this icon is very intrusive.

JimiC:
Duplicate of #514.

robertohuertasm:
@brenoc we're releasing the old one again during the week.

Issue #514, opened 2016-12-12
Return old nice JS icon!


VitalySenko:
Please, i very like your extension but new JS icon you made (latest version) make my eyes a little bleeding.
Give us back old good intelligent JS icon please :-)
P.S. I have light theme.
![image](https://cloud.githubusercontent.com/assets/12730311/21098935/b85b6796-c07c-11e6-9bd4-83e1199bd8ff.png)

robertohuertasm:
Here we go again! #508, #339 & #354.

Thanks @VitalySenko for your opinion. We'll have to start a debate on this. What I know is that at some point any decision me make is not going to please everybody. Let's see what people says in this thread.

Anyway, I'm starting to work in a new functionality that will allow more customization options to the extension. It's just raw material for the moment but I expect to have it soon. By now, just a PoC.

JimiC:
Been there, done that. It's a non ending story. Would a poll do the trick?

solomon23:
Using the dark theme it's the brightest thing on my screen now.  Large blocks of yellow for js files.  Maybe supporting themes would be a good way to solve it ? Right now I'll probably have to turn it off.

robertohuertasm:
@solomon23 what do you mean with `supporting themes`? VSCode api is what it is and doesn't really allow any kind of customization by the user. Their solution is to create an alternative theme, but there are already other themes in the market that provide slightly different icons (a lot of them are feeding from our repo).

I was reluctant to change JS and TS icons at the beginning but now I think I like how they look like. I intend to keep this thread open and see which are the opinions of users, although I guess this will be biased to the `don't like` feeling.

I've asked among my fellow colleagues and they all seem to like it but everything is susceptible of being discussed.

solomon23:
@robertohuertasm apologies, i can't really speak intelligently about solutions since i'm not really familiar with it.  The new yellow JS icon is garish enough that I had to turn it off since it's too distracting to my workspace.  Luckily for me there are other icon choices out there for me, so I'll just use one of those.  Thanks !

robertohuertasm:
@solomon23 no worries. I understand that pleasing everybody is an impossible mission. Anyway, I will try a new set of icons playing with inversion of colors (that was the idea behind the old icons) and hopefully will find a way for the user to decide which one to use (although I can't promise you anything).

JohhnyP:
[...]

PR #549, opened 2016-12-18
reduced a little ts and js icons

robertohuertasm:
They look a little bit bigger than I expected.


tmueller:
Hi,

is it possible to bring back the orgiginal icons of js and ts, that were introduced in #509? They looked so much better than the current ones.

Thanks,
Thomas

robertohuertasm:
Hi @tmueller, some people have been complaining of the official icons. They were claiming they were to much obstrusive so we have decided to make a rollback for the moment while we're preparing a user customization tool. We're going to deliver it in phases so it will be improving. First version will be have some presets so you can decide if you want to use the official ones or not, among other things. See #537, #514

tmueller:
Thanks for your quick reply. It's a pity other users found them to obstrusive. Boils down to personal likes I guess. Looking forward to the costumization tool.

robertohuertasm:
I've have a very rough version. I guess I will be able to provide something during the week. 😉

infinnie:
I want the new icons back…

JimiC:
@robertohuertasm FACEPALM

PR #563, opened 2016-12-21
Add support for more languageIds.

JimiC:
https://code.visualstudio.com/docs/languages/identifiers


JimiC:
The `diff` icon is not ideal. I'll fix it when I get on my home workstation.

JimiC:
@jens1o Cleanup. Those were duplicate icons. If you look closer they don't have the `@2x` suffix.

JimiC:
Those `extensions` that I have removed in `supportedExtensions` have been checked with each `language` extension in `vscode` that they are supported by the `language` extension itself.

PR #577, opened 2016-12-24
ng2 icons restored

robertohuertasm:
Latest svg icons batch replaced these icons.


JimiC:
Did so? I'm sorry. Was unintentional. Obviously I forgot to update those icons in Illustrator. Starting to lose count :tongue:

robertohuertasm:
No problem @JimiC. 😉 I fully understand it.

JimiC:
This is why we need a second and third pair of eyes. :smile:

PR #592, opened 2016-12-31
Remove redundant 'outFiles' declaration for test debugging.

JimiC:

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,10 +7,14 @@
             "type": "extensionHost",
             "request": "launch",
             "runtimeExecutable": "${execPath}",
-            "args": ["--extensionDevelopmentPath=${workspaceRoot}" ],
+            "args": [
+                "--extensionDevelopmentPath=${workspaceRoot}"
+            ],
             "stopOnEntry": false,
             "sourceMaps": true,
-            "outFiles": [ "${workspaceRoot}/out/src/**/*.js" ],
+            "outFiles": [
+                "${workspaceRoot}/out/src/**/*.js"
+            ],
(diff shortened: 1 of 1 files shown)


robertohuertasm:
I'm reverting this as there's no redundancy at all. I missed to see it when accepted the PR.

JimiC:
Can you explain to me why the test debugging needs reference to src folder?

JimiC:
You shouldn't be doing this. If the `src` code of the `build` process is wrong, you should be debugging it with another `Launch` config.
`Launch Tests` should only be for the tests.

Just my opinion.

jens1o:
I agree with @JimiC

robertohuertasm:
Sorry but I don't agree. What's wrong with it? Don't you step into your code in your C# unit tests out of the box? If you're referencing code from the `src` it's licit to step in if you want. Is not that the case o a test, to test some external functionality? What's the use of the debugger if I cannot step into any function but the ones generated in my tests as stubs, mocks or whatever?

JimiC:
@robertohuertasm The thing is that currently we have 3 major sections.
1. The extension code.
2. The icons json build code.
3. The tests code.

Now, ideally the tests shouldn't call into the build code directly but, they should `mock` the functionality of the build process. As I'm not so familiar with `mocking` in `JS` I took the easy way when I wrote the test cases.

Never the less, for me it's no big deal if you want to keep them this way. Works for me.

robertohuertasm:
@JimiC, if you are testing some `build code` it's normal to call your `build code` and mock only whatever  you're not testing. **Tests can call anything they want to test**, be it something in the `build code` or any other piece of code that we have.

PR #594, opened 2017-01-01
Convert PNG to SVG (part 4)

JimiC:
**Note:** `ng` icons are the correct ones (double and triple checked), just have been exported with optimization from Illustrator.


JimiC:
PR rebased.

Issue #456, opened 2016-11-23
Folder Icons Request


Kainax:
- Type: extension
      - File Name: .conf (same icon as for .config)

      - Type: folder
      - Folder Name: php
      - Folder Name: js
      - Folder Name: tools
      - Folder Name: assets
      - Folder Name: tmp (+ "temp" folder)
      - Folder Name: .ebextensions          (aws elastic beanstalk default folder starting with dot)
      - Folder Name: .elasticbeanstalk      (aws elastic beanstalk default folder, starting with dot)

      Amazon AWS services icons zip:
https://media.amazonwebservices.com/AWS-Design/Arch-Center/16.2.22_Update/AWS_Simple_Icons_EPS-SVG_v16.2.22.zip

PS - Thanks for the extension, you are doing great!

      EDIT:
      - Folder Name: resources (same icon as for assets folder)

Kainax:
Hi, I think it should be labeled both "file icons" and "folder icons". Thanks.

jens1o:
Surely ;) But I'm busy... Is there a logo for aws in svg? Since we're upgrading to svg, it would be nice if we don't have to convert all stuff.

robertohuertasm:
@jens1o, @Kainax has attached a zip with lots of icons from Amazon AWS in png and svg format, I guess as a suggestion to use. We'll evaluate this and see how many of them we'll eventually implement. 😉

Kainax:
Yes, as it says in a filename of the provided link  :) However, it is more appropriate if folder's icon will be not AWS, but Elastic Beanstalk's application icon. You can get .svg version of it here:

https://upload.wikimedia.org/wikipedia/commons/0/03/AWS_Simple_Icons_Deployment_Management_AWS_Elastic_BeanStalk_Applicaton.svg

Speaking of AWS icon (orange cloud), it would be nice to have it over ".aws" and "aws" folders. You can find aws cloud .svg icon here:

https://upload.wikimedia.org/wikipedia/commons/5/5c/AWS_Simple_Icons_AWS_Cloud.svg

This four folders:
.aws
aws
.ebextensions
.elasticbeanstalk

are probably most popular among AWS users.

It would be totally perfect to draw them on top of default folder icon though. Cheers!

JimiC:
@robertohuertasm @jens1o @Kainax I'll have a look at them on the weekend.

jens1o:
Thank you @JimiC 😉

JimiC:
[...]

PR #595, opened 2017-01-01
[Icon Request] Support .MIN.MAP and folder js

erroledge87:
**Changes proposed:**
* [x] Add

#456 Folder name: js via icon script
Because jQuery source maps is:  jquery-x.xx.xx.min.map so I add .min.map via icon jsmap


JimiC:
The correct way should had been `min.js.map` and not `min.map`, but this is `jQuery`. I'm suspicious that this may cause issues with minified `css` source maps although their extension is `min.css.map`.

JimiC:
I have decided to close this PR as we are going to support this via a new `map` icon.

untitled

Issue #212, opened 2016-08-28
Setting `hideCustomFolderIcons` to `true` has no effect.


jsynowiec:
**Checklist:**
- [x] I'm using Visual Studio Code Version [1.4.0] [Stable]
- [x] I'm using the extension version [1.8.2]
- [x] I've regarded to the [FAQ](https://github.com/robertohuertasm/vscode-icons/blob/master/README.md#faqs) and they did not help.
- [x] My system is: OS X
- [x] I'm sure this issue is not a *duplicate*?
- [x] I've forced Icons Update.
- [x] I crossed out "Description", "Steps to reproduce", "Expected behaviour" and "Actual behaviour".

**Description:**
Setting `hideCustomFolderIcons` to `true` has no effect.

**Steps to reproduce:**
1. Update extension to 1.8.2
2. Add `"vscode. hideCustomFolderIcons": true` to vs code settings
3. Restart vscode

_Expected behaviour:_
All special folders have to basic folder icon.

_Actual behaviour:_
All special folders still have to custom icons.

As a workaround I'm using `useFileAssociations` and `associations` to hide custom icons for all special folders like src, test, node_modules, etc.

robertohuertasm:
Sorry but this setting won't work for the moment until 1.8.2 will be released tomorrow. You have to use the associations workaround as stated in the readme as an alternative solution. I'm sorry for the inconveniences.

PR #596, opened 2017-01-02
User Customization

### LINKED ISSUE #432: Blue VScode Folder Icons! Thoughts?
Hey all! I'm new to this so excuse me if I miss something but I was suggest to post here to get everyone's thoughts so I'm glad to do it!

I know most love the yellow color for folder icons but I LOVE the bright blue that we see on top of the .vscode folders. I suppose that changing them all to blue would look all the same but maybe if they were all blue, the .vscode ones would stand out because they have the VS on top of it. Or perhaps a light, bright green for the .vscode would look great against the blue contrast? Idk, what does anyone else think? I'm sharing a screenshot just to show the blue that I'm talking about. Thanks all!

https://1drv.ms/i/s!AlnGeQSYLUVaoo1gWeVkgDfnq19OIQ

**Checklist:**

- [x] I want to create an icon request(if not, remove these lines below):
      Blue/Green Icons?!

### LINKED ISSUE #508: [Proposal] Switch to official logo where applicable
Would like to propose to adopt official logos when there is one.

### LINKED ISSUE #514: Return old nice JS icon!

### LINKED ISSUE #538: Angular Icon Matching Is Too Broad
The new Angular icons are nice, but the matching pattern is too hit and miss that I think there should be an option to disable these. I have a React Native project in which I end all my React Component files with _.component.js_, so they are being marked as Angular files.

Additionally, in my project at work, where we use Angular 1, the controllers are marked as _.controller.js_ and don't get the Angular icon. The Angular icons are a good idea, but it's confusing right now seeing them where they shouldn't be and missing where they should be.

### LINKED ISSUE #581: Assing icon to dot folders with specific names
Often people name specific purpose folders using leading dot.

```
.scripts
.webpack
.tests
```
Just to make them be on top of the list, now only `scripts`, `webpack` etc without dot got icon.

### LINKED ISSUE #598: Option for hide folder icons!
As title describe!

robertohuertasm:
***Fixes #514, fixes #432, fixes #508, fixes #538, fixes #581, fixes #598, references #212***

**Changes proposed:**
* [x] Add
* [ ] Prepare
* [ ] Delete
* [x] Fix

**Things I've done:**

This introduces user customizations. I've also made changes to the project structure and refactored some parts of the code like iconGenerator tests and others.

The Docs are still a draft version but we could go with that for the moment.


README.md
```diff
@@ -88,12 +82,107 @@ Go to **File > Preferences > File Icon Theme > VSCode Icons**.
...
+```
+
+**Custom Icons**
+Unfortunately, `VSCode` doesn't support absolute paths when generating the icons css. I'm willing to raise an issue in their repo and submit a PR but in the meantime we've come up with a simple solution.
+
+The idea is that you have to look for your `User folder`:
+
+- Windows: `C:\Users\<your_user>\.vscode\extensions`
+- Linux: `/home/<your_user>/.vscode/extensions`
+- Mac: `/Users/roberto/.vscode/extensions/`
```
JimiC: Consistency: Change to 
>Mac: `/Users/<your_user>/.vscode/extensions/`

README.md
```diff
@@ -88,12 +82,107 @@ Go to **File > Preferences > File Icon Theme > VSCode Icons**.
...
+
+and then create a folder there named like this: `robertohuertasm.vscode-icons.custom-icons`. Then, there you can put all your custom icons but, again, they have to follow `vscode-icons` naming conventions:
+
+- Files: `file_type_<value_of_icon_property>@2x.svg`
```
JimiC: We should add a list with the supported icon file extensions (i.e. `svg`, `png` etc).

README.md
```diff
@@ -88,12 +82,107 @@ Go to **File > Preferences > File Icon Theme > VSCode Icons**.
...
+// remove the older icon functionality while extends will keep older settings by 
+// putting yours on top.
+"vsicons.associations.files": [
+  { "icon": "myJs", "extensions": ["js", "custom.js"], "format": "svg", "overrides": "js" }
+]
+
+```
+
+**Custom Icons**
+Unfortunately, `VSCode` doesn't support absolute paths when generating the icons css. I'm willing to raise an issue in their repo and submit a PR but in the meantime we've come up with a simple solution.
```
JimiC: css inside ``.

src/icon-manifest/iconGenerator.ts
```diff
@@ -0,0 +1,356 @@
...
+
+export class IconGenerator implements IIconGenerator {
+  private settingsManager: ISettingsManager;
+  private iconsFolderPath: string;
+  private manifestFolderPath: string;
+
+  constructor(private vscode: IVSCode) {
+    this.settingsManager = new SettingsManager(vscode);
+    // relative to this file
+    this.iconsFolderPath = path.join(__dirname, '../../../icons');
```
JimiC: `path.join` has a serious side effect. The path separator produced is causing issues on different OS platforms.
Either override the function to produce `Unix` style path separator in `Windows` or remove it completely.

src/commands/index.ts
```diff
@@ -0,0 +1,95 @@
+import * as vscode from 'vscode';
+import { getConfig } from '../utils/vscode-extensions';
```
JimiC: It's common practice to always start the path with the current director a.k.a `./../utils/vscode-extensions`.

This means that this has to be changes everywhere were we have paths.

src/icon-manifest/supportedFolders.ts
```diff
@@ -1,33 +1,31 @@
 /* tslint:disable max-line-length */
-
 import { IExtensionCollection, IFolderExtension, FileFormat } from '../models/IExtension';
 
 export const extensions: IExtensionCollection<IFolderExtension> = {
   supported: [
-    { icon: 'aws', extensions: ['aws'], format: FileFormat.svg },
-    { icon: 'aws', extensions: ['aws'], dot: true, format: FileFormat.svg },
+    { icon: 'aws', extensions: ['aws', '.aws'], format: FileFormat.svg },
```
JimiC: I noticed that you are proposing the `format: "svg"` style for custom icons. Should this also be applied to the default extensions?

src/models/IExtension.ts
```diff
@@ -1,25 +1,26 @@
 import { ILanguage } from './ILanguage';
 
-interface IExtension {
-  icon: string;
-  extensions: string[];
-  format: FileFormat;
-  alternate?: boolean;
-  light?: boolean;
+export interface IExtension {
```
JimiC: Common practice is to have every public `class`, `interface` on separate files.

We can then add an `index.ts` file inside the folder and `export` all `interfaces` from inside this file, gaining the advantage to reference only the folder path.

src/models/IIconSchema.ts
```diff
@@ -0,0 +1,27 @@
+export interface IIconPath {
```
JimiC: Same as above review comment about separate `interface` files.

src/settings/index.ts
```diff
@@ -0,0 +1,95 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import * as semver from 'semver';
+import { vscodePath as getAppPath } from './vscodePath';
+import { version as extVersion, iconSuffix, customIconFolderName } from './extSettings';
+import { ISettings } from '../models/ISettings';
+import { IState } from '../models/IState';
+import { IVSCode } from '../models/IVSCode';
+
+export interface IExtensionStatus {
```
JimiC: Same as review comment about separate `interface` files.

src/utils/index.ts
```diff
@@ -0,0 +1,10 @@
+import { IVSCode } from '../models/IVSCode';
+
+export const vscode: IVSCode = {
+  env: { appName: 'Code' },
+  version: '1000.0.0',
```
JimiC: `1000.0.0` ? What's the purpose?

test/iconGenerator.test.ts
```diff
@@ -2,34 +2,26 @@
 /* tslint:disable only-arrow-functions */
```
JimiC: I tested this locally and I haven't had any issues converting to arrow functions. `Mocha` worked just fine.Maybe it's worth re-evaluating it?

test/merger.test.ts
```diff
@@ -0,0 +1,366 @@
+// tslint:disable only-arrow-functions
```
JimiC: Same as review comment about `arrow-functions`.

test/support/supportedExtensions.ts
```diff
@@ -0,0 +1,277 @@
+/* tslint:disable max-line-length */
```
JimiC: Why duplicate file for test?

test/support/supportedFolders.ts
```diff
@@ -0,0 +1,30 @@
+import { IExtensionCollection, IFolderExtension, FileFormat } from '../../src/models/IExtension';
```
JimiC: Why duplicate file for test?

README.md
robertohuertasm: Added a link to the list of supported extensions

src/commands/index.ts
robertohuertasm: What's the benefit of that? I've never heard of it (this not implies it's not true what you say :wink:). Even `vscode` and `typescript` teams use the same syntax that I'm using here.

src/icon-manifest/supportedFolders.ts
robertohuertasm: I used `svg` deliberately, yes, but I've added a link to the supported extensions (as per your suggestion).

I'm not sure what do you mean by default extensions here...

src/models/IExtension.ts
robertohuertasm: I'm aware of the use of barrels but I chosed simplicity in front of multiple files management when developing this. They're grouped by functionality.

Anyway, I'm ok with changing this to barrels. We can use folders as a way to group different `interfaces`.

test/iconGenerator.test.ts
robertohuertasm: I also like `arrow functions` but I will rather stick to [Mocha's recommendations](https://mochajs.org/#arrow-functions).

test/support/supportedExtensions.ts
robertohuertasm: This is just for isolation purposes. I don't want to depend on an external file that will eventually change its contents. I need the tests to be controlled that's why I duplicated the files.

test/support/supportedFolders.ts
robertohuertasm: Same answer given for `supportedFiles`

src/utils/index.ts
robertohuertasm: This is a mock for `vscode` module. Unfortunately, `vscode` methods can only be accessed while using Visual Studio Code (i.e. running tests through vscode itself). 

When we build the shipped `icons.json` by using `npm run build` we're not running into the Visual Studio Code process but node instead and thus, the process will fail to find `vscode` module.

The same happens when we run `mocha` directly in our CI environment. 

That's why I introduced the `IVSCode` interface to inject `vscode` module as a dependency in modules that do not need Visual Studio Code context.

Ideally, all the tests and processes should be executed in a Visual Studio Code context by running a VSCode process and run all through it but when dealing with node process you have to abstract `vscode` module away and inject it properly. This is also nice for testing purposes.

Note that there are some places that the injection of the `vscode` module is useless. That is what I commented you in Slack about `css` not supporting absolute paths in VSCode. I hope to have the time of writing a PR for this and send it to their Github repo. If it's not accepted then we'll remove this as it's really not needed at the moment but I'd rather to leave the code as the current solution is marked as a `temporary hack`.

src/icon-manifest/iconGenerator.ts
robertohuertasm: It seems to not affect the functionality in Unix but I'll rework it to return `Unix` style paths.

robertohuertasm:
@jens1o what do you mean by that?

README.md
```diff
@@ -88,12 +82,103 @@ Go to **File > Preferences > File Icon Theme > VSCode Icons**.
...
+- `Restore Default Icon Manifest`: This command will revert any changes you may have applied and restore the extension to its default state.
+
+But before you can even use them you will have to go to your `settings` and make your magic. The settings will provide you 2 different `array items`:
+
+- `vsicons.associations.files`: [IFileExtension[]](https://github.com/robertohuertasm/vscode-icons/blob/master/src/models/IExtension.ts#L18)
+- `vsicons.associations.folders`: [IFolderExtension[]](https://github.com/robertohuertasm/vscode-icons/blob/master/src/models/IExtension.ts#L23)
+
+Each item of the array represents a file or a folder icon. The functionality is the same for `files` and `folders`.
+
+Note that will important to know what are the currently [supported file extensions / icons](https://github.com/robertohuertasm/vscode-icons/blob/master/src/icon-manifest/supportedExtensions.ts) and [supported folder extensions / icons](https://github.com/robertohuertasm/vscode-icons/blob/master/src/icon-manifest/supportedFolders.ts)
```
JimiC: Still there is a syntax error.

> Note that it's important to know

README.md
JimiC: I was referring to the `FileFormat enum`.

src/commands/index.ts
JimiC: No benefit, just common code styling guideline.

src/icon-manifest/supportedFolders.ts
JimiC: We use the `enum` to reference the format, but you suggest the user to use a string directly.

robertohuertasm: Ah, yes. That's because they can only use simple types in vscode settings. That's why I changed format field to union type.

README.md
robertohuertasm: Yes, I meant that. Icon extensions. Semantic ambiguity...

src/icon-manifest/supportedFolders.ts
JimiC: So the question remains. Should we also switch from `enum` to "union"?

test/iconGenerator.test.ts
JimiC: Did knew about it.

test/support/supportedExtensions.ts
JimiC: As much as I would like to agree with, I don't. If the node structure of the supported icons changes so should the test. The tests will be useless if they use a redundant node entry structure.

JimiC: If you just want to reflect the structure then use a smaller file with as many entries as necessary, not the whole file. It will be more maintainable this way.

jens1o:
The svg is hardcoded everywhere, and you can't change other files... https://github.com/robertohuertasm/vscode-icons/pull/596/commits/fc8e8f837fe0602bd7fa774ece70ee1e6b3a1809#diff-ec3153353ded9d38b2283de751d804ceR32

I would like to have it responsive, so it could even work with png.

src/commands/index.ts
robertohuertasm: I'm leaving it this way for the moment. We can change this in a later PR if we feel like it.

src/icon-manifest/supportedFolders.ts
robertohuertasm: I'm not sure to understand your question... We're already using a `union type` for the format property.

test/support/supportedExtensions.ts
robertohuertasm: @JimiC, the thing here is to control the inputs of the function. If I use the original file I cannot be certain of what inputs it will receive. If the structure changes then these "duplicated" files should have to change accordingly and, obviously, the tests, too. It's a task that anyway the dev modifying the code will have to do in order to tests his changes. I don't see it as a problem.

Regarding the second point, you're right. 👍

JimiC:
@jens1o This is something that I'll change once @robertohuertasm merges the PR.

robertohuertasm:
@jens1o this is only for the default schema which it's controlled by us. If the user wants to change the icon she can use whatever extension we decide to support.

JimiC:
@robertohuertasm He means the file type association of the default icons.

robertohuertasm:
@jens1o @JimiC you're right.

src/settings/index.ts
```diff
@@ -0,0 +1,95 @@
...
+export class SettingsManager {
+  private settings: ISettings;
+
+  constructor(private vscode: IVSCode) { }
+
+  public getSettings(): ISettings {
+    if (this.settings) { return this.settings; };
+    const isInsiders = /insiders/i.test(this.vscode.env.appName);
+    const version = semver(this.vscode.version);
+    const isGt160 = semver.lt(version.major + '.' + version.minor + '.' + version.patch, '1.6.0');
```
JimiC: I think there is a bug here. In order for `isGt160` to be `true` `semver.gt` function needs to be used. Am I missing something?

jens1o: No, your right. It should be greater than, not lower than.

jens1o:
Thanks @JimiC, so I'm approving, when the bug has been fixed Jimi pointed out.

src/icon-manifest/iconGenerator.ts
robertohuertasm: Solved it!

src/models/IExtension.ts
robertohuertasm: Barrels everywhere in models 😁

test/support/supportedExtensions.ts
robertohuertasm: Reduced the test support files.

robertohuertasm:
I didn't have the time to end default icons customization functionality. I have prepared the code to end it tonight.

src/settings/index.ts
robertohuertasm: I removed it as it was not used anymore.

test/support/supportedExtensions.ts
```diff
@@ -0,0 +1,32 @@
...
+import {
+  IExtensionCollection,
+  IFileExtension,
+  FileFormat,
+  ISpecialFileCollection,
+  SpecialExtensionType,
+} from '../../src/models';
+
+export const extensions: (IExtensionCollection<IFileExtension> & ISpecialFileCollection) = {
+  special: {
```
JimiC: Why go the extra mile?

I believe a simple structure like 

```
export const extensions: IExtensionCollection<IFileExtension> = {
   defaultl: {},
   supported: {}
}
```
would suffice.

robertohuertasm: I didn't want to depend on the `icon` property in order to know what kind of default we're dealing with. Besides that, the `extension` property is completely irrelevant. We're dealing with a different type of model here and I wanted to make it evident. Not to just hack what we already have in order to make it work.

This structure simplifies the way we can apply merging for this kind of special files. If not, then we would have to use the current `extends` system and, as I said before, use the `icon` property value to determine what kind of `default icon` this is.

Anyway, this is a WIP merge, and I'd ask you to bear with me until is finished. I have to say that my first approach was to use the default API in order to make it work. But once I started to think about how I would have to alter the `iconGenerator` and the `manifestMerger`  I thought that making the API clearer would be a benefit in the long run.

JimiC: Fare enough. I'll judge by the result.

robertohuertasm:
Still WIP. Validations and tests pending.

robertohuertasm:
I think it's done. For sure there will be things that can be improved  and maybe docs will have to be extended. I've also provided a new preset for hiding all the folders (like in the old times) 😉 and new tests. 

As usual, feel free to comment or suggest anything.

src/icon-manifest/iconGenerator.ts
```diff
@@ -0,0 +1,385 @@
...
+    if (txt.indexOf('.') === 0) {
+      return txt.substring(1, txt.length);
+    }
+    return txt;
+  }
+
+  private buildIconPath(
+    defaultExtension: IDefaultExtension,
+    schemaExtension: IIconPath,
+    path: string,
```
JimiC: Rename `path` to `iconsFolderBasePath` to not conflict with the name of the package.

src/icon-manifest/manifestMerger.ts
```diff
@@ -0,0 +1,182 @@
...
+    supported: mergeSupported(sFiles, supportedFiles.supported),
+  };
+
+  const folders: IFolderCollection = {
+    default: mergeDefaultFolders(dFolders, supportedFolders.default),
+    supported: mergeSupported(sFolders, supportedFolders.supported),
+  };
+
+  const json = iconGenerator.generateJson(files, folders);
+  // iconGenerator.persist('icons2.json', json);
```
JimiC: Remove commented out code?

package.json
```diff
@@ -109,15 +109,25 @@
           "default": [],
           "description": "These custom associations will override the folder icon associations defined by default."
         },
-        "vsicons.associations.fileDefault": {
+        "vsicons.associations.fileDefault.file": {
           "type": "object",
           "default": null,
-          "description": "This setting will let you change default file icons (dark & light theme)."
+          "description": "This setting will let you change default dark file icon."
```
robertohuertasm: This contribution point is only suited to change the `default dark file icon`, thus the use of the singular. Note that there is a specific contribution point for every default icon possible.

jens1o: Ah, I got it.

JimiC:
@robertohuertasm Great work there. Just wondering if we can override the `Angular2Preset` icons?

src/models/extensions/defaultExtension.ts
```diff
@@ -0,0 +1,8 @@
+import { FileFormat } from './fileFormat';
+
+export interface IDefaultExtension {
+  icon: string; // name of the icon.
+  format: FileFormat | string; // format of the icon
+  _custom?: boolean; // user customization: flag for the icon generator.
```
robertohuertasm: I thought about this too. I don't usually like to use this kind of syntax but I wanted to make evident that this field was not meant to be edited by users. As I don't know users I thought that this was the easiest path for making them know that this property is not intended for modification. I don't rely on them writing the docs so I thought that this kind of syntax would put them in the right track.

src/icon-manifest/iconGenerator.ts
robertohuertasm: Agree :+1:

JimiC:
Regarding the absolute path issue, we have a `getRelativePath` function, can't we use it to support `absolute path -> relative path`?

JimiC:
Is there any mechanism that ensures that in case of a custom file/folder declaration, the custom folder and the file for the icon exists (assuming it follows the `vscode-icons` file naming convention)?

src/models/extensions/defaultExtension.ts
robertohuertasm: We could use another `interface` with just this property but since the users can enter whatever property they want in their settings (remember they don't rely on Typescript) they could easily inject it and the code complexity it would introduce I don't think is worth the effort. The good news is that nothing would happen anyway as the value of this property will get overwritten by the `merge functionality`.

JimiC: I currently working on a way to remove `_custom`. Implementation has finished but have to find a way to make tests pass.

robertohuertasm:
@JimiC, about the `angular2` preset... it can not be overriden at the moment, you're right. It's tightly coupled to the icon names. Maybe we can use some kind of convention... let's say, that it will enable/disable all icons starting with `ng2_`. I will rework it.

robertohuertasm:
@JimiC, about the mechanism to ensure the existence of an icon, it's not present. In the case the icon is not there it just won't be shown. This could be a nice to have that we could introduce in another PR.

robertohuertasm:
Angular2 renamed to Angular to follow AngularJS naming style and new convention about Angular icons: They must start with `ng_`. This way, anyone will be able to customize them and take advantage of the preset.

robertohuertasm:
@JimiC, finally the relative approach is working! It's already pushed! Thanks!

robertohuertasm:
Permission issues arising in Mac... as usual... this time is `package.json` modification by the `iconGenerator`. I'll make this part optional as it's only needed for us when we publish the extension.

README.md
```diff
@@ -19,37 +19,19 @@ I would also like to thank the whole community for their support: bringing in id
 
 ### State of the extension
 
-**With the release of 4.0.0 version, support for old functionality has completely been removed for VSCode**.
+** Release 5.0.0 introduces the ability for users to customize the icons without having to inject anything into the `Visual Studio Code's code`.**
```
JimiC: Remove excess whitespace between `** Release`.

README.md
```diff
@@ -88,12 +70,134 @@ Go to **File > Preferences > File Icon Theme > VSCode Icons**.
...
+"vsicons.associations.files": [
+  { "icon": "myJs", "extensions": ["js", "custom.js"], "format": "svg", "overrides": "js" }
+]
+
+```
+
+#### Custom Icons
+In order to place your custom icons you will have to create a specific folder. Depending on your OS the path will be:
+
+- Windows: `C:\Users\<your_user>\AppData\Roaming\<Code Folder>\vsicons-custom-icons`
```
JimiC: Correction:

>Windows: `C:\Users\<your_user>\AppData\Roaming\<Code Folder>\User\vsicons-custom-icons`

JimiC:
@robertohuertasm `custom_removal` has been rebased.

README.md
```diff
@@ -88,12 +70,134 @@ Go to **File > Preferences > File Icon Theme > VSCode Icons**.
...
+- `vsicons.presets.angular` (true by default)
+- `vsicons.presets.jsOfficial` (false by default)
+- `vsicons.presets.tsOfficial` (false by default)
+- `vsicons.presets.hideFolders` (false by default)
+
+These 4 different presets can be changed manually by using a `settings.json` file inside your project's `.vscode` folder or by changing `vscode user settings`. Take into account that the place where you set these presets, or for that matter, any configuration, will be very important. `User settings` are global so all your projects will be affected by them while `project's settings` are specific to the project and you will be able to switch presets by project. This can be interesting for the `angular` preset, for example.
+
+These presets can also be automatically set by leveraging a new set of commands that can be found by pressing `F1` and then writing down `icons`. You will be presented with some new commands:
+
+- `Toggle Angular Preset (Project Level)`: This command will enable/disable the `Angular 2 icons`.
```
JimiC: Correction:

>This command will enable/disable the `Angular icons`.

.vscodeignore
```diff
@@ -4,8 +4,6 @@
 test/**
 src/**
 **/*.map
-out/src/build-tools/**
-out/src/models/**
 out/test/**
```
JimiC: Add ignore entry for `out/src/example/**`

package.json
```diff
@@ -1,112 +1,174 @@
...
-        "vscode-icons"
+    "commands": [
+      {
+        "command": "extension.regenerateIcons",
+        "title": "Apply Icons Customization",
+        "category": "Icons"
+      },
+      {
+        "command": "extension.ngPreset",
+        "title": "Toggle Angular Preset (Project Level)",
```
JimiC: Better use `Workspace Level` instead of `Project Level` as `Workspace` is the word `vscode` team uses to describe the `project` scope.

PR #601, opened 2017-01-07
Custom property removal

JimiC:
This PR eventually does more than its title says.

1. Removes the use of `_custom` property.
2. Removes `outDir` as parameter (output directory gets resolved in `iconGenrator` constructor).
3. Excludes extension `vsix` for source control.
4. Changes the behavior of the `Apply` functionality, introducing the concept of canceling an action.
5. Remove hardcoded usage of the icon manifest file name and declares it as an extension setting.
6. Makes the `DEBUG` prelaunch task build the extension instead of just compiling it.


JimiC:
It's not necessary anymore as the outDir is set in constructor.

JimiC:
@robertohuertasm Ready for review.

PR #615, opened 2017-01-09
Introduce editorconfig rules.

JimiC:
This PR introduces `editorconfig` ruling. This way we will get rid of the constant changing of spacing in `package.json`.

@robertohuertasm If you want `package.json` spacing as the rest of the `json` files just remove the `[package.json]` rule.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,9 +8,6 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 charset = utf-8
 indent_style = space
-indent_size = 4
-
-[package.json]
 indent_size = 2
 
 [*.md]
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+indent_style = space
+indent_size = 4
+
+[package.json]
+indent_size = 2
+
+[*.md]
+trim_trailing_whitespace = false
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,11 +1,13 @@
 {
-  "tslint.enable": true,
-  "editor.tabSize": 2,
-  "files.exclude": {
-        "out": false 
+    "editor.tabSize": 2,
+    "files.exclude": {
+        "out": false
     },
     "search.exclude": {
-        "out": true
+        "out": true,
+        "images": true,
+        "icons": true,
+        "examples": true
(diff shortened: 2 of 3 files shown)


.vscode/settings.json
```diff
@@ -1,11 +1,13 @@
 {
-  "tslint.enable": true,
```
JimiC: Already a default setting in `vscode`. That's why I removed it.

robertohuertasm: Please, keep it. As I have it disabled in my global settings and I like to manage it per project.

JimiC: Oh, I see.

package.json
```diff
@@ -1,176 +1,176 @@
...
+    "Other",
+    "Themes"
+  ],
+  "keywords": [
+    "icon-theme",
+    "theme",
+    "icons",
+    "vscode-icons"
+  ],
+  "preview": false,
```
JimiC: Why? What's the purpose of it in this project? Are still in preview (beta)?
It was disabled in a previous commit.

.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+indent_style = space
+indent_size = 4
```
JimiC: Actually we aren't. But if so it's no problem.

package.json
robertohuertasm: I asked to keep it in preview in the previous PR but somehow I missed it and this change got in.

JimiC: Can you explain why we are in preview?

.editorconfig
robertohuertasm: If not, the idea was to use 2
```
"editor.tabSize": 2,
```

Which files are you seeing with 4 spaces?

JimiC: That darn `package.json`.

package.json
robertohuertasm: I like it that way. It induces the idea that we're constantly evolving the extension.

.editorconfig
robertohuertasm: 😂 For what I see, most of the .json files. I'll change that.

package.json
robertohuertasm: That's for sure! 😂

.editorconfig
JimiC: So are going with 2 or 4 spaces in `json` files?

robertohuertasm: The idea is we go with 2

JimiC: OK then, I'm done committing.

PR #620, opened 2017-01-10
Convert PNG to SVG (part 6).

### LINKED ISSUE #474: [Icon Request] Support for .EPS
- [x] I want to create an icon request:
      - Type: extension
      - Icon Name: eps
      - Sample original Icon: http://www.redcatstudios.net/wp-content/uploads/2015/03/eps-icon.png
      - Extensions: .eps, .epsf, .epsi

eps

### LINKED ISSUE #571: hpp files use same icon as cpp, should use a separate icon similar to c and h
C++ header files (hpp) use the same icon as C++ source files (cpp). C header files (h) uses a different icon C source files (c). Could so hpp files uses a different icon so they can be seen easily. Suggest an H++ similar to the C++ used for cpp files.

### LINKED ISSUE #584: [Icon Request] Support for .AU3
- [x] I want to create an icon request:
      - Type: extension
      - Icon Name: Au3Script or AutoIT
      - Sample original Icon: [AutoIT](http://seeklogo.com/images/A/autoit-logo-CC4B097310-seeklogo.com.png)
      - Extensions: .au3

### LINKED ISSUE #616: tfignore icon
image

### LINKED ISSUE #619: Icon Association
Can the extension xlsm be associated with the excel icon.

Thanks
Joe

JimiC:
***Fixes #474, fixes #571, fixes #584, fixes #616, fixes #619***

**Changes proposed:**
* [x] Add
* [ ] Prepare
* [ ] Delete
* [x] Fix

**Things I've done:**

This PR replaces 90 icons with an SVG version and introduces 40 more SVG icons.
Also, introduces new icon for a folder named `js`, proposed in #456.


jens1o:
missing the php icon v2(with the mascot). Was it delayed to part 7?

JimiC:
@jens1o It's in there https://github.com/robertohuertasm/vscode-icons/pull/620/commits/96162d415525e584edc9e871eedf2a070a7bd90b?short_path=14dbf70#diff-14dbf704f0f4bc8d6dce9a18f8703b78.

It's easy to miss when searching in 129 icons. 😄

src/models/contributions/vsicons.ts
```diff
@@ -6,6 +6,7 @@ export interface IVSIcons {
     angular: boolean;
     jsOfficial: boolean;
     tsOfficial: boolean;
+    jsonOfficial: boolean,
```
jens1o: Is the php icon missing here?

src/messages/index.ts
```diff
@@ -16,6 +16,7 @@ export const messages = {
   ngPresetMessage: 'Angular icons will be ',
   jsOfficialPresetMessage: 'Official JS icon will be',
   tsOfficialPresetMessage: 'Official TS icon will be',
+  jsonOfficialPresetMessage: 'Official JSON icon will be',
```
jens1o: Is the php icon missing here?

src/models/contributions/vsicons.ts
JimiC: No.

src/messages/index.ts
JimiC: No.

robertohuertasm: @jens1o we're only providing a JSON toggle. If you wan to change the `php` icon you can do it by disabling `php` and [enabling `php2`](https://github.com/JimiC/vscode-icons/blob/cd9e9021949f44043228758af5167d6c487ffa7e/src/icon-manifest/supportedExtensions.ts#L210).

README.md
```diff
@@ -103,6 +103,7 @@ These presets can also be automatically set by leveraging a new set of commands
 - `Toggle Angular Preset (Workspace Level)`: This command will enable/disable the `Angular icons`.
 - `Toggle Official JS Preset (User Level)`: This command will enable/disable the `Official JS icon`.
 - `Toggle Official TS Preset (User Level)`: This command will enable/disable the `Official TS icon`.
+- `Toggle Official JSON Preset (User Level)`: This command will enable/disable the `Official JSON icon`.
 - `Toggle Folder Icons Visibility (User Level)`: This command will enable/disable the `Visibility of the folder icons`.
```
robertohuertasm: Up in the text, a reference to the new `toggle` is missing. See lines 92-97

src/messages/index.ts
jens1o: Okey, I get it. Thanks.

README.md
JimiC: Thanks for point it out. Totally missed it.

BillDenton:
When are these updates going to the marketplace? Thanks.

JimiC:
They are scheduled for late Sunday, early Monday unless an asteroid hits Earth (or a bus us).

yomed:
@JimiC Hi, I noticed that you changed the color of the `.marko.js` icon from gray to blue in this PR. Would you be opposed if I sent a PR to change the fill back to gray? `.marko` is the primary file type, while `.marko.js` is the compiled version, which isn't used very often in the editor.

screen shot 2017-02-02 at 7 48 39 am

robertohuertasm:
Hi @yomed, so your point is that blue is too much color for this kind of file? Sorry if the question looks obvious but just want to be sure as I don't know much about `marko`.

yomed:
@robertohuertasm Yes, that's correct. Marko is a templating language, so users write templates using the `marko` language, and then it gets compiled to javascript to get executed. The `template.marko` file is the one that is edited by devs. The `template.marko.js` file is the compiled output which isn't edited at all, but is occasionally used for debugging.

robertohuertasm:
@yomed, thanks! I think we can change it to the color it was before. Let's wait for @JimiC in case he has some other reason to make it blue.

JimiC:
@yomed 
Since we decided to change the icons to SVG we agreed that we will use the official colors, with as few exceptions as possible.

For `marko` I used the official colors. Since `.marko` got the colorful icon, it was inevitable to use the [single color](https://github.com/marko-js/branding/blob/master/marko-logo-single-color.png?raw=true) for the `marko.js`.

If there was no color guideline I would have gone with yellow, as it's the official color of `JS` and the association is more than obvious.

Generally, we avoid using the gray color on an icon, but for the following exceptions:
1. The official icon color is gray.
2. The official icon color is black.
3. The official icon color is white.

In the above cases, we additionally create an opposite color contrasted icon for the appropriate theme (instead of white we use gray for better contrast in the dark theme). See `Drone`, `Stylelint`, `Tex`, etc.

Now, I have to guess that your request comes from a personal preference, right?
If this is the case, may I remind you that we've given the ability to the user to customize any icon to their liking, exactly for cases like yours. Read all about it [here](https://github.com/vscode-icons/vscode-icons#fine-tuning).

That said, @robertohuertasm has the final saying to what icons are used. If he decides to change it, I have no option but to go along and move along.

yomed:
@JimiC Thanks for the detailed response. Hopefully I can clarify a little bit. The official logo is the rainbow svg that is used for the `.marko` file type. What happened was, the Atom `file-icons` repo only accepts single color icons, so for that purpose I set up the single color logo (intended to be used for `.marko`. In that case I used a brown icon for the `.marko.js` compiled output files, since no gray was available.
Marko branding issue: https://github.com/marko-js/branding/issues/1
Atom `file-icons` PR: https://github.com/file-icons/atom/pull/306

JimiC:
Still, for a single color version, I would choose a color from the official palette. But as I said that's just me.

yomed:
@JimiC I understand your concerns. Would it be helpful if I checked with the marko maintainers to see if the gray fits in with their branding? I know this seems like personal preference, but these files are compiled output, and I think it makes sense to have them be less visually prominent if possible.

robertohuertasm:
Despite having our set of guidelines that somehow rule the way we may choose colors I think this is a case where changing the icon's color to gray makes a lot of sense. 

I share @yomed vision of these files being a subproduct and hence, that's why a less colorful icon would make perfect sense. These are files not to be noticed, at least at the same level the ones that produce them have.

If I would have to decide - and I'm sorry to differ from you @JimiC - I would choose gray. I'd agree with you if we were talking about making the `.marko` extension monochrome. Then there would be no question about it. Gray wouldn't be my choice. But I'm sharing @yomed opinion on this one.

Anyway, @yomed if you will, make the PR and we'll evaluate if the gray scale fits nicely with light and dark themes.

PR #633, opened 2017-01-14
Add support for yarn package manager.

JimiC:
There are several advantages for us to use `Yarn`.

1. Faster dependencies installation (speeds up the CI process noticeably).
2. Dependencies locking. No more different dependencies between machines (as last seen with CI).
3. Dependencies security.

Installation: https://yarnpkg.com/en/docs/install

***[Migrating to Yarn](https://yarnpkg.com/en/docs/migrating-from-npm)***
`npm install` => `yarn install`
`npm update` => `yarn upgrade`


JimiC:
@jens1o `Yarn` doesn't replace `npm`. Consider it as an improved `npm`.

PR #638, opened 2017-01-14
Support for different drives during development

### LINKED ISSUE #634: DEV: If using different drives the extension won't work properly
I use drive `D:` for developing but my user folder is located at `C:`. This makes the custom functionality to fail when trying to figure out what's the relative path.

robertohuertasm:
***Fixes #634***

**Changes proposed:**
* [ ] Add
* [ ] Prepare
* [ ] Delete
* [x] Fix

**Things I've done:**


src/icon-manifest/iconGenerator.ts
```diff
@@ -375,9 +375,30 @@ export class IconGenerator implements IIconGenerator {
...
-    return !this.avoidCustomDetection && this.hasCustomIcon(absPath, filename)
-      ? this.getRelativePath(this.manifestFolderPath, absPath, false)
-      : defaultPath;
+    if (!this.avoidCustomDetection && this.hasCustomIcon(absPath, filename)) {
+      const sanitizedFolderPath =
+        this.belongToSameDrive(absPath, this.manifestFolderPath) ?
+          this.manifestFolderPath :
+          this.overwriteDrive(absPath, this.manifestFolderPath);
+      return this.getRelativePath(sanitizedFolderPath, absPath, false);
+    } else {
```
JimiC: When you use a `return` statement inside an `if` condition, the use of `else` is redundant.

PR #639, opened 2017-01-15
Convert PNG to SVG (part 7).

JimiC:


src/icon-manifest/supportedExtensions.ts
```diff
@@ -78,8 +79,8 @@ export const extensions: IFileCollection = {
     { icon: 'cobol', extensions: ['cbl', 'cob', 'cpy'], format: FileFormat.svg },
     { icon: 'codeclimate', extensions: ['codeclimate.yml'], light: true, filename: true, format: FileFormat.svg },
     { icon: 'coffeescript', extensions: [], languages: [languages.coffeescript], format: FileFormat.svg },
-    { icon: 'config', extensions: ['env', 'config'], languages: [languages.properties], format: FileFormat.png },
```
jens1o: Why did you removed the config extension?

JimiC: `config` extension is used in `xml` language id.
https://github.com/Microsoft/vscode/blob/1.8.1/extensions/xml/package.json#L16

JimiC: As filename it's declared in `properties` language id.
https://github.com/Microsoft/vscode/blob/1.8.1/extensions/ini/package.json#L16

jens1o: Okey, thanks for the clarification.

PR #640, opened 2017-01-15
WIP: Add auto project detection.

### LINKED ISSUE #629: [Task] Auto-detect project on folder opening

Now with the functionality of switching project related icons on/off, an automation functionality could enhance the manual one.

JimiC:
***Fixes #629***

**Changes proposed:**
* [x] Add
* [ ] Prepare
* [ ] Delete
* [ ] Fix

**Things I've done:**

Unfortunately, the code functionality isn't covered via tests. This is something that should be done regardlessly.


src/init/detectProject.ts
```diff
@@ -0,0 +1,107 @@
...
+      }
+
+      const project = fs.readFileSync(res[0].fsPath, "utf8");
+      const projectJson = (typeof project === "string") ? JSON.parse(project) : null;
+
+      if (!projectJson) {
+        return false;
+      }
+
+      return checkForAngular2Project(projectJson, vsicons.presets.angular) || false;
```
JimiC: Just FYI, in case other project detection support is added in the future, this `return` statement can be extended.
i.e.
```
return checkForAngular2Project(projectJson, vsicons.presets.angular)
 || checkForEmberProject(projectJson, vsicons.presets.ember)
 || checkForMeteorProject(projectJson, vsicons.presets.meteor)
 || false;
``

File: src/messages/index.ts [hunk 2]
```diff
@@ -21,4 +23,9 @@ export const messages = {
   enabled: 'enabled',
   disabled: 'disabled',
   restart: '\'Restart\' for changes to take effect.',
+  ng2Detected: 'VSCode-Icons has detected an Angular 2 project. ' +
```
robertohuertasm: I would stick to `Angular` without the version number as it will eventually be changing every 6 moths.

JimiC: Do you know something that I don't know? Are going to release Angular 3 in six months?

robertohuertasm: To be precise, Angular 4 in March, and then a new version every six months according to what the team has been publishing lately

robertohuertasm: There are plenty of links in the internet about this matter. [This is is just an example](http://angularjs.blogspot.com.es/2016/12/ok-let-me-explain-its-going-to-be.html)

JimiC: I'll be damned. Thanks for pointing this out. Totally missed the news.

JimiC:
OK. Because I'm making a lot of changes to the code to support testing, let's keep this PR as a WIP.

robertohuertasm:
I agree.

robertohuertasm:
If it's ok for you I'll publish what we have right now.

JimiC:
I'll probably have it ready by tomorrow but you can never be sure. So the decision is yours.

robertohuertasm:
I think we can publish what we have right now and develop and test your great new addition without pressure. If it's ready during the week it will be published next Sunday. I think this is more than acceptable. Besides that, it doesn't introduce breaking changes to the extension so it won't pose a problem to the users.

PR #643, opened 2017-01-15
Fixes issue with prepublish script and prepares 7.0.1 publication

robertohuertasm:

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -18,4 +18,5 @@ Dockerfile
 tsconfig.json
 tslint.json
 npm-debug.log
+yarn.lock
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Change Log
 
 ## 7.0.0 (January 15, 2017)
```
JimiC: 7.0.1

.vscodeignore
```diff
@@ -18,4 +18,5 @@ Dockerfile
 tsconfig.json
 tslint.json
 npm-debug.log
+yarn.lock
```
JimiC: Good catch. Totally forgot about it.

PR #645, opened 2017-01-17
Add auto project detection.

### LINKED ISSUE #629: [Task] Auto-detect project on folder opening

JimiC:
***Fixes #629***

**Changes proposed:**
* [x] Add
* [ ] Prepare
* [ ] Delete
* [ ] Fix

**Things I've done:**


JimiC:
During coding I stumbled across a behavior bug in `vscode` and have raised an [issue](https://github.com/Microsoft/vscode/issues/18660) for it.

I worked around the reported issue to ensure correct functionality of the feature.

robertohuertasm:
I've been testing the functionality and it seems to be working ok. But while I was at it, I've come up with the need to disable the detection per project. I'm not suggesting this should be done in this PR but I think that maybe it could be a great addition.

I have several monorepos that leverage `Lerna` and contain several `package.json` files. The outer `package.json` has no `Angular` reference but I want the icons to be on because I know I have several `Angular` projects in there. Disabling the detection will be the way to go here, But I would like it to be disabled only when in this project not globally. 

I know I can workaround this by placing the detection setting in my workspace but if I can save myself this it would be great.

What do you think?

JimiC:
The current codebase has the ability to deep search for `package.json` files but I have restraint it only to the root level. We could change that behavior and in your case, the `Angular` icons will not be disabled. The only think I have to ensure though is the opened workspace is root level and not some deeper project. I think I might have a way of doing this. I'll let you know about it when I have something concrete.

PR #656, opened 2017-01-21
Improve the auto detection feature.

JimiC:
Auto detection now searches in subfolders too to determine if the opened project should show specific icons.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -20,6 +20,7 @@ import {
   showCustomizationMessage,
 } from './commands';
 import { getConfig, findFiles, asRelativePath } from './utils/vscode-extensions';
+import { parseJSON } from './utils';
 
 function initialize(context: vscode.ExtensionContext) {
   const config = getConfig().vsicons;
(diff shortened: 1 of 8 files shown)


src/index.ts
```diff
@@ -19,29 +19,37 @@ import {
...
   manageWelcomeMessage(settingsManager);
   detectProject(findFiles, config)
-    .then((path) => {
-      if (path != null) {
-        const project = fs.readFileSync(path, "utf8");
-        const projectJson = (typeof project === "string") ? JSON.parse(project) : null;
-        if (projectJson) {
+    .then((results) => {
+      if (results != null && results.length) {
+        const isInRootFolder = asRelativePath(results[0].fsPath).includes('/');
```
robertohuertasm: This method seems to not be working. I guess that's related to my issue of having different drives. In my case, the project I'm opening for testing purposes it's located in `D:\` while the process executes in `C:\`.

Just FYI:
- `asRelativePath('c:\\GIT\\gitlab\\valudio\\phone-notify-journal\\package.json');` returns the same path.
- `asRelativePath('d:\\GIT\\gitlab\\valudio\\phone-notify-journal\\package.json');` returns `package.json`

JimiC: Indeed. Will look into it.

JimiC: @robertohuertasm I can't reproduce your claim although I'm doing the reverse case. I'm starting the process in `D:` and opening a project that exists in `C:`. The result in both cases is `package.json`.

cwd: `"C:\Program Files (x86)\Microsoft VS Code Insiders"`
__dirname: `"d:\Projects\NodeJS\vscode-icons\out\src"`

And projects in either:
`"c:\Projects\test\package.json"` or `"d:\Projects\test\package.json"`

JimiC: Update: After I corrected the condition check for the `isInRootFolder` (yeah, I made an `oops` there) I  packaged and installed the extension via `vsix`. Everything seems to work just fine.

robertohuertasm: I'll test this again asap and let you know.

robertohuertasm:
I've tested it and it's working now! 👍

PR #657, opened 2017-01-22
Support for smart component extensions.

robertohuertasm:
Adds supports for `*.page.ts|js` (Ionic 2) and `*.container.ts|js` (ngrx).

Adds new icons, too.


robertohuertasm:
I thought about it too but considered that both are conceptually the same thing using different extensions. That's why I used smart_component as the name of the icon, so it can be the umbrella of lots of different extensions that signify that the component is not a merely presentational one. So the new color is just for remarking that this is a stateful component in the case you want to use the file extension for this purpose.

JimiC:
I have no idea about `Ionic 2`. Correct me if I'm wrong. So basically `page.*` and `container.*` are just other names to describe an Angular `component` (because Ionic is built on top of Angular) and thus why you use one icon for both and with similar coloring as the Angular `component`.

robertohuertasm:
@JimiC, that's absolutely right. They are, indeed, componets but the extension is remarking their special condition (stateful, not presentational, smart or whatever you use to name this kind of components).

PR #679, opened 2017-01-26
Improve the canceling behavior of auto project detection.

### LINKED ISSUE #673: Autodetection enhancement
I have this case where I open a project which is not angular. The autodetection tool tells me that and let's me decide what to do. I don't mind disabling the icon and I choose 'close'. Then the settings.json gets modified with a `vsicons.presets.angular: true` and provided that I'm using Git I get a change alert.

It would be nice if we could detect such scenario: closing the popup while having a true default.

JimiC:
***Fixes #673***

**Changes proposed:**
* [x] Fix

**Things I've done:**

### DIFF
--- a/src/commands/index.ts
+++ b/src/commands/index.ts
@@ -59,48 +59,58 @@ function toggleAngularPresetCommand() {
   const preset = 'angular';
   const value = getToggleValue(preset);
   const message = `${msg.ngPresetMessage} ${value ? msg.enabled : msg.disabled}. ${msg.restart}`;
-  togglePreset(preset, value, false);
-  showCustomizationMessage(message, [{ title: msg.reload }], applyCustomization, cancel, preset, !value, false);
+  const initValue = getConfig().inspect(`vsicons.presets.${preset}`).defaultValue as boolean;
+  togglePreset(preset, value, initValue, false);
+  showCustomizationMessage(message, [{ title: msg.reload }],
+    applyCustomization, cancel, preset, !value, initValue, false);
 }
 
 function toggleJsPresetCommand() {
   const preset = 'jsOfficial';
   const value = getToggleValue(preset);
   const message = `${msg.jsOfficialPresetMessage} ${value ? msg.enabled : msg.disabled}. ${msg.restart}`;
-  togglePreset(preset, value);
-  showCustomizationMessage(message, [{ title: msg.reload }], applyCustomization, cancel, preset, !value);
+  const initValue = getConfig().inspect(`vsicons.presets.${preset}`).defaultValue as boolean;
+  togglePreset(preset, value, initValue);
+  showCustomizationMessage(message, [{ title: msg.reload }], applyCustomization, cancel, preset, !value, initValue);
 }
 
 function toggleTsPresetCommand() {
   const preset = 'tsOfficial';
   const value = getToggleValue(preset);
   const message = `${msg.tsOfficialPresetMessage} ${value ? msg.enabled : msg.disabled}. ${msg.restart}`;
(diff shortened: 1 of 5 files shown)


src/commands/index.ts
```diff
@@ -59,48 +59,68 @@ function toggleAngularPresetCommand() {
   const preset = 'angular';
   const value = getToggleValue(preset);
   const message = `${msg.ngPresetMessage} ${value ? msg.enabled : msg.disabled}. ${msg.restart}`;
-  togglePreset(preset, value, false);
-  showCustomizationMessage(message, [{ title: msg.reload }], applyCustomization, cancel, preset, !value, false);
+  const values = getConfig().inspect(`vsicons.presets.${preset}`);
```
robertohuertasm: These lines (62-67) seem like the perfect candidate for a refactor into a function as they seem to be repeated along the rest of the commands.

src/commands/index.ts
```diff
@@ -59,48 +59,68 @@ function toggleAngularPresetCommand() {
...
+    applyCustomization, cancel, preset, !value, initValue, false);
 }
 
 function toggleJsPresetCommand() {
   const preset = 'jsOfficial';
   const value = getToggleValue(preset);
   const message = `${msg.jsOfficialPresetMessage} ${value ? msg.enabled : msg.disabled}. ${msg.restart}`;
-  togglePreset(preset, value);
-  showCustomizationMessage(message, [{ title: msg.reload }], applyCustomization, cancel, preset, !value);
+  const values = getConfig().inspect(`vsicons.presets.${preset}`);
```
robertohuertasm: We could use `object destructuring` to reduce l74 -76. Something similar to this:
```js
const {defaultValue, globalValue: initValue} = getConfig().inspect(`vsicons.presets.${preset}`);
```
If you want to specify the type then:
```js
const {defaultValue, globalValue: initValue} : {defaultValue: boolean, globalValue: boolean} = getConfig().inspect(`vsicons.presets.${preset}`);
```

src/commands/index.ts
JimiC: I thought of that too but wasn't sure if the flow then was understandable to some. I'm more than willing to refactor that.

robertohuertasm: I think that by extracting this piece of code into a function will make it more clear and understandable. People will know that commands are sharing common functionality just by looking at the name of the new created function. Now they have to compare the several lines of each command to notice it. I vote for that refactor! 😜

JimiC: Not debating. My initial approach was to do it as you comment.

src/commands/index.ts
JimiC: This is not doable as the returned types are `{ }` and we need to cast them to `boolean`. Still, I managed to use destructuring and then cast to appropriate type.

robertohuertasm: The second example is with type assertion. What do you mean is not doable?

JimiC: typescript complain in the second example. The types are not compatible.

JimiC: code has to be done like this
```
  const { defaultValue, globalValue, workspaceValue } = getConfig().inspect(`vsicons.presets.${preset}`);
  const initValue = global ? globalValue : workspaceValue as boolean;
  updatePreset(preset, value, defaultValue as boolean);
```

robertohuertasm: I thought the returned types were `any`. The truth is that I haven't tested it with the real code. Just thought of it while reviewing it here.

JimiC: It would have worked with `any` but the returned type is `{ }` as I said.

Issue #653, opened 2017-01-21
[Icon request] angular controller


josegoncalveswho:
- [X] I'm sure this issue is not a *duplicate*?
- [X] I want to create an icon request(if not, remove these lines below):
      - Type: folder
      - Icon Name: controller
      - Sample original Icon (32x32/svg/transparent whenever possible): [url]
      - Extensions: .controller.js

JimiC:
Regarding a request for a folder icon this has been already added in the list of #650.
Regarding a request for a filename icon, we'll see what we can do.

JimiC:
@robertohuertasm can you provide an icon for `controller` or should we use the `component` one?

josegoncalveswho:
Sorry Roberto,

I never got to think about one. I just figured that, for angular 1 projects, it still helps to have some way to distinguish files. Although, most of the times, controllers are coded alongside the directives.

I would go with component, at least.

Sorry I can't be of more help, but I'm on the road! :)

Sent from my iPhone

On 28 Jan 2017, at 16:24, Jimi (Dimitris) Charalampidis <notifications@github.com<mailto:notifications@github.com>> wrote:


@robertohuertasm<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_robertohuertasm&d=DwMCaQ&c=pWn2jKJ-j-AhxLuiRFe-Qw&r=9E3i1tw7ceGN-8DOzq68mU7K4ouHcWCBa-M2aVKQaMw&m=97gE4pSiNnm1gzE9uctyZL2KOK4HqC-f1w1uRjwTpPk&s=n0B06uPRFTHd2Q2o7PxO-f5a1fCjlrJOxofKS8bHtsg&e=> can you provide an icon for controller or should we use the component one?

-
You are receiving this because you authored the thread.
[...]

PR #684, opened 2017-01-27
support angular 1.x detection

RaananW:
**Changes proposed:**
* [X ] Add

**Things I've done:**
The current angular detection only works with angular 2.x projects, detecting the presence of the angular/core package. When using angular 1, the icons will not load correctly.
This checks for 'angular' package, which is expected to be present when using angular 1.x

[short comment]
I very much liked the typescript & angular icons support, would be great to have it back again.

### DIFF
--- a/src/init/autoDetectProject.ts
+++ b/src/init/autoDetectProject.ts
@@ -60,7 +60,7 @@ export function iconsDisabled(name: string): boolean {
 export function isProject(projectJson: any, name: string): boolean {
   switch (name) {
     case 'ng':
-      return (projectJson.dependencies && (projectJson.dependencies['@angular/core'] != null)) || false;
+      return (projectJson.dependencies && (projectJson.dependencies['@angular/core'] != null || projectJson.dependencies['angular'] != null)) || false;
     default:
       return false;
   }


jens1o:
looks good, but why does Travis fail linting a file?

JimiC:
@jens1o Because it exceeds the max line length.

JimiC:
From my perspective, and with all due respect @RaananW, I don't think we need to do what is proposed here. For Angular 1 projects we need to add a different logic so to enable a subset of the Angular icons and at some point, we need to add an icon for `controller`.

Unfortunately, my vote goes to reject the PR, but we should surtainly add an issue to address the request.

RaananW:
Hi guys,
I'd say this - it used to work, and work wonderfully. I use .component and .service, using TypeScript. 

This really is just a suggestion. But think about what you offer - angular detection. I find it very odd when my vs code restarts saying no angular was detected. Again, especially after it worked so wonderfully. 

I'd be happy fixing the line length issue, if it is needed.

JimiC:
The Angular detection was originally designed for Angular 2. I hadn't had the time to reflect on how to support Angular 1 detection. This is something that I want do but given the limited free time I have and so many things to do, this may take a while.

RaananW:
Look, I'm not arguing (at jest I'm not trying to :smile: ), but letting the people choose would be a nicer option than simply forcing it upon them. The flag exists. If the support doesn't work well, the user can disable it. 

It's not my decision, and I totally understand that. But again, this feature was working and was working rather well. You don't have people complaining there are no icons for angular controllers or directives. It was just nice to have.

JimiC:
@RaananW We have icons for `directive` but not for `controller` which is requested (#653). And as you said if the detection doesn't work for you, you can disable it for now. I'm not saying we won't properly support Angular 1 detection, I'm just saying it will take some time to do it and do it right. I'm all in favor of quality rather than quick and dirty software.

RaananW:
At the moment the software forces me to disable the icons. Even when I set the flag to true, it always sets it back to false. Because the project is not detected as an angular project. Even when I set detection to false. So even if I want to icons, I can't have them. 

I wouldn't call this quick and dirty. Old versions of this software supported it. And I don't think vscode is neither quick nor dirty. 

All good! I can only suggest, it doesn't have to be accepted. Closing, have a wonderful Saturday.

JimiC:
@RaananW If you believe there is a bug please report it with reproduction steps. If I set `"vsicons.projectDetection.disableDetect": true` in `workspace` settings, my `vscode` doesn't execute the detection and if I have set `"vsicons.presets.angular": true` then the icons are always shown.

RaananW:
```javascript
"vsicons.projectDetection.disableDetect": true,
"vsicons.presets.angular": true
```
Are already set (in the workspace settings and the user settings, just to be sure). but the icons are still not shown. Might be a local issue, but I actually thought this was a nice solution to it.
Quickly debugging this, this function - https://github.com/vscode-icons/vscode-icons/blob/master/src/init/autoDetectProject.ts#L22 , is executed. icons will be disabled, no matter what the presets.angular's value is (enableIcons is false, disableIcons is true, all because of the angular detection function, the one that was changed in this PR). But there might be something I am not seeing. Or I didn't follow correctly.

I think this issue has grown to be way bigger than its proportions, so I will leave it as is. Carry on doing awesome job.

JimiC:
F1> Icons: Apply Icons Customization

JimiC:
Opened issue #686 to address this.

robertohuertasm:
@RaananW,  first of all thanks for this PR and your invaluable feedback. Have you tried last @ JimiC suggestions? That should leave the extension as it was before last release.

The thing is autodetection is enabled by default now. By disabling it you will avoid it but that's not enough to make your angular icons appear again.  The fact that you set angular=true won't be enough as the extension doesn't apply automatic transformations on project load. You have to manually apply for it (which is what @JimiC was pointing out). Maybe that's another issue to debate,; whether it would be a benefit or not to auto apply transformations automatically when the project loads.

Anyway, I think is worth discussing how to handle angular detection so I will follow this on the newly created issue #686.

@RaananW I'll be more than glad to hear your thoughts and having you in that discussion so we can agree on a proper solution that make everyone happy.

RaananW:
Hi @robertohuertasm -

Yep, that did the trick. I think adding a short line to the configuration would be enough to help me find the reset function myself. Or to understand it is needed in that case.

I will follow the issue and comment gladly. I feel however, that you guys know what you are doing, and that eventually it will work out just fine anyhow :smile:

@JimiC - I am not discouraged. Whenever I see something I can help with, I do my best to help. 
I totally understand that you guys know the code and the features much better than anyone else. I would say that however - if I had an easy option to revert the extension back to an earlier version, I would. Because I think those auto restarts (which can be disabled, I know) and the removal of detection (IMO without thinking about most of angular vscode users, which still use 1.x) were done too quick. 

But again, I really think this has grown way out of proportions. It is just a nice-to-have thing.

PR #700, opened 2017-01-31
Change display name of extension.

JimiC:
This is a proposal PR.

Currently, the extension's name is displayed as `vscode-icons` in the marketplace. I'm proposing to change it to `VSCode Icons` to get in line with the naming convention the majority of the extensions are using.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-icons",
-    "displayName": "vscode-icons",
+    "displayName": "VSCode Icons",
     "description": "Icons for Visual Studio Code",
     "version": "7.1.2",
     "publisher": "robertohuertasm",


robertohuertasm:
Hi @JimiC, I really appreciate your suggestion but I'm not really willing to change the name. I like it the way it is: a little bit different from the rest, lower casing, appealing the dev spirit that most of the vscode users have in them.

JimiC:
Would you like to reconsider?
screenshot - 04_02_2017 12_45_15

jens1o:
I really agree with Roberto, I like it how it goes.

robertohuertasm:
@JimiC , I know this messes with you ocd 😜 and that we have some kind of schizophrenia going on as the `label` (which determines how the name of the icon set is displayed when the user makes the selection) is set to `VSCode Icons` and not `vscode-icons`, which is not really consistent, but I'm happy with it for the moment. 

In my defense, I have to say that this decision obeys to something purely emotional and I have not any single rational argument to throw at you. I really understand your motivations and I see the logic behind it, but I'm somehow attached to the state of things regarding this matter.

PR #714, opened 2017-02-04
Add tmLanguage extensions

Gama11:


robertohuertasm:
Hi @Gama11, I merged this but then I realized that this shouldn't be marked as an extension. `JSON` is supported via `language id` so you should manage it via `vscode` file associations. I removed it in #712 revision.

JimiC:
@robertohuertasm Can you please revert all `tmLanguage` addition? This is not the correct way to support them.

robertohuertasm:
I reverted it in #712

JimiC:
Not all.

robertohuertasm:
I see...

JimiC:
https://github.com/vscode-icons/vscode-icons/blob/master/src/icon-manifest/supportedExtensions.ts#L355

https://github.com/vscode-icons/vscode-icons/blob/master/src/icon-manifest/supportedExtensions.ts#L357

Gama11:
@robertohuertasm I don't think that works for me, since I'm using the vscode-tmLanguage extension. It registers these as separate languages to get proper highlighting (specific to the tmLanguage format, regexes etc) for them: https://github.com/Togusa09/vscode-tmlanguage/blob/master/package.json#L52-L75

JimiC:
@Gama11 Just let us see how we can support it. I'll work on it tomorrow.

robertohuertasm:
#714 opened by @JimiC

PR #720, opened 2017-02-04
new animated gif as screenshot

### LINKED ISSUE #665: [Task] Update README screenshot.

robertohuertasm:
***Fixes #665***


robertohuertasm:
I don't know why here, in the changes section, the gif is not completely reproduced. I'm checking it.

robertohuertasm:
You can see how it really looks like [here](http://downloads.codecoding.com/screenshot.gif)

robertohuertasm:
Finally fixed. It was GitHub changing the end of line.

.vscodeignore
```diff
@@ -11,6 +11,7 @@ examples/**
 images/*.psd
 images/*.ai
 images/screenshot_folders.png
```
JimiC: Do we need to keep this?

README.md
```diff
@@ -10,7 +10,7 @@
 
 Bring icons to your VS Code.
 
-<img src="https://raw.githubusercontent.com/robertohuertasm/vscode-icons/master/images/screenshot_folders.png" >
+![demo](/images/screenshot890x486.gif)
```
JimiC: Any reason to use the size on the filename?

.vscodeignore
```diff
@@ -11,6 +11,7 @@ examples/**
...
+images/screenshot890x486.gif
```
JimiC: Since you are using internal path in README.md to the image you need to remove this line so the image gets into the distro.

JimiC:
And don;t forget to remove the old image.

.vscodeignore
JimiC: And I'm not sure internal path works in the editor.

JimiC:
On an HD monitor the `gif` looks really blurry.

jens1o:
@JimiC is right. Same here.

.vscodeignore
robertohuertasm: I'm not sure either. They seem to be using a browser plugin, so I thought it was similar to an iframe, thus enabling the relative path. I'll check it.

README.md
robertohuertasm: Just to remember me the Github's max image size. I know that I can take it from the image itself but this way is faster in case I need to rework it (which I know it was going to be the case).

.vscodeignore
robertohuertasm: Not really. But as I wasn't sure if the `gif` was going to work I didn't remove it, in case I had to prepare an static image. This is not a finished PR. I wanted you to check the `gif`. If we solve `gif` issues then I'll remove it.

robertohuertasm:
Are you seeing it full screen or in the original proportions? The file' size is around 950Kb. The original one was about 2MB... I don't think we should deliver such a big image, so the options were to reduce it or to think about different alternatives: maybe keeping a static screenshot and then adding small animated gif showing the set of icons in the ìcons folder`.

JimiC:
I don't think that a 2MB `gif` is a problem if you link it to the repo. That way it doesn't get included in the distro. On the other hand, if you want to include it, I still believe there is no problem (see https://github.com/twxs/vs.language.cmake/blob/master/README.md).

JimiC:
But then again the extension is 1.5MB without the screenshot. The screenshot will be larger than the actual extension. I get your point. The only viable solution I can see is to link the image.

README.md
```diff
@@ -10,7 +10,7 @@
 
 Bring icons to your VS Code.
 
-![demo](/images/screenshot890x486.gif)
+![demo](https://raw.githubusercontent.com/robertohuertasm/vscode-icons/master/images/screenshot.gif)
```
JimiC: it's `https://raw.githubusercontent.com/vscode-icons/vscode-icons/master/images/screenshot.gif`

PR #721, opened 2017-02-05
Fix several issues.

JimiC:
1. Refactored ignore files to exclude "npm-debug" files in general. When in debug mode `npm` generates files like "npm-debug.log.1234567" which we don't want to include anywhere.

2. Fix issues with example generator.

3. Add "vim" support via `language id`.

4. Concatenate declaration in `supportExtensions` and fixes duplicate references.


JimiC:
I discovered that there is an issue with associating `Package.pins` file to `swift`. This is not our problem as the icon manifest file gets generated correctly.

JimiC:
I found out what was wrong. Apparently `vscode` doesn't like `PascalCase` names.

PR #730, opened 2017-02-07
Added tfvars, tfstate and tfstate.backup to terraform extensions

supergibbs:
Just added a few more terraform extensions

### DIFF
--- a/src/icon-manifest/supportedExtensions.ts
+++ b/src/icon-manifest/supportedExtensions.ts
@@ -278,7 +278,7 @@ export const extensions: IFileCollection = {
     { icon: 'svg', extensions: ['svg'], format: FileFormat.svg },
     { icon: 'swift', extensions: ['package.pins'], filename: true, languages: [languages.swift], format: FileFormat.svg },
     { icon: 'tcl', extensions: ['tcl', 'exp'], format: FileFormat.svg },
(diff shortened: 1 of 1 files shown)


robertohuertasm:
@supergibbs can you make the suggested change so we can merge your PR? 
You can use [custom icon associations](https://github.com/vscode-icons/vscode-icons#fine-tuning) for the `tfstate.backup` file.

JimiC:
@robertohuertasm IF this doesn;t get followed up by the weekend I'm going to add the `tfstate` support on an upcoming PR and close this unmerged.

robertohuertasm:
@JimiC agreed.

supergibbs:
Sorry for the delay, I've removed those two extensions.

robertohuertasm:
Thanks @supergibbs! @JimiC while we're at it... what do you think about including `tfvars` considering that we're removing non-native language ids support in the near releases?

JimiC:
If `vscode` fixes the language id bug then there is no need to remove our feature. That's why I have this rollback on hold.

robertohuertasm:
I thought we were removing this feature no matter what `vscode` team did because of 3rd party language id extensions not being consistent: `flow` extension using `javascript` as an id and so on.

JimiC:
Still not sure how this is going to play out. I'll just give it some time.

robertohuertasm:
Ok then! Merging this PR! Thanks @supergibbs! 👍

PR #733, opened 2017-02-10
Fix missing dot for .codeclimate.yml, closes #732

### LINKED ISSUE #732: .codeclimate.yml not using CodeClimate icon anymore
I contributed an icon for `.codeclimate.yml` files a while back: #225

However, even though the icon is [still there](https://github.com/vscode-icons/vscode-icons/blob/master/icons/file_type_codeclimate.svg), it's not used anymore:

Using latest vscode-icons (7.2.0) and VSCode 1.9.1.

Gama11:
I think this should fix it anyway, the config for `.travis.yml` also has a leading dot and it works for that.


robertohuertasm:
Fixes #732 
Thanks @Gama11 👍

Issue #526, opened 2016-12-13
[Task] Provide messages in available VSCode languages


JimiC:
- [x] I'm sure this issue is not a *duplicate*?

I know this was experimented on at some point but I think it will add value to this extension.

VSCode available languages:

- [x] English (US)	[en]
- [x] Simplified Chinese [zh-CN]
- [ ] Traditional Chinese [zh-TW]
- [x] French	[fr]
- [x] German	[de]
- [x] Italian	[it]
- [x] Japanese [ja]
- [ ] Korean	[ko]
- [x] Russian	[ru]
- [x] Spanish [es]

jens1o:
I'll do German, since I am German.

JimiC:
@jens1o Only mark the language when it's done with a PR.

jens1o:
Oh, sorry. I misunderstood. But how should the file be called?

JimiC:
May have to consult @robertohuertasm .

robertohuertasm:
The only file containing strings is called `messages.js` but I don't really care much about the naming. If you come up with a cooler name, it is ok for me.

In fact, this is for me a very low priority issue, that's why I dismissed it. I didn't want to loose time in translations every time I changed a string.

sabinrimni:
I saw you were missing the Italian translation so I was working on it following the examples of the other translations. Is there anything else to translate other than `package.nls.json` and `lang.en.ts`? If you still need it of course.

JimiC:
@sabinrimni Sure. It also requires some wiring but if you aren't familiar with `Typescript` just provide the translations and we'll do the wiring.

JimiC:
`vscode-icons` now uses `Crowdin` as it's translation service. All respectful translations contributors, who wish to continue providing their services, are kindly requested to create an account with `Crodwin`, using their `GitHub` account and register with the language they provide translations for.
Thanks in advance.

/cc @MonsieurMan @mariotsi @sabinrimni @ksvitkovsky @iFun @g-plane

JimiC:
Closing in favor of https://github.com/vscode-icons/vscode-icons/wiki/Translation

PR #734, opened 2017-02-11
Add support for i18n messages.

JimiC:
***#526***

**Changes proposed:**
* [x] Add
* [x] Prepare

Providing the codebase to support i18n messages to the extension.

The feature includes fallback functionality to English messages.

Tests included.

@robertohuertasm I have already created a placeholder `lang.es.ts` file for the Spanish messages.
@jens1o I have already created a placeholder `lang.de.ts` file for the German messages.


src/i18n/index.ts
```diff
@@ -0,0 +1,61 @@
...
+            }
+          }
+          msg += message;
+          return;
+        }
+        throw new Error(`${stringifiedKey} is not valid`);
+      }
+
+      stringifiedKey.split('').forEach((char) => {
+        if (char.match(/[a-zA-Z0-9!@%&\(\)-;:"'\[\]\.,<>\?\s]/g)) {
```
jens1o: will this regex will pass ä, ü, ö?

JimiC: Not sure. I'll check.

test/i18n.test.ts
```diff
@@ -0,0 +1,131 @@
...
+              .to.throw(Error, /is not valid/);
+          });
+
+        it('characters',
+          function () {
+            const literalString = '#';
+            const i18nManager = new LanguageResourceManager(resourceCollection);
+            expect(i18nManager.getMessage.bind(i18nManager, 'en', literalString)).to.throw(Error, /is not valid/);
+          });
+
```
jens1o: maybe add a test case for these special chars?

src/i18n/index.ts
jens1o: We should also check for spanish letters/symbols, like the stylized question mark and apostrophes(é...)

jens1o: It won't. See http://regexr.com 
image

Kürbis = pumpkin

jens1o: This works for me: `[a-zA-Z0-9!@%&\(\)-;:"'\[\]\.,<>\?\s\u00F0-\u02AF]`

JimiC: @jens1o We will have to support more than German and Spanish. Chinese is also in the menu. So I'm looking into the `reverse` solution.

jens1o: hmm...

JimiC:
The invalid character list is just a concept of mine. It can be removed or expanded as seen fit.

src/i18n/index.ts
```diff
@@ -47,12 +47,12 @@ export class LanguageResourceManager {
       }
 
       stringifiedKey.split('').forEach((char) => {
-        if (char.match(/[a-zA-Z0-9!@%&\(\)-;:"'\[\]\.,<>\?\s]/g)) {
-          msg += char;
-          return;
+        if (char.match(/[#^*|\\/{}+=]/g)) {
```
jens1o: `\\` is a bad idea. We need it. (Don't show me this message next time)

JimiC: `\\` catches the literal `\`.

jens1o: Huh? 
image

JimiC: Not correct. The literal input is `Don't show`.

jens1o: Okey, then I thought the wrong way. Sorry.

JimiC: Don't get confused with the escaping character.

src/i18n/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const langResourceCollection: ILangResourceCollection = {
+  de: langDe,
+  en: langEn,
+  es: langEs,
+};
+
+export class LanguageResourceManager {
+
+  constructor(private resourceCollection?: ILangResourceCollection | {}) {
```
robertohuertasm: In order to avoid passing the `language` on each `getMessage` call we could set the default language in the constructor and make the `language` parameter  optional in the `getMessage` method, just in case somebody would like to retrieve a specific translation.

I think this would help to reduce the boilerplate in all the translations while keeping the same functionality that we have right now.

JimiC: Good idea. I also have to look into making `langResourceCollection` an independent file. It's far from optimal but I didn't had much time to iterate on it.

PR #746, opened 2017-02-16
Add settings.json and extensions.json to vscode extensions

Gama11:
`extensions.json` can be used for "workspace recommendations": https://code.visualstudio.com/Docs/editor/extension-gallery#_workspace-recommended-extensions

### DIFF
--- a/src/icon-manifest/supportedExtensions.ts
+++ b/src/icon-manifest/supportedExtensions.ts
@@ -308,8 +308,10 @@ export const extensions: IFileCollection = {
       icon: 'vscode',
       extensions: [
         'vscodeignore.json',
+        'settings.json',
         'launch.json',
         'tasks.json',
+        'extensions.json',
         'jsconfig.json',
         'tsconfig.json',
         '.vscodeignore',


JimiC:
@Gama11 Sorry but we have decided that those are too generic to associate them with `vscode`. This could only be applicable if there was a way to determine that those files are under the `vscode` folder.

Gama11:
Quoting @robertohuertasm:

>I'm including tsconfig.json for the moment. settings.json and tasks.json are too generic to apply a vscode icon to them.

`tasks.json` is still part of the list though?

JimiC:
@Gama11 You are seeing the tree and missing the forest.

robertohuertasm:
I'm more leaning towards completely remove the `vscode` icon to all these `json` files than to associate them all with it.  If we want to be consistent, and provided we don't have a way to detect the file tree position of the file, that would be for me the only alternative.

Gama11:
@robertohuertasm Yes, I would agree with that. The main reason behind this PR was that the inconsistent icon usage in my .vscode folders bothered me... If the files use the json or vscode icon doesn't matter that much to me in the end.

JimiC:
@Gama11 Welcome to my "O.C.D. Club".

robertohuertasm:
@JimiC do you have any objections on using `json` icons for `launch.json` and `tasks.json`?

JimiC:
None.

robertohuertasm:
@Gama11 are you willing to create a new PR with those files out? 

As a side note, eventually, `tsconfig.json` and `jsconfig.json` are going to have their own icon but we can keep them at the moment.

PR #751, opened 2017-02-18
Add eco to html extensions

Gama11:
eco is for "Embedded CoffeeScript templates": https://github.com/sstephenson/eco

Couldn't find any logo for eco specifically, but HTML is better than no icon / the default.


Gama11:
Nvm, I can just use `files.associations`.

JimiC:
This is a good candidate to have its own icon like `erb` has. Something like the `html` icon with the `coffescript` in front instead of the `5`.

PR #754, opened 2017-02-20
fix(angular): change routing module icon to allow all routing modules

delasteve:
**Changes proposed:**
* [ ] Add
* [ ] Prepare
* [ ] Delete
* [x] Fix

**Things I've done:**

Angular allows for more than 1 routing module definition. This change makes it so all routing files will get the routing file icon as long as they follow the format `(.*)-routing.module.(ts|js)`.


jens1o:
did you've tested it? I'm not quite sure that this works...

JimiC:
@delasteve Unfortunately, your PR is invalid simply because `vscode` doesn't support `glob` on file associations.

jens1o:
I'd known that. I was right11111111 😆

robertohuertasm:
@delasteve, @JimiC and @jens1o are right. This won't work even if you've removed the `filename` attribute. In this case, you'll be seeing the `module(ts|js)` icon, not the routing one. 

Your changes will only show the yellow icon of routes in case the file was named `*.-routing.module.ts` (note the dot after the *).

You can also take a look at this [vscode's issue](https://github.com/Microsoft/vscode/issues/12493#issuecomment-249117649) if you want to know when this will be supported, if ever... 😞 

I'm closing this. Thanks anyway for your support!

jens1o:
Wait. What about the tests?

JimiC:
@jens1o There is no proper way to catch those cases. At least I can't think of any atm.

jens1o:
@JimiC Maybe check wether the "extension" starts with a point or an alphabetical thing?

JimiC:
I don't know. We could introduce a `regex` but it will restrict things. I can't think of the implications atm. I leave it to @robertohuertasm.

robertohuertasm:
@jens1o that would be limitating some of the probable extensions. I don't see the point here. The `dot` it's implicit. We cannot test against unappropiate uses, i.e: someone who associates `jss` to `js`.  This kind of cases have to be manually validated because they're not breaking any logical rule. 

The same is happening in this case. If @delasteve hadn't made clear his intention in the comment it would be an educated guess to know if the PR was right or wrong. We would need more information about the change from the user, because maybe he was indeed trying to associate `-routing.module.ts` extension to the routing icon (which would be a suspicious case on the other hand).

jens1o:
Got it, thanks!

PR #755, opened 2017-02-21
Add _test and _tests mapping for test folder.

maroy1986:
We use to name our tests folder "_test" or "_tests". I customized my own personal VSCode but I'm making this PR to make this folder name available for everyone.

**Changes proposed:**
* [x] Add
* [ ] Prepare
* [ ] Delete
* [ ] Fix

Just adding yet another way to name the tests folders. Proposing to add _test and _tests.

### DIFF
--- a/src/icon-manifest/supportedFolders.ts
+++ b/src/icon-manifest/supportedFolders.ts
@@ -33,7 +33,7 @@ export const extensions: IFolderCollection = {
     { icon: 'script', extensions: ['script', 'scripts'], format: FileFormat.svg },
     { icon: 'src', extensions: ['src', 'source', 'sources'], format: FileFormat.svg },
     { icon: 'style', extensions: ['style', 'styles'], format: FileFormat.svg },
-    { icon: 'test', extensions: ['tests', 'test', '__tests__', '__test__', 'spec', 'specs'], format: FileFormat.svg },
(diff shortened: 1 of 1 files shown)


jens1o:
Do you have any open-source project that uses this for the tests? I think that this is just custom, I've never seen this ones...

maroy1986:
I have to admit I never seen that before my current project. Neither I saw __test(s)__ before so that's why I told myself why not make a PR for it. The reason behind _test(s) is to have that folder on top of the list. Should be the same reason for __test(s), I guess. The other thing, we're using a monorepo structure so we have one _tests folder per package.

I let the decision to you to include it or not. For now I customized the mapping via my user .settings in VSCode, but then everyone in the team will need to do this. Not a big deal but I think this could profit to others too. I found _test(s) much cleaner then __test(s)__ in a folder structure.

Thanks for your time!

robertohuertasm:
If the motivation of the use of the underscore is just to make your folder appear at the top of the folders tree maybe you can align to the `jest` naming convention which is similar to yours and already supported by `vscode-icons`: `__test(s)__` 

Would this be an option for you and your team?

maroy1986:
Well, I guess I'll close the PR. I thought it was a more common way to name tests folders I wasn't aware of but too bad. Thanks anyway for your time :)

PR #760, opened 2017-02-25
Remove annoying "is active" output

Gama11:
I think this is bad practice - it quickly becomes annoying when you're looking for meaningful output in the dev console while developing extensions.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -71,8 +71,6 @@ function initialize(context: vscode.ExtensionContext) {
 
 export function activate(context: vscode.ExtensionContext) {
   initialize(context);
-  // tslint:disable-next-line no-console
-  console.log('vscode-icons is active!');
 }
 
 // this method is called when your vscode is closed


JimiC:
Personally, I don't find it annoying and I've been contributing to this project a lot.

jens1o:
I disagree with @Gama11, I think it isn't annoying at all.

robertohuertasm:
@Gama11 you have now the option to [disable vscode extensions](https://code.visualstudio.com/Docs/editor/extension-gallery#_disable-an-extension) in case they're getting in the middle of your work.

From my point of view this is just a small log that we drop in order to check that everything is working. This is one of the few sanity logs that we have and I honestly don't find how this could be annoying. I understand that if you have lots of extensions and all of them are leaving logs it can be rapidly become a mess but, as I said, this is just a one time log in the life of the extension and the disable button is just there for you to get rid (temporarily) of the extensions that you want.

Gama11:
I still don't really see what this adds, no other extension I have installed does this.. Perhaps it makes sense while working on the vscode-icons extension itself, but for releases? :/

Gama11:
Still editing the extension sources each time it auto-updates to remove this line again. :P

PR #761, opened 2017-02-25
Refactored tests into groups.

JimiC:
Because I get lost with all these tests, I figure I group them.
It makes the messages more readable and I believe easier to find in the file.


test/iconGenerator/functionality.ts
```diff
@@ -0,0 +1,174 @@
+/* tslint:disable only-arrow-functions */
```
robertohuertasm: is there a reason why this file has not a `test.ts` extension? I already merged it but I can change it in the pending 7.3.0 PR.

JimiC: Ooops.

robertohuertasm: Don't worry @JimiC, I'll change it :wink:

PR #764, opened 2017-02-25
Add support for bithound analysis.

JimiC:
I strongly recommend adding [bithound](https://www.bithound.io)'s dependencies and code analysis support to the project. It's really valuable.

https://www.bithound.io/popular

### DIFF
--- a/.bithoundrc
+++ b/.bithoundrc
@@ -1,0 +1,17 @@
+{
+  "ignore": [
+    "**/.vscode/**",
+    "**/node_modules/**"
+  ],
+  "test": [
+    "**/test/**"
+  ],
+  "critics": {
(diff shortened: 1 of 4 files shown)


.bithoundrc
```diff
@@ -0,0 +1,17 @@
...
+    "**/node_modules/**"
+  ],
+  "test": [
+    "**/test/**"
+  ],
+  "critics": {
+    "lint": {
+      "engine": "tslint"
+    },
+    "wc": {
```
jens1o: `wc`?

robertohuertasm: [Max limit for file length](http://blog.bithound.io/configuring-bithound-101/)

PR #772, opened 2017-02-26
Improve detection flow and add a vscode directory handler on cancel.

JimiC:
I have noticed that when a project doesn't have a user level `settings.json` file or a `.vscode` directory in general, the way we handle the project detection leaves some `garbage` behind when the user cancels the action on the prompt message.

This PR improves the detection flow and handles canceling cases by removing its `echo`.


JimiC:
Looks like I broke `bithound`. :hushed:

JimiC:
For the record https://www.bithound.io/github/JimiC/vscode-icons/detection_flow_improvement

PR #773, opened 2017-02-27
Exclude '.bithoundrc' from distribution.

JimiC:
Sorry guys, my bad. Forgot to add it in `.vscodeignore`.


JimiC:
https://www.bithound.io/github/JimiC/vscode-icons/exclude_bithoundrc_from-distro

PR #774, opened 2017-02-27
Fix the failed file delete issue with default extensions tests.

JimiC:


JimiC:
😯 It failed? Didn't see that coming.

PR #777, opened 2017-02-27
Use 'TSLint' instead of 'ESLint'.

JimiC:


JimiC:
Not sure where this file is used though.

jens1o:
In fact, is was planned in the early stages with the extension. Orginnally, it was my idea. But I think it's obsolete.

JimiC:
That's what I thought. @robertohuertasm Does this go to the `Trash Bin`?

robertohuertasm:
This is from when we used `vanilla js`. It has no sense now.

jens1o:
Why did you merged this? Shouldn't it removed? 😄

JimiC:
Apparently, it has  sentimental value to @robertohuertasm 🤣

robertohuertasm:
Do you want to remove the file? I thought you were talking about the Eslint reference.

jens1o:
nah, we want to remove the line. 😆

JimiC:
Why not remove the file entirely?

jens1o:
Because maybe someone does not know the common conventions when creating PRs?

robertohuertasm:
This document is the one everyone creating an issue or a PR has access to via a link:
image

Why do you want to remove it?

JimiC:
PRs have their own `md`.

JimiC:
So that where it's used, finally. So either we keep the line with reference to `TSLint` or we remove the line as linting is embedded in the CI.

jens1o:
Hmm, I would let it there. So the PR is just fine.

robertohuertasm:
I'll keep the reference so they are advised to test it locally before arriving to CI.

PR #781, opened 2017-02-28
Change Angular preset to false.

JimiC:
As discussed.


jens1o:
Why is bitHound not working?

JimiC:
Because when I create the PR the analysis hasn't finished on my fork and it tries to create an analysis from the commit. If I `force` the PR again it will be fixed.

JimiC:
See?

robertohuertasm:
@JimiC we should change the readme accordingly 

> vsicons.presets.angular (true by default)

jens1o:
Ah, I get it. Thanks.

PR #782, opened 2017-02-28
Add preset to show all folders with folder default icon.

JimiC:
This PR adds a new feature to display all folder with the default folder icon.

This gives the ability to the user to add a custom default folder icon and in combination with this feature to have all folder be displayed with the custom default folder icon as requested in #432.


package.json
```diff
@@ -51,42 +51,47 @@
         ],
         "commands": [
             {
-                "command": "extension.regenerateIcons",
+                "command": "vscode-icons.regenerateIcons",
```
JimiC: I changed the naming to get in line with what other extension maker use. It's kinda a naming convention to avoid conflicts.

src/commands/index.ts
```diff
@@ -1,21 +1,10 @@
...
 import { LanguageResourceManager } from '../i18n';
-import {
-  IconGenerator,
-  mergeConfig,
-  schema,
-  toggleAngularPreset,
-  toggleOfficialIconsPreset,
-  toggleHideFoldersPreset,
-} from '../icon-manifest';
+import * as iconManifest from '../icon-manifest';
```
JimiC: The `import` list was getting a bit long so I figured I switch to the `allias` convention.

robertohuertasm: Why did you decide to make a generic import at the cost of making the code more verbose?

src/commands/index.ts
```diff
@@ -1,21 +1,10 @@
...
 import { extensions as files } from '../icon-manifest/supportedExtensions';
 import { extensions as folders } from '../icon-manifest/supportedFolders';
-import {
-  IFileCollection,
-  IFolderCollection,
-  LangResourceKeys,
-} from '../models';
+import { IFileCollection, IFolderCollection, LangResourceKeys } from '../models';
```
JimiC: So I figured that if we reference up to three (3) imported members we can have it in one line. At least that's the guideline I'll follow.

robertohuertasm: That's ok for me! 👍

src/i18n/langResources/lang.de.ts
```diff
@@ -27,6 +27,7 @@ export const langDe: ILangResource = {
   tsOfficialPresetMessage: 'Offizielles TS Icon wird',
   jsonOfficialPresetMessage: 'Offizielles JSON Icon wird',
   hideFoldersPresetMessage: 'Ordner Icons werden',
+  foldersAllDefaultIconPresetMessage: 'Ordner Icons Auflegemaske werden',
```
JimiC: @jens1o I would like your input on the usage of the word `Auflegemaske`.

src/icon-manifest/manifestMerger.ts
```diff
@@ -122,7 +122,21 @@ export function toggleHideFoldersPreset(
...
+  if (folders.default.folder) {
+   collection.default.folder.disabled = false;
+  }
+  if (folders.default.folder_light) {
+    collection.default.folder_light.disabled = false;
+  }
+  return collection;
+}
+
+// Note: generics and union types don't work very well :(
```
JimiC: I changed the word `HACK` to `Note` so it doesn't get flagged by `bithound`.

robertohuertasm: I noticed that! 👍

src/commands/index.ts
JimiC: I think I have already answered that.

src/i18n/langResources/lang.de.ts
jens1o: Err. That's not German, and I don't understand what this means.

robertohuertasm: Will wait for @jens1o validation to merge this.

jens1o: I'm here 😆

JimiC: It should say `Ordner Icons Auflegemaske werden aktiviert/deaktiviert`.

robertohuertasm: I'm answering almost at the same time as you and I'm missing context! Same happened with @JimiC  in a previous comment! 😅

src/commands/index.ts
robertohuertasm: Yes, timing issues with messaging!! 😅

src/i18n/langResources/lang.de.ts
jens1o: I'm not quite sure what this feature will offer, correct me if I'm wrong: With this setting it's possible to deactivate any custom folder icon, instead, it will show the normal folder icon?

JimiC: It will toggle the folder icons between specific icon and default.

robertohuertasm: I have to say that maybe the thing is that the message is not very self explanatory. The same happens in Spanish and French. Talking about `folder overlay` is maybe too abstract.  Maybe we should be more straight about what will happen. Something like: "Custom folder icons are going to be disabled/enabled"

jens1o: @robertohuertasm That's what I do. But you can't mix it up with enable or disable.

JimiC: @robertohuertasm Not true cause if the user has a custom default folder icon then that will be displayed. That was the point of the feature.

jens1o: ... Help me!

robertohuertasm: @JimiC then "Out-of-box custom icons" or "shipped custom icons". My point is not the correctness of my suggested message but that the intention is not clear in the current one.

JimiC: @jens1o Unfortunately the way we have implemented the message creation forces us to use `enabled/disabled` as the action is a toggle.

jens1o: That doesn't fit.

robertohuertasm: @jens1o I also accommodated the Spanish translations to that structure, which wouldn't be the one I would have preferred but it works.

jens1o: `Standardmäßige Ordner-Icons aktivieren/deaktivieren` would be okayish. But I don't know wether the message is clear.

JimiC: Guys, I'm open to change the message. But it has to comply with the usage of `enabled/disabled`.

jens1o: I think I'd found it. But it's **definitly** not the message I would prefer.

JimiC: @jens1o We are forced to strictly use `aktiviert/deaktiviert`. When I faced an issue with the French translation my sis proposed to use `(s)` for the plural parts (i.e. activée(s)/désactivée(s)).

jens1o: It's not the plural `s`, but instead the grammatical structure.

JimiC: OK. Let's take the rest of the day to ponder on what message to use that can be translated to all languages with the current way we construct the displayed message.

jens1o: I'm worried about the asian languages...

robertohuertasm: 👍 I think that German is the one that should lead the change as it seems to be the most difficult one to accommodate to the structure.

JimiC: @jens1o As the U.S.A. say "Let's cross that bridge when we come to it".

JimiC: @robertohuertasm What could the English message be when taking into consideration that `custom` and `overlay `should not be used?

JimiC: `Out-of-box folder icons will be enable/disabled` maybe?

robertohuertasm: Or maybe `Out-of-box special folder icons`?

JimiC: @robertohuertasm @jens1o Do you want to propose translations for `Out-of-box special folder icons will be`?

jens1o: That's what I translated before, because the other thing completly doesn't fit with enable/disable.

JimiC: @jens1o Is `Out-of-Box spezielle Ordner Icons werden (aktiviert/deaktiviert)` OK with you?

jens1o: wut, completly no. `Out-of-Box` is not even German 😄 Sounds like Google Translate doesn't translate unknown vocabularies. As I proposed before: `Standardmäßige Ordner-Icons aktivieren/deaktivieren`

JimiC: Man we can't use that.

jens1o: But you **have** to adjust them.

JimiC: So, according to what you are saying `Ordner Icons werden (aktiviert/deaktiviert)` is also wrong.

jens1o: No, that's right. 😆 Plural goes over gentive case. I can't explain that well on English, but it right. Maybe your sister has an idea?

JimiC: I'll ask her when she gets of work.

JimiC: Or maybe we change the implementation of the message construction so it can compensate for such scenarios. That we leave though to slightly more message keys that change only on a word.
@robertohuertasm You opinion?

robertohuertasm: My guess is that this is going to happen also with many of the remaining languages... so it's maybe a good idea to get rid of the composed messages and have more translations. I think it will be better for everyone to allow us to have the best translations possible.

JimiC: Agreed then. So I'll make that change too and add `Standardmäßige Ordner-Icons aktivieren/deaktivieren` to the German translation. Care to provide the Spanish one?

robertohuertasm: I was thinking about as there's not any adjective that describes `out-of-the-box`. Let me think about it and I'll come back to you with it.

jens1o: There is also no equivalent in German, I'd used standard, but maybe use default?

JimiC: Back to the old drawing board...

JimiC: So after a lot of pondering I concluded that the English message should be `Specific folder icons will be enabled/disabled`. I'm using `Specific` and not `Special` because those folder icons don't do anything "special". Their function is the same as the others but they are "specific" to their folder naming.

Are we good?

jens1o: Fine.

JimiC: So should the German translation be:
1. `Spezifischer Ordner-Icons werden aktiviert/deaktiviert`
2. `Spezifischer Ordner-Icons aktivieren/deaktivieren`

jens1o: I would use the former, because it's in the nearly future, which fits perfectly, since a restart is needed.

JimiC: Can you be more specific? `1.` or `2.`?

jens1o: The first.

JimiC: @robertohuertasm Spanish translation should be:
1. `Los iconos específicos de Carpeta se va a activarán/desactivarán`
2. `Los iconos de Carpeta específicos se va a activarán/desactivarán`

robertohuertasm: Are we going to keep the same structure that we had before? I mean, are we still composing these sentences dynamically or are these new complete sentences? I'm not sure because you have used `activarán/desactivarán` instead of `activar\desactivar`.

Neither 1 or 2 are completely correct because of the tense that you're using. If you create the sentence like this you should have to use `activar/desactivar`. In case you want to use the future you will have to get rid of `van a`.

The order of the adjective has a very subtle difference... I would use `2.` But depending on whether we're composing this sentence dynamically or not maybe we could improve it.

JimiC: @robertohuertasm Each sentence will be its own once you merge the improvement PR.

JimiC: And the French will be:
`Les icônes de dossiers spécifiques seront activées`
`Les icônes de dossiers spécifiques seront désactivées`

PR #783, opened 2017-02-28
Add support for angular cli tsconfig variants

robertohuertasm:
Last version of @angular-cli (rc1) introduces two variants of `tsconfig.json` to target the `app` and the `tests`

### DIFF
--- a/src/icon-manifest/supportedExtensions.ts
+++ b/src/icon-manifest/supportedExtensions.ts
@@ -306,7 +306,7 @@ export const extensions: IFileCollection = {
     { icon: 'todo', extensions: ['todo'], light: true, format: FileFormat.svg },
     { icon: 'toml', extensions: [], languages: [languages.toml], format: FileFormat.svg },
     { icon: 'travis', extensions: ['.travis.yml'], filename: true, format: FileFormat.svg },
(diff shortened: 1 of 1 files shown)


JimiC:
When did that happened? They release `rc0` on Friday and I just updated to that.

robertohuertasm:
Then take a look inside your `src` folder and in your `.angular-cli.json`

robertohuertasm:
I guess you just updated the `npm package` but if you create a new project you'll definitely see it.

JimiC:
I mean there is no `rc1` release.

JimiC:
And while you are at it I propose we also add `.angular-cli` to `angular` icon association cause it looks like they are heading that way.

robertohuertasm:
I agree but I'll wait until it happens .
I'm quoting you 😜 :
> "Let's cross that bridge when we come to it"

JimiC:
What about `tsconfig.e2e.json`?

robertohuertasm:
I totally missed it! 👍

JimiC:
I suspect that you don't use `e2e` on your projects. We neither but we do on the [ng-stompjs-demo](https://github.com/sjmf/ng2-stompjs-demo) that I contribute.

robertohuertasm:
To be honest, I haven't had the need of it for the moment. Unit tests are covering pretty much what we need as you can test events and interaction with them in your components.

PR #791, opened 2017-03-04
Add french translation for commands and configuration.

JimiC:


JimiC:
Neither does @robertohuertasm but an approval is now required in order to merge any PR.

robertohuertasm:
To honour the truth I have to say I do speak a little French but certainly my level is not the one needed to review any French translation :stuck_out_tongue_winking_eye:

package.nls.fr.json
```diff
@@ -0,0 +1,25 @@
+{
+  "command.regenerateIcons.title": "Appliquer Icônes Personnalisation",
```
robertohuertasm: I would say that this translation is not good. It sounds like a direct English translation.I would say something like `Appliquer la personnalisation des icônes`

package.nls.fr.json
```diff
@@ -0,0 +1,25 @@
...
+  "command.ngPreset.title": "Basculer Angular Préréglé (Niveau d'espace de travail)",
+  "command.jsPreset.title": "Basculer JS Officiel Préréglé (Niveau de l'utilisateur)",
+  "command.tsPreset.title": "Basculer TS Officiel Préréglé (Niveau de l'utilisateur)",
+  "command.jsonPreset.title": "Basculer JSON Officiel Préréglé (Niveau de l'utilisateur)",
```
robertohuertasm: These `toggle preset` translations also sound a little strange to me. I miss the article. If it's ok to you I'll ask one of my work colleagues to review these translations and will come back to you. He's from Belgium.

package.nls.fr.json
JimiC: Sis also expressed a reservation on the translation but I insisted on keeping them as minimal as possible. She told that even so the meaning gets through.

package.nls.fr.json
JimiC: Same as above. No objection to a second opinion.

package.nls.fr.json
```diff
@@ -0,0 +1,25 @@
...
+  "command.hideFoldersPreset.title": "Basculer la visibilité des icônes des dossiers (Niveau de l'utilisateur)",
```
robertohuertasm: I think the whole translation should be reviewed by a French native before being published. 

After consulting with my Belgian colleague he said to me that some parts of the texts are not real French (as I suspected). Of course any French man would understand it but that doesn't mean it's properly structured. In my opinion, we should not publish any translation until it's properly validated by a native.

JimiC:
Because this may take a while, I'm closing the PR.

JimiC:
@robertohuertasm Should we also disable the already added French translations for the messages?

robertohuertasm:
I will consult it with my colleague and let you know.

PR #794, opened 2017-03-07
Simplify scripts in package.json.

JimiC:
1. For `compile` script `-p ./` is implied for `tsc` and thus is redundant.
2. For `tslint` script, linter has a `--project` switch that reads the `include` and `exclude` definitions from the `tsconfig` thus ensuring that correct files get linted and we don't mistakenly again break linting.


tsconfig.json
```diff
@@ -13,9 +13,13 @@
     "sourceMap": true,
     "rootDir": "."
   },
+  "include": [
```
robertohuertasm: I guess that if we use the `include` section, the `exclude` section is no longer necessary.

JimiC: I have deliberately left both sections as they are now shared by the `Typescript` compiler and the `tslint`. I haven't tested removing the `exclude` section.

robertohuertasm: I did and it works. I also looked at the `tsconfig` [docs ](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) just in case and the `exclude` section can be removed without any issue.

JimiC: I read that too but wasn;t 100% sure. I'll remove it and see in the process.

Issue #650, opened 2017-01-19
[Request] Specific folder icons


JimiC:
### We are keeping here a list of requested folder icons.
We are going to update this list as we go along adding icons.

Before requesting a folder icon make sure to check out the already [supported folder list](https://github.com/vscode-icons/vscode-icons/wiki/ListOfFolders).

**Any new requests SHOULD be submitted via comments on this issue and we'll update the list accordingly.**

- [ ] arangodb
- [ ] cmake
- [ ] console, cli, commandline
- [ ] controllers
- [ ] data
- [ ] enum
- [ ] environments
- [ ] gulp
- [ ] graphql
- [ ] grunt
- [ ] minikube
- [ ] `__mocks__`, mocks
- [ ] polyfills
- [ ] resources
- [ ] serializers
- [ ] services
- [ ] shared, common, core

### Debatable
`Debatable` are folder names that are debatable on whether we should provide a specific icon.

- [ ] vendor

robertohuertasm:
`bin`, `lib`, `packages`, `plugins` could also be candidates.

JimiC:
@robertohuertasm Added to the list.

g5codyswartz:
typescript ^^

AmrEldib:
`data` and `config` could be included

kristianmandrup:
`stylus` with `.styl` extension, already [included in manifest](https://github.com/vscode-icons/vscode-icons/blob/master/src/icon-manifest/supportedExtensions.ts#L293) just needs an extension `['styl']` to be added.

JimiC:
@kristianmandrup Folders manifest is [here](https://github.com/vscode-icons/vscode-icons/blob/master/src/icon-manifest/supportedFolders.ts).

kristianmandrup:
Oh, turns out that using [language-stylus](https://marketplace.visualstudio.com/search?term=stylus&target=VSCode&sortBy=Relevance) extension adds an icon by itself, so not sure about conflict resolution strategy in this case?

JimiC:
What do you mean with "icon conflict resolution" strategy?

kristianmandrup:
I mean if two or more VS code extensions each try to define their own icon for a particular file extension.

JimiC:
Extensions don't define icons, we do. We just rely on the extension `language id` to enable the icon for the filenames and extensions they support.

kristianmandrup:
Gotcha! Anyways, it works. Thanks.

[...]

PR #799, opened 2017-03-09
Expand extensions list for `dist` folder icon

zackschuster:
**Changes proposed:**
* [x] Add
* [ ] Prepare
* [ ] Delete
* [ ] Fix

**Things I've done:**

_This PR partially addresses one of the goals in #650 by adding an icon for `lib` folders. If this isn't acceptable it can be dropped._

Reason: `lib` and `release` are common patterns for distribution folders; as an example, Angular UI Router has used both: [release](https://github.com/angular-ui/ui-router/blob/624d94ea924b4d0c113e3bf7ee8c0439e73e8247/package.json#L92), [lib](https://github.com/angular-ui/ui-router/blob/master/package.json#L62)

Ambiguities: Some libraries use `lib` in different ways; for instance, as a source folder or a side-by-side resource with `dist` in the published library. However, the common case (as I understand it) is for `lib` to be some kind of resource meant to be consumed by an end user, mildly analogous to `node_modules`.

### DIFF
--- a/src/icon-manifest/supportedFolders.ts
+++ b/src/icon-manifest/supportedFolders.ts
@@ -13,7 +13,7 @@ export const extensions: IFolderCollection = {
     { icon: 'aws', extensions: ['aws', '.aws'], format: FileFormat.svg },
     { icon: 'bower', extensions: ['bower_components'], format: FileFormat.svg },
     { icon: 'css', extensions: ['css'], format: FileFormat.svg },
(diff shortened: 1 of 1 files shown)


zackschuster:
@JimiC I've updated the PR to remove the association, pending approval from @robertohuertasm.

robertohuertasm:
I agree with @JimiC. Merging this wihout the `lib` association.

PR #803, opened 2017-03-11
Improve CI.

JimiC:
Add build for osx.
Build also against latest stable `node.js`.
Use yarn cache.


JimiC:
Don't worry. First build for OSX takes a while for Travis Ci to setup the wiring for the project. Every subsequent build start along with the Linux ones and ends in about the same time.

.travis.yml
```diff
@@ -1,5 +1,11 @@
 language: node_js
-node_js: "6"
+os:
+  - linux
+  - osx
+node_js:
+  - "node"
```
robertohuertasm: Why do we use 2 different versions?

JimiC: `node` is for latest current `v7.x` and `6` is for the `LTS`.

robertohuertasm: I know that but I'm not sure that we need to use both checks. I'm using node >7, I don't know you and @jens1o.

JimiC: I use 6.10.x which is the LTS.

robertohuertasm: Ok then 😄 It makes sense.

JimiC: I thought about adding also build for node v4 cause some still use it, but then decided against it. People using node should update to the latest LTS at least.

robertohuertasm: I guess that the only meaningful version is the one that is going to be used when publishing the extension. I wouldn't consider <6 either.

robertohuertasm:
By the way, the `osx` build is taking a lot to complete! 😱

JimiC:
When i first enabled it took 2 hours to initiate. So I suggest you leave this PR last.

robertohuertasm:
I would do so! 2 hours!!

PR #811, opened 2017-03-14
spanish translations

robertohuertasm:
Translations for the commands.

Existing translations have been updated using impersonal and formal forms when required according to `vscode` translations style.


JimiC:
LGTM.

PR #814, opened 2017-03-15
Add issue and PR metrics badges.

JimiC:


JimiC:
https://www.traviscistatus.com/

PR #816, opened 2017-03-18
Ignoring a few folders in .vscodeignore

robertohuertasm:
Just checked this and there were some folders missing.


JimiC:
Nice catch. Totally forgot... again.

PR #821, opened 2017-03-19
Improve code coverage by tests.

JimiC:
This PR improves the code coverage and increases the `codecov` rating to 100%.
Alas, it doesn't cover the entire project.

The following files need to be covered also in a future PR:
 - `index.ts`
 - `commands/index.ts`
 - `example/exampleGenerator.ts`
 - `init/welcome.ts`
 - `utils/vscode-extensions.ts`


src/icon-manifest/iconGenerator.ts
```diff
@@ -2,28 +2,19 @@ import * as fs from 'fs';
...
-  IVSCode,
-  IIconSchema,
-  ISettings,
-  IFolderCollection,
-  IFileCollection,
-  IDefaultExtension,
-  IIconPath,
-} from '../models';
+import * as utils from '../utils';
+import * as model from '../models';
```
robertohuertasm: I would use `models` instead of `model` given that we usually  mimic the name of the module when we do a full import.

src/icon-manifest/manifestMerger.ts
```diff
@@ -1,49 +1,39 @@
...
-  IFolderCollection,
-  IExtensionCollection,
-  IExtension,
-  FileFormat,
-  IIconSchema,
-  IIconGenerator,
-  IFileDefault,
-  IFolderDefault,
-} from '../models';
+import * as model from '../models';
```
robertohuertasm: Same comment as before about the the import alias.

robertohuertasm: In fact, I don't really see the benefit of doing such an import. We reduce the size of the import but we lose readability and gain verbosity in our code.

src/init/autoApplyCustomizations.ts
```diff
@@ -6,7 +6,7 @@ export function manageAutoApplyCustomizations(
   userConfig: IVSIcons,
   applyCustomizationCommand: () => void): void {
   if (!isNewVersion) { return; }
-  const propObj = packageJson.contributes.configuration.properties || {};
+  const propObj: any = packageJson.contributes.configuration.properties as object;
```
robertohuertasm: What's the use of `as object` if you have notated `propObj` as any? In fact, the any is implicit in `packageJson.contributes.configuration.properties`.

src/init/autoDetectProject.ts
```diff
@@ -16,12 +16,7 @@ export function detectProject(
   }
 
   return findFiles('**/package.json', '**/node_modules/**')
-    .then((results) => {
-      return results && results.length ? results as model.IVSCodeUri[] : [];
-    },
-    (rej) => {
-      return [rej];
-    });
+    .then((results) => results && results.length ? results as model.IVSCodeUri[] : [], (rej) => [rej]);
```
robertohuertasm: You can omit the parentheses when the arrow has only one argument.

src/init/autoDetectProject.ts
```diff
@@ -66,7 +61,9 @@ export function iconsDisabled(name: string): boolean {
 export function isProject(projectJson: any, name: string): boolean {
   switch (name) {
     case 'ng':
-      return (projectJson.dependencies && (projectJson.dependencies['@angular/core'] != null)) || false;
+      return (projectJson.dependencies && (projectJson.dependencies['@angular/core'] != null)) ||
+        (projectJson.devDependencies && (projectJson.devDependencies['@angular/core'] != null)) ||
+        false;
```
robertohuertasm: What's the use of this final `false`?

src/settings/settingsManager.ts
```diff
@@ -38,36 +38,34 @@ export class SettingsManager implements ISettingsManager {
   }
 
   public getState(): IState {
-    try {
-      const state = fs.readFileSync(this.settings.settingsPath, 'utf8');
-      return JSON.parse(state) as IState;
-    } catch (error) {
-      return {
+    const state = fs.readFileSync(this.settings.settingsPath, 'utf8');
```
robertohuertasm: The `try catch` was also protecting from any error while reading the file.

src/models/settings/settingsManager.ts
```diff
@@ -6,7 +6,7 @@ export interface ISettingsManager {
   getSettings: () => ISettings;
   getState: () => IState;
   setState: (state: IState) => void;
-  setStatus: (sts: ExtensionStatus) => void;
+  updateStatus: (sts: ExtensionStatus) => void;
```
robertohuertasm: This interface is not consistent with the use of the function you make in the `SettingsManager` class. It should return an `IState`

src/icon-manifest/iconGenerator.ts
JimiC: That's what happens when you blindly type and don't pay attention.

src/icon-manifest/manifestMerger.ts
JimiC: The benefit is fewer lines in files, something that `bithound` is complaining about for many of the project's files.

I don't agree with your claim that we lose readability. Now when I see `IFolderDefault`, for example, I know it can be found in `models`. Formerly, I had to scroll back up to the `import` to figure it out.

src/init/autoDetectProject.ts
JimiC: Agreed. Although I'm used to adding the parenthesis in every lamda expression I'll follow the project's coding guidelines.

src/init/autoDetectProject.ts
JimiC: Consider it a force of habit to add `guard clause` for `null`. In this case, it's indeed redundant as we already check for `null`. Nice catch.

src/settings/settingsManager.ts
JimiC: Agreed.

src/models/settings/settingsManager.ts
JimiC: Forgot to change it as the renaming was automated.

src/init/autoApplyCustomizations.ts
JimiC: Now, this is a case of a functionality not been implemented by others. `Typescript` in 2.2.x introduced the use of `object` versus `Object`. When I used that, `tslint` complained that `popObj` was not marked as an `Object` (although technically it is). So I was forced to use `any` until `tslint` fixes the rule.

src/init/autoDetectProject.ts
JimiC: Revising my comment. It turns out that the `guard clause` is indeed needed (praise unit testing).
You see the case is that:
`projectJson.dependencies && (projectJson.dependencies['@angular/core'] != null)` can be `false` and `projectJson.devDependencies && (projectJson.devDependencies['@angular/core'] != null)` can be `undefined`.
`false || undefined` returns `undefined`, thus `|| false` is needed.

src/icon-manifest/manifestMerger.ts
robertohuertasm: I generally try to not use generic imports. Basically because I'm used to think about `tree-shaking` (which is not our case). Personally, I don't like much the verbose style. Maybe I'm used to .NET, where you usually import namespaces in order to not do what you're proposing here. 

Anyway, as this is a matter of style preference I'm not going to discuss it. 😉

JimiC: I think that I'm exactly doing what we do in `.NET`. Alas, here we don't have `namespaces` to import from.

src/init/autoApplyCustomizations.ts
robertohuertasm: It's curious because I'm also using 2.2.1 and I'm not seeing this warning. My previous comment was more about why you annotate the variable `propObj` with two types if, in the end, typescript is going to use the `any` (in this case). If you want to use `object` why the `any`? And if you want to use `any`, why the `object`?

robertohuertasm: I now got the error when using `Object` (I misread your statement). But the point still stands. I mean, what you have in this line is completely equal to `const propObj = packageJson.contributes.configuration.properties`, without any annotation at all.

src/init/autoDetectProject.ts
robertohuertasm: Nice catch!! I tend to use the `!!` operator in these cases but given the length of these expression I think your solution is good!

src/init/autoApplyCustomizations.ts
JimiC: As it is now you are totally right, and indeed it makes no sense to anyone that sees it.
Let me explain further my intention.

The code should be 
```
const propObj = packageJson.contributes.configuration.properties as object
for (const key in propObj) {
```
which is totally valid.

If I use
```
const propObj: Object= packageJson.contributes.configuration.properties as object
for (const key in propObj) {
```
then `tslint` `ban-types` rule complains.

Not here that we want to explicitly annotate `propObj` as an `Object` and not as `any`.

But because `ts` is complaining* I'm forced to change it to 
```
const propObj: any = packageJson.contributes.configuration.properties as object
for (const key in propObj) {
```
to workaround the issue.

as soon as `ts` fixes the issue `: any` can be removed.

*https://github.com/Microsoft/TypeScript/issues/14792

robertohuertasm: If that's a reminder of a future change then I'm OK with it.

JimiC: Maybe I should add a note?

robertohuertasm: Ups, I already merged it.

JimiC: NP. I'll add it to the next PR.

PR #868, opened 2017-04-09
Properly support API Blueprint and MSON via language ids.

JimiC:


JimiC:
After https://github.com/XVincentX/vscode-apielements/pull/72 we need to properly support the `language ids` changes.

PR #894, opened 2017-04-24
Move constant values to their own file.

JimiC:


src/constants/index.ts
```diff
@@ -0,0 +1,5 @@
+export const constants = {
+  urlReleaseNote: 'https://github.com/robertohuertasm/vscode-icons/blob/master/CHANGELOG.md',
```
robertohuertasm: We should use the new url `https://github.com/vscode-icons/vscode-icons/blob/master/CHANGELOG.md`

JimiC: Aha! And figure that I used copy-paste.

Issue #833, opened 2017-03-23
Improvement on angular detection


robertohuertasm:
Now, if you have the Angular preset on (let's say you're working on an Angular project with a specific vscode instance) and you decide to open a non-Angular project, the extension will warn you that you're not in an Angular project anymore. So far, so good.

The problem is that it will also set `"vsicons.presets.angular": false`, and even if you choose `Close` it will modify the `settings.json`.

I was thinking that this modification is not necessary. We don't need to mark every project as `angular: false` as that's the default state for the Angular preset. So if we find a non-Angular project and we happen to have the Angular preset on, we should have to only toggle the preset and not write into the `settings.json`.

JimiC:
I'm already working on that but gave a priority on code coverage,

robertohuertasm:
Cool @JimiC! I forgot to mention the case where we're currently having the Angular preset on and we access into another Angular project. Detection rises again asking for a restart. In this case, nothing should happen or, at least, only suggest to the `settings.json` modification as the restart is unnecessary.

JimiC:
Also noticed it and preparing a fix. The whole detection logic will get a refactor as some things have been fixed from the side of the `vscode` team and the pre-preseting is now redundant. Preset will now change upon selecting an action and in the case of a `Close` actions, no actions will be taken resulting in a cleaner handling.

Canaan-Jin:
I also met.
When i set ``vsicons.presets.angular`` as ``true`` and restart.
Then vscode will restart automatically two times, also set ``vsicons.presets.angular`` as ``false``.
Please forgive me for my poor English.
I don't know if I'm right.

Okoyl:
I don't get angular icons, even when I set `"vsicons.presets.angular": true` at workspace settings.
Used to work flawlessly a few versions back without messing with any settings file.
my user settings:
```json
{
  "vsicons.projectDetection.disableDetect": false,
  "vsicons.projectDetection.autoReload": true,
  "vsicons.presets.foldersAllDefaultIcon": false
}
```

my workspace settings:
```json
{
  "vsicons.projectDetection.disableDetect": false,
  "vsicons.presets.angular": true
}
```

JimiC:
@Okoyl It's mandatory to perform `F1 > Icons: Apply Icons Customizations` whenever you set a preset.

Issue #826, opened 2017-03-20
Restoring the icon manifest doesn't reset the presets.


JimiC:
- [x] I'm sure this issue is not a *duplicate*?

While I was testing the toggle functionality on some improvements I'm making, I noticed that restoring the icon manifest has a side effect because the presets don't get reset.

Steps to reproduce:
 1. Toggle any preset.
 2. Restore the icon manifest.
 3. Toggle the preset again.

Notice here that on step 3 the toggle message is wrong and upon toggling in step 3 any other set preset is reapplied.

Awaiting confirmation.

jens1o:
I thought this is a feature 😆 Do we need to fix it? I like the way how it goes.

JimiC:
@jens1o Some fix is needed but I'm not sure which one, cause there are two ways to fix this.

jens1o:
What are the options?

JimiC:
Either reset the presets or scan the `icons.json` file to determine the state of the icons, in order to present the correct message..

robertohuertasm:
I would go with the second option.

PR #896, opened 2017-04-24
Project detection refactor

JimiC:
This PR intends to fix issue #833 and issue #826 along with improving the `PAD` (project auto detection) logic.

Toggling of the `preset` and `PAD` logic have become smarter and are now relying on icon presence in the icon manifest file to determine their course of action.


JimiC:
I decided to make the `project detection` feature totally unobtrusive.
I believe that the PR can move to the `RC1` stage.

JimiC:
I believe it's prudent to describe the changes made in this PR so you can better evaluate it.

- Presets states will only change according to the user's action. Choosing `Restart` will update the state and `Close` will do nothing. The extension will no longer change the preset state upon showing the informative message and change it back if the user cancels the action.
- Preset state is now determined by the presence or absent of the icon definition in the icon manifest file. This change resolves issues with the logic that determines what informative message will be shown to the user.
- Project Auto Detection (`PAD`) functionality is now totally unobtrusive. This means that either manually restarting or auto-restarting the editor to regenerate the icon manifest, the extension will not write to the user settings file. All this is possible thanks to the above-mentioned change.

From my part, I'm confident that the PR can be merged.

README.md
```diff
@@ -53,11 +53,13 @@ If you're willing to collaborate with us feel free to join our [Github repositor
...
+As from **release 7.7.0**, the `project detection` feature has become smarter and is totally unobtrusive. If you have it disabled, because you found it annoying for any reason, **we urge you to re-enable it** and check out its new functionality.
 
-[@JimiC](https://github.com/JimiC) has improved and further optimized this functionality specifically for this release. But he also has provided the extension with the ability to support multi-language! Take a look at the translation's section if you want to know more or contribute.
+From **release 7.4.0**, as a result of hearing you out, we've revised how the `project detection` works. Now `Angular` icons are disabled by default. So if you want to enable them you will have to actively set the preset to true or rely on the `project detection` functionality.
 
-We still support the ability for users to customize the icons without having to inject anything into the `Visual Studio Code`'s code. 
+We support ***localization*** of the extension! We intend to expand the supported languages as the `Visual Studio Code` adds them. Take a look at the translation's section if you want to know more or contribute.
```
jens1o: Maybe link the `translation's section`?

README.md
```diff
@@ -53,11 +53,13 @@ If you're willing to collaborate with us feel free to join our [Github repositor
...
+
+We keep supporting the ability for users to customize the icons without having to inject anything into the `Visual Studio Code`'s code.
```
jens1o: `... and we never will stop shipping this feature!`?

README.md
JimiC: Good idea!

README.md
JimiC: It's implied.

jens1o: Yeah, but maybe we can make this clear. Anyway...

robertohuertasm:
Hi @JimiC, I've had the opportunity of testing a little bit the functionality and here you are some things that I've found for your consideration:

**Premise 1**

I have an `Angular` project but I have decided that I don't want the icons (because I'm weird) and I've set my preset to `vsicons.presets.angular: false`.

**Cases**

1. If the `Angular` icons are disabled, whenever I open the project, `PAD` will ask me for a restart, forcing me to disable it in order to not suffer the constant warning.

2. If the `Angular` icons are enabled, `PAD` will not  be shown.

**Premise 2**

I have a non `Angular` project but I have decided that I want the icons (because I'm weird) and I've set my preset to `vsicons.presets.angular: true`.

**Cases**

1. If the `Angular` icons are enabled, `PAD` will not be shown.

2. If the `Angular` icons are disabled, `PAD` will not  be shown.

**Premise 3**

I have a non `Angular` project and I have decided that I dont want the icons and I've explicitly set my preset to `vsicons.presets.angular: true`.

**Case**

1. If the `Angular` icons are enabled, `PAD` will not be shown.

**Premise 4**

I have a non `Angular` project but I have my `Angular` icons enabled, because I was just working in an `Angular` project.

**Case**

1. If the `Angular` icons are enabled, `PAD` will not be shown.


**Opinion**

I think custom preset should take precedence in case it contradicts the project's nature.

In non `Angular` projects, one may think that maybe it's not worth to disable the `Angular`icons because probably the user will not be using any of the `suffixes` that we use for `Angular` but from my point of view, those cases should also be covered by `PAD`. The same way it detects that a project is an `Angular` project it should detect that another one it isn't.

Sorry for the long comment. 😜 If I find something else I'll let you know. This PR is a really great improvement to this feature. Congratulations for your hard work on this! 👍

README.md
```diff
@@ -53,11 +53,13 @@ If you're willing to collaborate with us feel free to join our [Github repositor
 
 ## State of the extension
 
-With **release 7.4.0**, as a result of hearing you out, we've revised how the `project detection` works. Now `Angular` icons are disabled by default. So if you want to enable them you will have to actively set the preset to true or rely on the `project detection` functionality.
+As from **release 7.7.0**, the `project detection` feature has become smarter and is totally unobtrusive. If you have it disabled, because you found it annoying for any reason, **we urge you to re-enable it** and check out its new functionality.
 
-[@JimiC](https://github.com/JimiC) has improved and further optimized this functionality specifically for this release. But he also has provided the extension with the ability to support multi-language! Take a look at the translation's section if you want to know more or contribute.
+From **release 7.4.0**, as a result of hearing you out, we've revised how the `project detection` works. Now `Angular` icons are disabled by default. So if you want to enable them you will have to actively set the preset to true or rely on the `project detection` functionality.
```
robertohuertasm: Maybe we should change `project detection` to `project auto-detection` and introduce the acronym `PAD`?

JimiC:
@robertohuertasm 
>**Premise 1**
I have an `Angular` project but I have decided that I don't want the icons (because I'm weird) and I've set my preset to `vsicons.presets.angular: false`.
**Cases**
>1. If the `Angular` icons are disabled, whenever I open the project, `PAD` will ask me for a restart, forcing me to disable it in order to not suffer the constant warning.

A side effect that can not be avoided. Exactly the point for the `Disable Detection` option.

>2. If the `Angular` icons are enabled, `PAD` will not be shown.

PAD works as intended.

This case beats the purpose of PAD and should be considered extreme (weirdo).

>**Premise 2**
I have a non `Angular` project but I have decided that I want the icons (because I'm weird) and I've set my preset to `vsicons.presets.angular: true`.
**Cases**
>1. If the `Angular` icons are enabled, `PAD` will not be shown.
>2. If the `Angular` icons are disabled, `PAD` will not be shown.

If you haven't `Apply`, PAD detects no `Angular` icons in the manifest, PAD message will not be shown.
If you have `Apply`, PAD detects `Angular` icons in manifest, PAD message will be shown.

This case beats the purpose of PAD and should be considered extreme (weirdo).

>**Premise 3**
I have a non `Angular` project and I have decided that I don't want the icons and I've explicitly set my preset to `vsicons.presets.angular: false`.
**Case**
>1. If the `Angular` icons are enabled, `PAD` will not be shown.

PAD detects `Angular` icons in manifest, PAD message will be shown.

>**Premise 4**
I have a non `Angular` project but I have my `Angular` icons enabled, because I was just working in an `Angular` project.
**Case**
>1. If the `Angular` icons are enabled, `PAD` will not be shown.

PAD detects `Angular` icons in manifest, PAD message will be shown.

> I think custom preset should take precedence in case it contradicts the project's nature.

As I explicitly mentioned the new functionality ignores the preset and acts according to the icons presents or absent in the manifest file. `Angular` preset is only taken into consideration when using `Toggle` feature or manually changing the preset and `Apply Icons Customization` is executed. If we take preset into consideration then the user can manipulate it and `PAD` messages become inaccurate (we are back to issue #826).

>In non `Angular` projects, one may think that maybe it's not worth to disable the `Angular`icons because probably the user will not be using any of the `suffixes` that we use for `Angular` but, from my point of view, those cases should also be covered by `PAD`. The same way it detects that a project is an `Angular` project it should detect that another one it isn't.

Project detection is still part of the equation. PAD determines the action based on icons in the manifest file and that the project is a detectable one (in the current implementation we detect only `Angular` ones).

P.S.:
I believe in some cases you lost the feeling of the icons being present or not in the manifest. Happened to me too while testing. I resolved it by always applying the preset.

README.md
JimiC: Feel free to change that before releasing.

robertohuertasm:
I guess this issue would be better clarified by using our Slack channel. 

I'll retry it again just to be sure but I've tested it more than once with the same results.

Just a quick note about your comment regarding the return of #826. I'm ok with using `icons.json` as the source of truth in order to know what's the current state of the toggle. What I was saying is that PAD should have to take the preset into account when deciding to offer the option to restart or auto-restart. For me, if the user has explicitly inserted the preset value into the project's settings that should be honored (just in terms of PAD).

robertohuertasm:
Premise 3 confirmed. PAD is not being shown.

README.md
robertohuertasm: @JimiC, I forgot to tell you that you can change the Changelog and introduce your own message. I will add the rest once the release is ready. 😉

JimiC:
After concluding the dev meeting @robertohuertasm has proposed a new behavior for `PAD`. The following table demonstrates the different proposed behaviors.

image

@robertohuertasm In case l haven't understood your proposal please feel free to correct me.
@jens1o Thoughts?

jens1o:
I'm fine with it, since it looks like it is mostly backward-compatible.

robertohuertasm:
@JimiC, your table is ok with what we discussed yesterday.:+1:

README.md
JimiC: @robertohuertasm I decided not to change the changelog as I have communicated the message via the README. Don't keep the PR for this.

PR #901, opened 2017-05-01
chore(package): update @types/chai to version 3.5.2

JimiC:


robertohuertasm:
How is it that there's an empty comment by @JimiC every time Greenkeeper creates a MR? Is this something that we can configure? @JimiC do you know something about it?

PR #925, opened 2017-05-12
support folder icon `lib`

jens1o:
**Changes proposed:**
* [X] Add
* [ ] Prepare
* [ ] Delete
* [ ] Fix

**Things I've done:**

Some well-known php libraries use the lib folder like the src folder.
Example: https://github.com/doctrine/cache


src/icon-manifest/supportedFolders.ts
```diff
@@ -28,7 +28,7 @@ export const extensions: IFolderCollection = {
     { icon: 'meteor', extensions: ['.meteor'], light: true, format: FileFormat.svg },
     { icon: 'sass', extensions: ['sass', 'scss'], light: true, format: FileFormat.svg },
     { icon: 'script', extensions: ['script', 'scripts'], format: FileFormat.svg },
-    { icon: 'src', extensions: ['src', 'source', 'sources'], format: FileFormat.svg },
+    { icon: 'src', extensions: ['src', 'source', 'sources', 'lib'], format: FileFormat.svg },
```
jens1o: can't alphabetize...

robertohuertasm:
I'm not sure of this change. I tend to use of `lib` folder as a folder for libraries and seeing the same icon of the source folder would seem confusing to me. I would expect it to have some books or something like that. 

@JimiC, how do you see it?

jens1o:
That's a good point, and actually you are right. The example I showed is an under-the-hood dependency, which doesn't to anything except providing classes/interfaces/methods/functions...

But I didn't knew about what it should look like, that's why I decided to use the plain src-icon.

JimiC:
We have already discussed that 'lib' folder should be non language specific and I already proposed it to be a stack of books icon referencing to a 'library'.

jens1o:
Okay, I'll close it till we have the icon.

robertohuertasm:
Just for the sake of @JimiC's [mention](https://github.com/vscode-icons/vscode-icons/pull/799#pullrequestreview-26132887)

PR #927, opened 2017-05-13
Fix null reference of PAD result object.

### LINKED ISSUE #923: Angular icons not shown
The icons for Angular were shown but somehow they are not there anymore.

I already tried some solutions like editing the workspace settings:
vs icons

But no changes:
vs icons explorer

The .angular-cli.json is shown correctly.

I'm using version 7.7.0.

What could be the problem?

JimiC:
***Fixes #923***

**Changes proposed:**
* [x] Fix

**Things I've done:**

@robertohuertasm Consider a patch release.


jens1o:
Maybe add a test, so this problem won't be get again?

JimiC:
Unfortunately, every file that has a dependency on `vscode` is impossible to add tests to it. We will have to heavily refactor those files.

PR #948, opened 2017-05-27
Fix new version check flow.

### LINKED ISSUE #857: Lose icons customization each time extension update
**Checklist:**

Hi, I noticed that each time there's a new update available for vscode-icons, after installing it and reload Visual Studio Code, I loose all my icons customization I made into my settings.json. There still there but not applied. The only thing I need to get them back is to open the command palette and choose `Icons: Apply icon customization `. Then everything is back to normal.

This is becoming an hassle to do each time the extension is updated. Is there a reason why you need to reset those when you update the extension?

That would be great If it keeps the customization or reapply them automatically when the extension is updated and reloaded.

Thanks!

JimiC:
***Fixes #857***

**Changes proposed:**
* [x] Fix

**Things I've done:**

### DIFF
--- a/src/init/welcome.ts
+++ b/src/init/welcome.ts
@@ -14,11 +14,8 @@ export function manageWelcomeMessage(settingsManager: ISettingsManager) {
     return;
   }
 
-  if (settingsManager.isNewVersion()) {
-    settingsManager.updateStatus(state.status);
-    if (!getConfig().vsicons.dontShowNewVersionMessage) {
-      showNewVersionMessage(settingsManager);
-    }
+  if (!getConfig().vsicons.dontShowNewVersionMessage) {
+    showNewVersionMessage(settingsManager);
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/init/welcome.ts
```diff
@@ -14,11 +14,8 @@ export function manageWelcomeMessage(settingsManager: ISettingsManager) {
     return;
   }
 
-  if (settingsManager.isNewVersion()) {
-    settingsManager.updateStatus(state.status);
-    if (!getConfig().vsicons.dontShowNewVersionMessage) {
-      showNewVersionMessage(settingsManager);
-    }
+  if (!getConfig().vsicons.dontShowNewVersionMessage) {
```
robertohuertasm: This message should only be shown when a new version is delivered. Now it will be shown always unless the `dontShowNewVersionMessage` preset is set to false.

JimiC: Indeed. Good catch. Fixed.

PR #956, opened 2017-05-28
Remove chai-as-promised dependency and update to chai v4.

JimiC:
This PR removes the dependency to `chai-as-promised` to facilitate the update to `chai v4`.


jens1o:
I wonder why v4 does not have any breaking change...

PR #963, opened 2017-05-28
preparing 7.8.0 version

robertohuertasm:


robertohuertasm:
I had an issue while publishing and 2 latest PR didn't got published so 7.8.1 is out with those two.

Issue #384, opened 2016-10-13
Icons Enable command is no longer supported for VSCode >= v1.6.0.


simonguo:
Icons Enable command is no longer supported for VSCode >= v1.6.0.

jens1o:
That's true. You can only use the File Icon Theme.

PR #996, opened 2017-06-04
Make it possible to underscore folder in include paths with css supersets

jens1o:
**Changes proposed:**
* [X] Add

image


JimiC:
Can you provide any doc or reasoning for this PR?

jens1o:
For example here: https://github.com/imperavi/kube

In this repository, and in others, they use it to mark dependencies(that are included in the framework), so not any file is directly at the top but in a clean view.

robertohuertasm:
I thought this underscore convention was just for partial files in `sass` and `less` and somehow assumed that it had extended to folders too, but I've found [#384](https://github.com/sass/sass/issues/384) and it seems to be a very old issue that still has not been closed.

After googling a little bit for "_sass" and "_scss" I haven't found any official docs about the use of this syntax. I've seen that some projects like `Jekyll` and `Docker` make use of it, but I'm under the impression that they use it to remark some kind of prebuild vs final state. And I say so because they also use this convention with `_includes` , `_layouts`...

JimiC:
So if we are to map also underscored folder names (along with the dotted ones) I suggest we add a marking enum mechanism (like we do with `language`) and generate the mapping internally.

jens1o:
I like that idea. (Closing because this will be handled otherwise)

Issue #998, opened 2017-06-04
Provide mechanism to map folder names to underscored and dotted ones in icons manifest


JimiC:
- [x] I'm sure this issue is not a *duplicate*?

JimiC:
@robertohuertasm @jens1o 
1. I'm terrible at naming. What should we name this new `enum`?
2. I'm thinking of making this `enum` as flag* type like in `.NET`.

Thoughts?

* https://www.typescriptlang.org/docs/handbook/enums.html
https://basarat.gitbooks.io/typescript/docs/enums.html#enums-as-flags

jens1o:
`AdditionalFileextensionTypes`

JimiC:
@jens1o May I remind you that we are atalking about folders.

jens1o:
I had written to many php files today(around 35 to get something done). Sorry.

`AdditionalFolderExtensionTypes`

An extension of folders grouped together... somehow?

JimiC:
Forget `flags` enum they create confusion in the `supportFolders` API.
Naming still pending...

JimiC:
@jens1o I'm looking something for something descriptive like the `languages` we have in `supportedFiles`.

jens1o:
brainfuck got real.

robertohuertasm:
Sorry @JimiC what would be the idea here? I'm not sure to understand how you want to tackle this.

robertohuertasm:
If I understand correctly. You're planning to create some sort of automation that will allow the folders to support `_` and `.` syntax.  I guess these flags will affect all the extensions in a specific folder definition.If that's the case I think this would not be a desirable scenario.

Let's take this as an example:
```
{ icon: 'git', extensions: ['.github', '.git', 'submodules', '.submodules'], format: FileFormat.svg },
```
Let's say I want to support underscores for `submodules` and only dots for the others. How would a flag support this?

In my opinion, just adding the `_`  as another extension is the easiest an clearer way to proceed. Unless your idea solves this issue :wink:

JimiC:
@robertohuertasm By introducing a new flags `enum`.

```
enum Map {
  none,
  dotted = 1 << 1,
  leadingUnderscore = 1 << 2,
  trailingUnderscore = 1 << 3,

  leadingUnderscoreAndDotted = leadingUnderscore | dotted,
  fullyUnderscored = leadingUnderscore | trailingUnderscore,
  all = dotted | fullyUnderscored,
}
```
and API like:
```
[...]

PR #1004, opened 2017-06-07
Fix lodash remove predicate parameter.

### LINKED ISSUE #1003: Extension doesn't get removed in 'mergeSupport'

I stumbled upon the issue where an `extension` entry wasn't removed from the extensions array.
Investigating it I found that we are incorrectly using `lodash remove`. That's not totally our mistake because the blame is on the `lodash` types for `remove`.
You see we are using the `remove(array: any[], prediacate: string)` declaration which is wrong. According to `lodash` itself (https://github.com/lodash/lodash/blob/4.17.4-npm/remove.js) the correct declaration is `remove(array: any[], predicate: () => {})` as predicate expects a function which gets invoked per iteration.

JimiC:
***Fixes #1003***

**Changes proposed:**
* [x] Fix

**Things I've done:**

### DIFF
--- a/src/icon-manifest/manifestMerger.ts
+++ b/src/icon-manifest/manifestMerger.ts
@@ -76,7 +76,7 @@ function mergeSupported(
     file.extensions.forEach(ext => {
       final
         .filter(x => x.extensions.find(y => y === ext))
-        .forEach(x => _.remove(x.extensions, ext));
+        .forEach(x => _.remove(x.extensions, el => el === ext));
     });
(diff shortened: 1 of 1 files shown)


JimiC:
We didn't have any side effects till now but the implementation is wrong. Stumbled upon it when coding feature of #998.

By the way, somehow I managed to push the branch to the `main` repo. By bad.

@robertohuertasm `appveyor/branch` is missing `Required`. Please fix.

PR #1007, opened 2017-06-07
Update dependencies.

JimiC:
Should be merged right before next release.


src/commands/index.ts
```diff
@@ -226,6 +228,8 @@ function handleAction(btn: models.IVSCodeMessageItem, callback?: (...args: any[]
         handlePreset();
       }
       break;
+    default:
+      break;
```
jens1o: Wut?

JimiC: To make `bitHound` happy (without it, it complains about it with a warning).

jens1o: I mean, why don't we remove that rule?

JimiC: It's not our rule but `bitHound`'s.

jens1o: :(

PR #1018, opened 2017-06-10
New method retrieving a correctly merged settings object

### LINKED ISSUE #934: Change workspace configuration behavior
Now, if you define some icons in your workspace configuration it will override the user configuration, so you will end up losing all the customizations you have. 

Ideally, the workspace configuration should be merged to the user configuration. It seems that `vscode.workspace.getConfiguration` method does the overriding so the only way to overcome this is to provide a slighty different syntax in the workspace. 

The idea is to keep backwards compatibility so the current behavior will be kept and people will not lose what they already have but also to provide this new syntax that will allow to keep both workspace and user customizations.

robertohuertasm:
***Fixes #934***

**Changes proposed:**
* [ ] Add
* [ ] Prepare
* [ ] Delete
* [x] Fix

**Things I've done:**


JimiC:
I have a few questions:
1. If same declarations are in both settings, which is proceeding?
2. Any reason behind redeclaring `vscode.d.ts`. For the record, I'm familiar with the technic (I've done it [here](https://github.com/TBear79/winston-fast-rabbitmq/blob/master/typings/winston-fast-rabbitmq.d.ts)). I think `IExtendedWorkspaceConfiguration` was suitable as it allowed the use in tests.

src/vscode.d.ts
```diff
@@ -0,0 +1,9 @@
+import { WorkspaceConfiguration } from 'vscode';
+import { IVSIcons } from '../contributions';
+
+declare module 'vscode' {
+  interface WorkspaceConfiguration {
+    vsicons: IVSIcons;
+    inspect<T>(section: string): {defaultValue: T, globalValue: T, key: string, workspaceValue: T} | undefined;
+  }
+}
```
jens1o: I don't get why this is here. Visual Studio Code imports that automatically?

JimiC: This is a technic to extend the types declarations of existing typings.

jens1o: Okay.

robertohuertasm: 1. Workspace always wins except in file and folder associations, where a merge is done and removed identical entries.

2. As we're not using any of this in the tests I thought to provide a more natural way of dealing with de vscode api. I would consider having the `IExtendedWorkspaceConfiguration` if needed in the tests, which it seems not to be the case for the moment.

robertohuertasm: I'll indulge you and add it to the `test support` folder 😉

robertohuertasm:
This PR is not for that issue. This fixes a bug: workspace associations overriding all the user associations.

JimiC:
Personally, I don't consider it a bug. It doesn't break the extension it just doesn't have the behavior a user would expect. On another note, I believe it's better to move the merging function in another file, that doesn't have `vscode` importation dependency and add test cases to it.

robertohuertasm:
Whatever, it doesn't work as expected. If I put a workspace setting suddenly `all` my user settings are gone. I'd call it a bug but never mind. 

I'll take a look at your suggestion. 😉

robertohuertasm:
I've added the `Mockery` package so we can mock full modules. This is specially helpful for testing the inaccessible `vscode` module. I guess we can now test the whole application if we want.

robertohuertasm:
image

Although I get this in the coverage page I get this in the console report:
image

Any idea?

JimiC:
@robertohuertasm The tests don't cover the code. `Red` number is the line that isn't covered at all. The `yellow` number indicates that some code cases aren't covered. In this case, it's `branches` (condition branching).

I'll have a look myself.

JimiC:
@robertohuertasm Found it. The test doesn't cover the `default` case of the `config.inspect.swtch`.

jens1o:
How do you find all these cases?

JimiC:
@jens1o It's fairly simple.
1. From Browser: Add the `Codecov` extension for your browser. Then inspect the [file](https://github.com/vscode-icons/vscode-icons/blob/29c8093c745501501b03b90ce4fcc1f7ce49ec92/test/utils/vscode-extensions.test.ts) in question.
2. Locally: Run `npm t`, then locate the `index.html` file, under `coverage\lcov-report` folder, and open it in your browser. Navigate to the file in question.

jens1o:
Okay, thanks for the heads up. Will consider this in next prs.

robertohuertasm:
I don't know what happened yesterday. As you could see, my coverage file showed 100%. Now it seems to work and shows me the default case in yellow. Something weird...

JimiC:
@robertohuertasm I'll already on it in fixing it.

JimiC:
@robertohuertasm Ohh, come on. Your are sucking all the fun. 👅

robertohuertasm:
@JimiC I've already pushed 100% coverage. 

My issue was I was not able to find the pending line as my coverage page was not showing it. Indeed, it was showing incorrect stats. But, as I said, this morning I built the code again and it worked.

jens1o:
Is there a reason why we don't use `() =>` syntax there?

JimiC:
Recommended pattern by `Mocha` so not to lose `this` reference.

PR #1026, opened 2017-06-18
Improve handling of manually changed configurations.

### LINKED ISSUE #1019: [Task] Improve handling of manually set preset and association settings

As mentioned in #1015 there is need to better handle and improve the UX when a `preset` or `association` is manually set.

JimiC:
***Fixes #1019***

**Things I've done:**


JimiC:
@robertohuertasm @jens1o @ksvitkovsky @g-plane @iFun @mariotsi @sabinrimni 
Please provide translation for `configuration.dontShowConfigManuallyChangedMessage.description` configuration property.

robertohuertasm:
Si 'true', no se mostrará el mensaje 'reiniciar' al cambiar manualmente las configuraciones.

jens1o:
Wenn auf 'true' gesetzt, wird bei manuellen Änderungen an der Konfiguration keine "Neu starten"-Nachricht angezeigt.

ksvitkovsky:
Если установлено значение true, сообщение с просьбой перезапуска больше не будет отображаться при ручном изменении настроек

src/utils/index.ts
```diff
@@ -100,3 +100,20 @@ export function getDrives(...paths: string[]): string[] {
...
+    const hasKeys = !!Object.keys(value).length;
+    return !isArray && !isBuffer && isΟbject && hasKeys;
+  };
+  return Object.assign({}, ...function _flatten(child, path = []) {
+    return [].concat(...Object.keys(child)
+      .map(key => isValidObject(child[key])
+        ? _flatten(child[key], path.concat([key]))
+        : { [path.concat([key]).join(separator)]: child[key] }));
+  }(object));
+}
```
robertohuertasm: Even at risk of being less terse, I don't really like using a named function expression recursively using an IIFE with spread operators. I think that, although correct, this syntax is really convoluted and something like this will improve readability:

```ts
  const _flatten = (child, path = []): object[] => {
    return [].concat(...Object.keys(child)
      .map(key => isValidObject(child[key])
        ? _flatten(child[key], [...path, key])
        : { [[...path, key].join(separator)]: child[key] }));
  };

  return Object.assign({}, ..._flatten(obj));

```

Note that I've modified some of the `[].concat` with `spread operators` and only left the strictly needed ones.

Another suggestion would be to change the `object parameter` with `obj` or something similar to avoid confusion with the `object` ts keyword.

Finally, I would also add the return type annotation to the `_flatten` function in order to make it more clear.

robertohuertasm:
As another note, I was expecting a button (like in the Project Detection feature) that would change my user settings automatically. I mean, now the only way to disable this feature is to manually add the configuration. It would be nice to have the option in the same prompt that asks you for a restart. What do you think?

g-plane:
如果

g-plane:
如果设置成 true，手动修改配置后不提醒「重启后生效」

JimiC:
>As another note, I was expecting a button (like in the Project Detection feature) that would change my user settings automatically. I mean, now the only way to disable this feature is to manually add the configuration. It would be nice to have the option in the same prompt that asks you for a restart. What do you think?

The concept was to use the existing message and the underlying functionality to achieve the implementation of this feature. Although I thought about it, I didn't investigate it further. I'll have a look if this can be achieved without heavy refactoring.

src/utils/index.ts
JimiC: My personal coding preference is to achieve the most functionality with less code. Optimize performance but also optimize code lines.

I do though understand your need to simplify the code, to achieve better readability and make the code more appealing to new developers.

robertohuertasm: Thanks for the change! 

Note that you have sacrificed both lines of code and performance in this same piece of code in the `isValidObject` function: you haven't take profit of the shorcircuit abilities of the `&&` operator for the sake of readability. 😜

JimiC: Indeed. I'm in a constant struggle between condensing the code and keeping it readable.

JimiC:
Disabling button added.

JimiC:
We are now only missing the Italian translation to complete this feature.

sabinrimni:
Se 'true', il messaggio di 'riavvio' non verrà mostrato quando si cambiano le configurazioni manualmente.

JimiC:
Speak of the devil. lol

sabinrimni:
Yeah 😸 ...sorry for the delay just saw it this morning

JimiC:
@robertohuertasm Have a look at the proposed changes to the documentation of README and adjust accordingly if you like so.

README.md
```diff
@@ -52,12 +52,13 @@ If you're willing to collaborate with us feel free to join our [Github repositor
 ---
 
 ## State of the extension
+With the release of **7.10.0**, we improved the way the extension is handling your manually changes to a `presets` or `associations` configuration. By default, every time you change any of the aforementioned configurations, you will be presented with the message to `Restart` the editor, for the changes to take effect. Of course, you can always disable this behavior. See the [Configuration](https://github.com/vscode-icons/vscode-icons#configuration) section for more details.
```
robertohuertasm: `manual` instead of `manually`

JimiC: Good catch. I was rewriting that several times and obviously missed some parts.

lumaxis:
@JimiC Thanks for doing this! 👏 🎉  Very happy to see my suggestion get picked up and land in `7.10.0` this morning 🙂

PR #1040, opened 2017-06-25
7.10.0

robertohuertasm:


README.md
```diff
@@ -269,6 +269,21 @@ Once you have created the folder you can put all your custom icons there. But, a
...
 See [here the supported file extensions for an icon](https://github.com/vscode-icons/vscode-icons/blob/master/src/models/extensions/fileFormat.ts).
 
+#### Optional Angular icons
+Since **7.10.0**, along with the usual `Angular` icons, we shipped some optional icons for users wanting to have some specific icons for `component templates` and `component styles`.
+
+In order to enable them just add this to your settings:
+
+```js
+
+{ icon: 'ng_component_html', extensions: ['component.html', '.component.htm'], format: FileFormat.svg, disabled: true },
```
JimiC: Disable needs to be false.

JimiC: And format better be 'format: "svg"'

JimiC: Dot not required in front of extension declaration.

robertohuertasm: I've already changed the format and also the single quotes.

Disabled is true to be consistent with normal angular declarations. It will work no matter the value of the property.

README.md
```diff
@@ -269,6 +269,21 @@ Once you have created the folder you can put all your custom icons there. But, a
...
 See [here the supported file extensions for an icon](https://github.com/vscode-icons/vscode-icons/blob/master/src/models/extensions/fileFormat.ts).
 
+#### Optional Angular icons
+Since **7.10.0**, along with the usual `Angular` icons, we shipped some optional icons for users wanting to have some specific icons for `component templates` and `component styles`.
+
+In order to enable them just add this to your settings:
+
+```js
+
+{ "icon": "ng_component_html", "extensions": ["component.html", ".component.htm"], "format": "svg", "disabled": true },
```
JimiC: Again. In order to enable them "disabled" should be "false" or not declared at all.

JimiC: And the dot in front of the extensions declaration should be removed.

jens1o: There is no dot before the extensions property? And I think it's better to say it's disabled, otherwise we get tons of issues why it does not work because "it is not disabled by default"?

JimiC: Am I the only one that sees it? ".component.htm"

Disabled can't be false when you are saying how to enable them.

robertohuertasm: As I said, it's just for the sake of consistency. It works no matter the value. You can try it if you want.

JimiC: I'm not talking about the functionality. The text contradicts with the settings example. From a user's perspective I would have wondered how am I enabling them if I set them as disabled?

robertohuertasm: Removed the dot and added a note into the readme explaining why we keep this syntax.

PR #1043, opened 2017-06-26
Fix restart loop issue with manual config changes detection.

### LINKED ISSUE #1042: The extensions is always asking for a Restart when having a workspace customization
VSCode: 1.13.1 & 1.14.0-insider

Since 7.10.0 has been published it seems that we introduced some kind of bug and the extension is always asking me for a restart when I'm working in my Angular project. 

This doesn't happen in any other kind of projects.

I think the project detection system is messing with the `manual change detection` in the settings.

For the moment, I have to completely disable the new feature 😢 

It's strange because I never observed this behavior when developing or testing.

JimiC:
***Fixes #1042***

**Changes proposed:**
* [x] Fix

**Things I've done:**
* [x] My pull request fixes an issue, I referenced the issue

### DIFF
--- a/src/commands/index.ts
+++ b/src/commands/index.ts
@@ -10,7 +10,7 @@ import { folderIconsDisabled, iconsDisabled, manageApplyCustomizations } from '.
 import * as helper from './helper';
 
 const i18nManager = new LanguageResourceManager(vscode.env.language);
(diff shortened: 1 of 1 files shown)


JimiC:
The issue was caused because the manual config changes detection doesn't take into account the merged config settings. My original intention was to include that (after it was merged) but forgot to do so with all the fuzz going around.

PR #1062, opened 2017-07-08
Add flat version of PHP logo.

### LINKED ISSUE #1056: [Icon request] Alternative php icon
As you have this already for JS, JSON and some other file types I think, you have the official icon and alternatives for those so it would be cool if we can have a second alternative PHP icon which looks more modern. The official looks old, the elephant ks funny but I prefer only a light purle PHP like for JS, C, C++, C# etc.

Regards

Chris

JimiC:
***Fixes #1056***

**Changes proposed:**
* [x] Add

**Things I've done:**

This is yet another logo version for PHP. This on is a flat version for those wanting a minimalistic version.


JimiC:
Personally, I'm not in favor of adding this into the toggling mechanism.

PR #1064, opened 2017-07-09
Update dependencies

JimiC:
Updating dependencies, excluding `@types/node`, until releases reaches stability.

Also, fixes issue with using `valudio` as `npm` package registry.
@robertohuertasm we should avoid using it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,17 +2,18 @@
     "name": "vscode-icons",
     "version": "7.10.1",
     "lockfileVersion": 1,
+    "requires": true,
     "dependencies": {
         "@types/chai": {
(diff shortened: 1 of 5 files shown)


jens1o:
Are we sure we're releasing in more than a week?

JimiC:
We better merge this PR now as it contains fixes to the broken build, updates to `tslint` and `ts` as well, which improves our code quality.

robertohuertasm:
Sorry, I use a private npm as default and I didn't take it into account when I last updated the `package-lock.json` file.

JimiC:
The issue was with yarn not so much with npm.

PR #1067, opened 2017-07-13
Add white background to Erlang icon

mkaput:
Add white background to Erlang icon, so it looks much better with dark themes.

Before: 
zrzut ekranu z 2017-07-13 18-05-34

After: 
zrzut ekranu z 2017-07-13 18-05-17


mkaput:
I think this PR won't gain any traction as Erlang isn't popular enough, especially among VSCode users ;) 

I thought about negative version, and although it would be much more readable, the "e" would still be dark and wouldn't match original logo style. 

Other approach would be to borrow project icon from intellij erlang: https://github.com/ignatov/intellij-erlang/blob/master/resources/icons/erlang-small-16.png, but personally I don't like it

jens1o:
I can't decide...

JimiC:
@mkaput Our art guidelines dictates to use the official logo, which is the one we use (the `e` is transparent). 
Another guideline is to use icons that are not so much distracting to the user's eye. The purpose of the icons is to give the user a sense of what the file is about, without looking at the file extension.
Of course, we had cases where we deviated from our guidelines whenever it made sense to change the icon.

Now, in your case I see the following options:
1. The team accepts your PR.
2. The team agrees to use a revert color version (which is rarely used).
3. Use your suggested icon with the [custom icons feature](https://github.com/vscode-icons/vscode-icons#fine-tuning).

robertohuertasm:
@mkaput, first of all,  thanks for the PR.

To be honest, I'm a bit conflicted here. I agree with @JimiC's point of view. We have more or less used the official logos if they were available in most of the occasions.

In my mind, I only contemplate 1 and 3 options (from previous @JimiC's message).

Reason for 3 is pretty obvious. In this case I don't see it too difficult to distinguish the erlang icon. Although I agree that maybe it should be improved I don't feel that the original icon is so unreadable.

Reason for 1 is that, although the icon is indeed transparent, the official website has a white background as you can see here:

image

Even the favicon is white:

image

That made me think that maybe the original idea is that the background should be white and they have made the original logo transparent to make it easy to change the background if needed.

JimiC:
Logo with a black background. http://www.xappsoftware.com/wordpress/2015/08/11/erlang-setup-and-hello-world/
I believe that logo was designed with transparency to fit any background.

mkaput:
I wouldn't assume that one usage of Erlang logo on non-white background on not Erlang/Ericsson related website means that logo was designed to be used with transparency.

I couldn't either find official place where I could download the logo nor any guidelines, as they just do not exist (https://bugs.erlang.org/browse/ERL-41), but for example given Erlang homepage we can see that the logo in header has explicit white background (http://www.erlang.org/img/erlang.png).

But honestly I don't see any sense in arguing whether white background is part of logo. The problem is that in my opinion this icon is pretty invisible unless you use light theme and I believe this aspect should be beyond any logo-related guidelines.

JimiC:
As we currently don't see eye to eye with your opinion, I suggest you resolve to option 3 of my above comment.

JimiC:
FYI we got the logo from [here](https://commons.wikimedia.org/wiki/File:Erlang_logo.svg)

robertohuertasm:
I've tested the icon in all the dark themes and I prefer the transparent one over the white one. But that's just me. I don't find it difficult to be identified, honestly, so I also vote for 3.

Let's keep this PR opened for a while to see if anyone else reacts to it.

robertohuertasm:
I've added the white logo as a custom icon. @JimiC feel free to review this.

mkaput:
Thanks!

PR #1072, opened 2017-07-15
Refactor configuration changes detection.

JimiC:
Simplifying code, according to https://lodash.com/docs/4.17.4#isEqual, it can be used to determine configuration objects equality.

### DIFF
--- a/src/init/applyCustomizationsManager.ts
+++ b/src/init/applyCustomizationsManager.ts
@@ -35,13 +35,7 @@ function configChanged(prevConfig, currentConfig) {
       : prevConfig[key];
     const parts = key.split('.').filter(x => x !== 'vsicons');
     const newValue = parts.reduce((prev, current) => prev[current], currentConfig);
-    const cond1 = Array.isArray(oldValue)
-                  && Array.isArray(newValue)
-                  && (newValue.length !== _.intersectionWith(oldValue, newValue, _.isEqual).length);
-    // this is to equal null == undefined as vscode doesn't respect null defaults
-    // tslint:disable-next-line triple-equals
-    const cond2 = !Array.isArray(oldValue) && oldValue != newValue;
-    if (cond1 || cond2) {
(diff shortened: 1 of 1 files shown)


JimiC:
@robertohuertasm Couldn't figured out why you chose `intersectionWith`.

src/init/applyCustomizationsManager.ts
```diff
@@ -35,13 +35,7 @@ function configChanged(prevConfig, currentConfig) {
...
     const parts = key.split('.').filter(x => x !== 'vsicons');
     const newValue = parts.reduce((prev, current) => prev[current], currentConfig);
-    const cond1 = Array.isArray(oldValue)
-                  && Array.isArray(newValue)
-                  && (newValue.length !== _.intersectionWith(oldValue, newValue, _.isEqual).length);
-    // this is to equal null == undefined as vscode doesn't respect null defaults
-    // tslint:disable-next-line triple-equals
-    const cond2 = !Array.isArray(oldValue) && oldValue != newValue;
-    if (cond1 || cond2) {
+    if (!_.isEqual(oldValue, newValue)) {
```
jens1o: why did you used lodash here instead of plain js?

JimiC: `lodash` was already in use from @robertohuertasm.

jens1o: Personally, I would use vanilla js, but I don't want to be the blocking person...

JimiC: `_.isEqual` is not the same as `===`. Check the link in the PR description.

jens1o: I see. I retract my suggestion.

robertohuertasm: Please, see #1075 to understand why I used `intersectionWith`

JimiC: Makes sense.

PR #1074, opened 2017-07-16
Fix opening urls.

JimiC:
URLs don't reside in `i18n` manager anymore. My bad for not correcting the code in the first place.


JimiC:
While at it I'd like to propose to change the welcome message and introduce an `Activate` button. Thoughts?

jens1o:
@JimiC Technically, we can do that. Because then we modify the user settings, and boom, we're activated... But I don't know whether this fits ethically...

Okay. As long as we're sticking to that question, it's fine for me, when thinking about that there are well-watched videos on YouTube on how to activate our theme, where I think... wut?

robertohuertasm:
@JimiC, I think [material icons](https://marketplace.visualstudio.com/items?itemName=PKief.material-icon-theme) already does that:
image

I guess we could follow the lead.

JimiC:
@robertohuertasm Exactly what I saw and came up with the proposal.

robertohuertasm:
note that they also add a new setting:
image

It's ok for me.

robertohuertasm:
@JimiC, I accept this PR and you'll do that in another one?

JimiC:
This PR and the proposal are diff things.

JimiC:
@robertohuertasm You can accept this any time now.

PR #1075, opened 2017-07-16
test verifying altered order won't trigger config changes

robertohuertasm:
Related to #1072.

If we introduce the proposed change then order alterations will trigger change detection. That's why `intersectionWith` was being used. I've added the test so we won't miss this next time.

### DIFF
--- a/test/init/applyCustomizationsManager.test.ts
+++ b/test/init/applyCustomizationsManager.test.ts
@@ -127,10 +127,10 @@ describe('ApplyCustomizations: tests', function () {
           { icon: "js", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
           { icon: "js2", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
         ];
-        const newConfig = { ...userConfig, associations: { ...userConfig.associations } };
-        newConfig.associations.files = [...userConfig.associations.files];
-        newConfig.associations.files.reverse();
-        manageApplyCustomizations(newConfig, userConfig, spy);
+        const initConfig = { ...userConfig, associations: { ...userConfig.associations } };
+        initConfig.associations.files = [...userConfig.associations.files];
+        userConfig.associations.files.reverse();
+        manageApplyCustomizations(initConfig, userConfig, spy);
         expect(spy.called).to.not.be.true;
       });
 
--- a/test/init/applyCustomizationsManager.test.ts
+++ b/test/init/applyCustomizationsManager.test.ts
@@ -130,8 +130,7 @@ describe('ApplyCustomizations: tests', function () {
         const newConfig = { ...userConfig, associations: { ...userConfig.associations } };
         newConfig.associations.files = [...userConfig.associations.files];
         newConfig.associations.files.reverse();
-        const initConfig = JSON.parse(JSON.stringify(newConfig));
-        manageApplyCustomizations(initConfig, userConfig, spy);
+        manageApplyCustomizations(newConfig, userConfig, spy);
         expect(spy.called).to.not.be.true;
       });
 
--- a/test/init/applyCustomizationsManager.test.ts
+++ b/test/init/applyCustomizationsManager.test.ts
@@ -120,6 +120,21 @@ describe('ApplyCustomizations: tests', function () {
         expect(spy.called).to.not.be.true;
       });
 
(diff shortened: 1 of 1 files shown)


test/init/applyCustomizationsManager.test.ts
```diff
@@ -120,6 +120,21 @@ describe('ApplyCustomizations: tests', function () {
...
+      function () {
+        const spy = sinon.spy();
+        userConfig.associations.files = [
+          { icon: "js", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
+          { icon: "js2", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
+        ];
+        const newConfig = { ...userConfig, associations: { ...userConfig.associations } };
+        newConfig.associations.files = [...userConfig.associations.files];
+        newConfig.associations.files.reverse();
+        const initConfig = JSON.parse(JSON.stringify(newConfig));
```
JimiC: I believe there is no need to stringify and parse here.

robertohuertasm: You're right. I just copied the line like a fool! 😂

test/init/applyCustomizationsManager.test.ts
```diff
@@ -120,6 +120,20 @@ describe('ApplyCustomizations: tests', function () {
...
+        const spy = sinon.spy();
+        userConfig.associations.files = [
+          { icon: "js", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
+          { icon: "js2", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
+        ];
+        const newConfig = { ...userConfig, associations: { ...userConfig.associations } };
+        newConfig.associations.files = [...userConfig.associations.files];
+        newConfig.associations.files.reverse();
+        manageApplyCustomizations(newConfig, userConfig, spy);
+        expect(spy.called).to.not.be.true;
```
JimiC: The function signature dictates `oldConfig`, `newConfig`.

robertohuertasm: done

JimiC:
Why do I have the sense that 
```
 userConfig.associations.files = [
          { icon: "js", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
          { icon: "js2", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
        ];
        const initConfig = { ...userConfig, associations: { ...userConfig.associations } };
        initConfig.associations.files = [...userConfig.associations.files];
        userConfig.associations.files.reverse();
```
can be written as 

```
userConfig.associations.files = [
          { icon: "js", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
          { icon: "js2", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
        ];
        const newConfig = { ...userConfig };
        newConfig.associations.files.reverse();
```
or will it reverse also the `userConfigassociations.files`?

robertohuertasm:
Yes, it will keep the same references so you cannot do that.

JimiC:
So `{ ...userConfig }` is a ref copy instead of cloning.

robertohuertasm:
You're right.

JimiC:
I just remembered why I did `stringify` and `parse`. It's the vanilla js for cloning.

robertohuertasm:
I also use this for cloning huge objects.

JimiC:
So the test can be written as 
```
        userConfig.associations.files = [
          { icon: "js", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
          { icon: "js2", extensions: ["myExt1", "myExt2.custom.js"], format: "svg" },
        ];
        const initConfig = JSON.parse(JSON.stringify(userConfig));
        userConfig.associations.files.reverse();
        manageApplyCustomizations(initConfig, userConfig, spy);
```
to keep uniformity with the others.

robertohuertasm:
I've already added it... 

To be honest, I favor the new syntax over the `JSON.parse` trick. I only clone like this when I really need to clone the whole object which in this case... could make sense 😉 but I stick to the new ways when only having to modify a single property. 

I leave it up to you to you if you want to include the change in #1074.

JimiC:
The thing is that with stringify-parse you achieve the same result with fewer code lines which fall into my coding guidelines (we have already discussed this in another case). I think I'm gonna make the change.

PR #1084, opened 2017-07-22
Add 'activate' button on welcome message.

### LINKED ISSUE #1083: [Task] Add Activate button on new installation
Add an `Activate` button when extension gets installed for the first time.

JimiC:
***Fixes #1083***

**Changes proposed:**
* [x] Add

**Things I've done:**

This PR does the following:
- Adds an `Activate` button in the welcome message.
- Removes the message about the activation path for each platform.
- Reshows the welcome message if the user selects any option other than `Activate` or `Close`.
- Adds a new command in the commands list to activate the icon theme.
- Handles the manual definition of the icon theme in settings.

I've done also some noticeable refactorings:
- Moves some hard coded strings, that repeat several times in the code, to `constants`.
- Removes redundant keys and adds new ones in `LangResourceKeys`.
- Changes the logic of when `status` gets set.
- Renames status `notInstalled` to `notActivated`.


JimiC:
Just want to point out that I haven't found so far a way to detect the change, when the extension gets disabled, in order to change the status of the extension. Not that it matters at this moment, as we don't rely on the status for any flow logic.

JimiC:
Additionally, there is a `vscode` issue that is reproducible with every icon theme extension that has an `Activate` command.
Reproduction steps:
1. Install the extension.
2. Activate extension from commands list.
3. Open user settings.
4. Manually remove the icon theme setting.
5. Set the setting again manually (or paste it in)
6. Icon theme doesn't get activated.

package.nls.de.json
```diff
@@ -1,4 +1,5 @@
 {
+  "command.activateIcons.title": "",
```
jens1o: `Die Icons von VSCode Icons nutzen.`

somewhat like this?

JimiC:
@jens1o Thanks for reminding.

New translations are required for every `src/i18n/langResources/lang.*.ts` and `package.nls.*.json`

Respected contributors, please provide the translation for: 
German: @jens1o: 
Spanish: @robertohuertasm 
Italian: @sabinrimni 
Russian: @ksvitkovsky 
Chinese (Simplified): @g-plane

package.nls.de.json
JimiC: According to `vscode` translation for `Icon Theme` I'm think something in the line of `Dateisymboldesign aktivieren`.

JimiC: And for Spanish, I'm thinking something like `Activar el tema de icono` but I'm leaving that to @robertohuertasm.

jens1o: `Dateisymboldesign ` is really shitty. We say `icons` really natively... I don't know why the translators at transifix like to use this "word"... 🤢

src/i18n/langResources/lang.de.ts
```diff
@@ -1,17 +1,13 @@
...
   seeReadme: 'Erfahren Sie mehr über diese Erweiterung (Englisch)',
-  welcomeBegin: 'vscode-icons benutzt nun die offizielle API. Gehen Sie zu',
-  welcomeEnd: 'und wählen Sie VSCode Icons, um zu aktivieren.',
-  activationPath: {
-    darwin: 'Code > Einstellungen > Datei-Symboldesign',
-    linux: 'Datei > Einstellungen > Datei-Symboldesign',
-    win32: 'Datei > Einstellungen > Datei-Symboldesign',
-  },
+  welcome: `${constants.extensionName} benutzt nun die offizielle API.`,
+  activate: '',
```
jens1o: `aktivieren`

package.nls.de.json
jens1o: In fact, Duden(you may know that) also says that `Icon` is German aswell. http://www.duden.de/rechtschreibung/Icon

Dateisymboldesign is a neologism...

JimiC: I'm leaving that up to you as well.

jens1o: I'm going to suggest it to them...

src/i18n/langResources/lang.it.ts
```diff
@@ -1,17 +1,13 @@
...
   seeReadme: 'Sapere di più sull\'estensione',
-  welcomeBegin: 'vscode-icons ora usa la API ufficiale. Vai a',
-  welcomeEnd: 'e seleziona VSCode Icons per attivarla.',
-  activationPath: {
-    darwin: 'Code > Preferenze > Tema Icona File',
-    linux: 'File > Preferenze > Tema Icona File',
-    win32: 'File > Preferenze > Tema Icona File',
-  },
+  welcome: `${constants.extensionName} ora usa la API ufficiale.`,
+  activate: '',
```
sabinrimni: `activate: 'Attiva'`

package.nls.it.json
```diff
@@ -1,4 +1,5 @@
 {
+  "command.activateIcons.title": "",
```
sabinrimni: And I suppose something like this here:
`"command.activateIcons.title": "Attiva tema icone VSCode"`

JimiC: Thanks.

JimiC: @mariotsi Are you OK with the translation.

mariotsi: Seems ok to me

package.nls.json
```diff
@@ -1,4 +1,5 @@
 {
+  "command.activateIcons.title": "Activate VSCode Icon Theme",
```
robertohuertasm: I'm not quite sure this is a correct text. It suggest to activate the theme, but which theme? I mean, if you have 2 or 3 themes this doesn't identify our theme. 

Our theme is called `vscode-icons`, so I guess something like `Activate vscode-icons Icon Theme` would be a more suitable.

What do you think?

JimiC: Well, the thing is that the label says `VSCode Icons` and that's what the extension is known for.

JimiC: Pop up the icon themes list with `F1` and you'll my point.

robertohuertasm: Ok, then it should be `VSCode Icons` not `VScode Icon`.

robertohuertasm: To be more specific, `Activate VSCode Icons Icon Theme`

JimiC: If you wonna be precise then it should be `VSCode Icons Icon Theme`.

JimiC: Look @ https://github.com/vscode-icons/vscode-icons/pull/1074#issuecomment-315618707

JimiC: And in our case the command will show `Icons: Activate VSCode Icon Theme`. `Icons` is already mentioned implicitly.

robertohuertasm: What do you mean by that? They rely on the prefix. I would stick to `Activate Icon Theme` or `Activate VSCode Icons` or `Activate VSCode Icons Icon Theme`.

JimiC: THe alternative is to change the label to `vscode-icons` as well and dispute gets solved.

robertohuertasm: I would use the same solution that `Material Icons` uses: `Activate Icon Theme`. No confussion and we stick to the same convention.

JimiC: Yeah, that's the thing. `Material Icons` uses its label name in the command. We use just `Icons` where we should had used `VSCode Icons` or `vscode-icons` if you prefer.

JimiC: Honestly, I was troubled too. Took me some time to decide but I believe I chose the best solution in our case.

robertohuertasm: We used `Icons` because, as you know, this was the only icon set available when VSCode didn't even support the feature. I don't have in mind changing this, to be honest. I think that this positions us as the main reference in terms of icons. Even more, people may be used to this shortcut and I wouldn't like to move their cheese.

If we want to use the name in this command, then use it fully or not use it at all, which is indeed what the current message is doing. The thing is that using `VSCode` leads to confusion with the name of the icon set and people may think that there's an error (I did myself) with `VSCode Icon` missing an `s`.

JimiC: I'm with you. SO let's settle to `Activate VSCode Icons`. Short and simple.

robertohuertasm: 👍

package.nls.es.json
```diff
@@ -1,4 +1,5 @@
 {
+  "command.activateIcons.title": "",
```
robertohuertasm: Activar VSCode Icons

src/i18n/langResources/lang.es.ts
```diff
@@ -1,17 +1,13 @@
...
   seeReadme: 'Sobre esta extensión (en inglés)',
-  welcomeBegin: 'vscode-icons ahora usa la API official. Vaya a',
-  welcomeEnd: 'y seleccione VSCode Icons para activar la extensión',
-  activationPath: {
-    darwin: 'Code > Preferencias > Tema de iconos de archivo',
-    linux: 'Archivo > Preferencias > Tema de iconos de archivo',
-    win32: 'Archivo > Preferencias > Tema de iconos de archivo',
-  },
+  welcome: `${constants.extensionName} ahora usa la API official.`,
+  activate: '',
```
robertohuertasm: Activar

jens1o:
You may want to include my suggestions.

JimiC:
Yeah. We didn't decide on the command title. Will it be `Aktivieren VSCode Icons`?

jens1o:
It should be `VSCode Icons aktivieren`. 😉

src/i18n/langResources/lang.zh-cn.ts
```diff
@@ -1,17 +1,13 @@
...
   seeReadme: '了解关于此插件的更多信息',
-  welcomeBegin: 'vscode-icons 现在支持官方 API，可前往',
-  welcomeEnd: '并选择 VSCode Icons',
-  activationPath: {
-    darwin: 'Code > 首选项 > 文件图标主题',
-    linux: '文件 > 首选项 > 文件图标主题',
-    win32: '文件 > 首选项 > 文件图标主题',
-  },
+  welcome: `${constants.extensionName} 现在支持官方 API`,
+  activate: '',
```
JimiC: Thanks.

JimiC:
@robertohuertasm I'm gonna ask you to write the appropriate text in `README` of this enhancement for the release. Thanks in advance.

PR #1099, opened 2017-07-26
Add types for semver.

JimiC:
This is a proposal PR.


JimiC:
It turns out that we don;t use anywhere the `vscode` version variable. Maybe we should consider deprecating it.

PR #1169, opened 2017-09-13
fixing tslint configuration

robertohuertasm:
Fixes 'tslint.json' file and a few `double quotes`. Several minor package updates.


package.json
```diff
@@ -216,7 +216,7 @@
         "precompile": "rimraf ./out",
         "compile": "tsc",
         "compile:w": "npm run compile -- -w",
-        "lint": "tslint --project tsconfig.json",
+        "lint": "tslint --project tsconfig.json --type-check true",
```
JimiC: I believe `true`  is not necessary

tslint.json
```diff
@@ -13,6 +13,7 @@
       "allow-pascal-case",
       "allow-leading-underscore"
     ],
+    "no-submodule-imports": false,
```
JimiC: If we are going to use the setting then better set it to true and exclude `vscode`. 
i. e. `no-submodule-imports: [true, "vscode"]`

robertohuertasm: I'm used to have this enabled from my work with `Angular` and `rxjs` but given that here we don't use `webpack` and we don't do any kind of `tree-shaking` I can live with it 😉

JimiC: Fine.

robertohuertasm:
Wait to approve it as I'm also updating the Changelog.

robertohuertasm:
Note that I updated the changelog with September 17 as the new release date but we can advance it if you are ok with it.

package.json
robertohuertasm: Not really, you're right.

PR #1182, opened 2017-09-23
Improve templates.

JimiC:
Proposed changes to templating.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,23 +1,27 @@
-_**Instruction Template**_
+# Issue Template
 
-**THIS IS AN EXAMPLE TEMPLATE**
+THIS IS AN EXAMPLE TEMPLATE
 
-**Make sure the following boxes are checked (and the placeholders replaced) *BEFORE* submitting this issue**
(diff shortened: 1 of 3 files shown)


JimiC:
Because you will wonder, I have removed any markdown syntax from lines that are getting deleted, because they don't play any role. Those that are left are purely for making `markdownlint` stop complaining.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,23 +1,27 @@
...
+Please remove any unused content (including these instructions) before submitting your issue.
 
-***IMPORTANT:* For folder icon requests please first see issue #650**
+IMPORTANT: For folder icon requests please first see issue: [650](https://github.com/vscode-icons/vscode-icons/issues/650)
 
-**Thanks :smile:**
+Thanks.
 
-**Checklist:**
+Note: Always remove this line and all above.
```
jens1o: this is already stated two lines before.

JimiC: You can never be certain that the user read through the entire instructions and understood them. Like apps, instructions need to be idiot proof.

PR #1183, opened 2017-09-23
Add ability to retrieve project info (version).

JimiC:
**Changes proposed:**
* [x] Prepare

Gives the ability to PAD to act depending on project version. Will come handy supporting `AngularJS` projects.


test/init/projectAutoDetection.test.ts
```diff
@@ -118,9 +118,57 @@ describe('AutoDetectProject: tests', function () {
...
+
+        afterEach(() => {
+          sandbox.restore();
+        });
+
+        it('the project info if it matches',
+          function () {
+            sandbox.stub(fs, 'readFileSync').returns('{ "dependencies": { "@angular/core": "0.0.0" } }');
+            const results = [{ fsPath: '' }] as models.IVSCodeUri[];
+            const projectInfo: models.IProjectInfo = adp.getProjectInfo(results, 'ng');
```
robertohuertasm: Why are we using different keys (ng & Angular) for `Angular`? I think we should unify these identifiers.

src/init/projectAutoDetection.ts
```diff
@@ -76,14 +76,37 @@ function getIconManifest(): string {
...
+    }
+    const devDepExists = (projectJson.devDependencies && (projectJson.devDependencies[key] != null));
+    if (devDepExists) {
+      return { name: infoName, version: projectJson.devDependencies[key] };
+    }
+    return null;
+  };
+
   switch (name) {
     case 'ng':
```
robertohuertasm: I think we should unify these identifiers (`ng` with `angular`) so we could avoid this mapping

src/models/projectDetector/projectInfo.ts
```diff
@@ -0,0 +1,4 @@
+export interface IProjectInfo {
+  name?: string;
```
robertohuertasm: Instead of using a plain string I think we could use a union type of string literal types or an enum.

test/init/projectAutoDetection.test.ts
JimiC: The key is `ng` and `Angular` is an optional name for the project info. Currently, project info name is not used anywhere but, you know me, I thought it was nice to have.

src/models/projectDetector/projectInfo.ts
JimiC: You mean like an enum of project names?

i.e.
```
enum Projects {
  none: ''
  angular: 'ng',
  angularjs: 'ngjs'
}
```

robertohuertasm: yes, that or this:
```ts
export type Project = 'ng' | 'ngjs' | 'react';
// and then

interface IProjectInfo {
name?: Project;
``` 

I don't have any preference.

PS: In case anyone is not familiar with `string literal types` you can have more info [here](https://basarat.gitbooks.io/typescript/docs/types/literal-types.html)

JimiC: I'm familiar but I think `enum` gives more flexibility. Will try them and see what fits best.

JimiC:
@robertohuertasm Is that what you had in mind?

src/commands/index.ts
```diff
@@ -102,27 +102,27 @@ function resetProjectDetectionDefaultsCommand(): void {
 }
 
 function toggleAngularPresetCommand(): void {
-  togglePreset('angular', 'ngPreset', false, false);
+  togglePreset(models.PresetNames[models.PresetNames.angular], 'ngPreset', false, false);
```
robertohuertasm: In order to make the first parameter less verbose why not to pass `models.PresetNames.xxx` and resolve it to string inside `togglePreset`?

src/models/projectDetector/projectInfo.ts
robertohuertasm: I think this should be a `Projects` enum

src/init/projectAutoDetection.ts
```diff
@@ -76,14 +76,37 @@ function getIconManifest(): string {
   }
 }
 
-export function isProject(projectJson: any, name: string): boolean {
+export function getProjectInfo(results: models.IVSCodeUri[], name: string): models.IProjectInfo {
```
robertohuertasm: Name should be a `Projects` enum

src/init/projectAutoDetection.ts
```diff
@@ -76,14 +76,37 @@ function getIconManifest(): string {
...
+  results.forEach(result => {
+    const content = fs.readFileSync(result.fsPath, 'utf8');
+    const projectJson = parseJSON(content);
+    projectInfo = getInfo(projectJson, name);
+    if (!!projectInfo) { return; }
+  });
+  return projectInfo;
+}
+
+export function getInfo(projectJson: any, name: string): models.IProjectInfo {
```
robertohuertasm: Same thing here.

src/init/projectAutoDetection.ts
```diff
@@ -76,14 +76,37 @@ function getIconManifest(): string {
...
+  if (!projectJson) { return null; }
+
+  const getInfoFn = (key: string, infoName?: string): models.IProjectInfo => {
```
robertohuertasm: Same for `infoName`

src/init/projectAutoDetection.ts
```diff
@@ -76,14 +76,37 @@ function getIconManifest(): string {
...
+      return { name: infoName, version: projectJson.dependencies[key] };
+    }
+    const devDepExists = (projectJson.devDependencies && (projectJson.devDependencies[key] != null));
+    if (devDepExists) {
+      return { name: infoName, version: projectJson.devDependencies[key] };
+    }
+    return null;
+  };
+
   switch (name) {
```
robertohuertasm: if `name` is a `Projects` enum, then `getInfoFn` function can be simplyfied by not needing the second parameter at all as it can be accesed via closure.

src/commands/index.ts
JimiC: OK.

PR #1184, opened 2017-09-23
We no longer check if the extension is in use anymore

robertohuertasm:
**_Fixes #1176_**

**Changes proposed:**

- [ ] Add
- [ ] Prepare
- [ ] Delete
- [x] Fix

**Things I've done:**

Translations are still missing as I changed the welcome message.

### DIFF
--- a/src/init/welcome.ts
+++ b/src/init/welcome.ts
@@ -10,8 +10,9 @@ const i18nManager = new LanguageResourceManager(vscode.env.language);
 
 export function manageWelcomeMessage(settingsManager: ISettingsManager): void {
   const extensionVersion = settingsManager.getSettings().extensionSettings.version;
+  const themeName = getConfig().inspect(constants.vscode.iconThemeSetting).globalValue;
 
-  if (!settingsManager.getState().welcomeShown) {
+  if (!settingsManager.getState().welcomeShown && themeName !== constants.extensionName) {
     showWelcomeMessage(extensionVersion);
     return;
   }
--- a/src/init/welcome.ts
+++ b/src/init/welcome.ts
@@ -5,27 +5,27 @@ import { getConfig } from '../utils/vscode-extensions';
 import { ISettingsManager, LangResourceKeys } from '../models';
 import { constants } from '../constants';
 import { activationCommand } from '../commands';
+import { extensionSettings } from '../settings/extensionSettings';
 
 const i18nManager = new LanguageResourceManager(vscode.env.language);
 
 export function manageWelcomeMessage(settingsManager: ISettingsManager): void {
-  const extensionVersion = settingsManager.getSettings().extensionSettings.version;
   const themeName = getConfig().inspect(constants.vscode.iconThemeSetting).globalValue;
 
   if (!settingsManager.getState().welcomeShown && themeName !== constants.extensionName) {
-    showWelcomeMessage(extensionVersion);
+    showWelcomeMessage();
     return;
   }
 
   if (settingsManager.isNewVersion() && !getConfig().vsicons.dontShowNewVersionMessage) {
-    showNewVersionMessage(extensionVersion);
+    showNewVersionMessage();
   }
 }
 
-function showWelcomeMessage(extensionVersion: string): void {
+function showWelcomeMessage(): void {
   const displayMessage = () => {
     vscode.window.showInformationMessage(
-      `${i18nManager.getMessage(LangResourceKeys.welcome)} v${extensionVersion}`,
+      `${i18nManager.getMessage(LangResourceKeys.welcome)} v${extensionSettings.version}`,
       { title: i18nManager.getMessage(LangResourceKeys.activate) },
       { title: i18nManager.getMessage(LangResourceKeys.aboutOfficialApi) },
       { title: i18nManager.getMessage(LangResourceKeys.seeReadme) })
@@ -56,9 +56,9 @@ function showWelcomeMessage(extensionVersion: string): void {
   displayMessage();
 }
 
-function showNewVersionMessage(extensionVersion: string): void {
+function showNewVersionMessage(): void {
   vscode.window.showInformationMessage(
-    `${i18nManager.getMessage(LangResourceKeys.newVersion)} v${extensionVersion}`,
+    `${i18nManager.getMessage(LangResourceKeys.newVersion)} v${extensionSettings.version}`,
     { title: i18nManager.getMessage(LangResourceKeys.seeReleaseNotes) },
     { title: i18nManager.getMessage(LangResourceKeys.dontShowThis) })
     .then(btn => {
--- a/src/init/welcome.ts
+++ b/src/init/welcome.ts
@@ -13,6 +13,7 @@ export function manageWelcomeMessage(settingsManager: ISettingsManager): void {
   const themeName = getConfig().inspect(constants.vscode.iconThemeSetting).globalValue;
 
   if (!settingsManager.getState().welcomeShown && themeName !== constants.extensionName) {
+    settingsManager.setWelcomeShownToTrue();
     showWelcomeMessage();
     return;
   }
--- a/src/init/welcome.ts
+++ b/src/init/welcome.ts
@@ -24,8 +24,7 @@ export function manageWelcomeMessage(settingsManager: ISettingsManager): void {
 
 function showWelcomeMessage(): void {
   const displayMessage = () => {
-    vscode.window.showInformationMessage(
-      `${i18nManager.getMessage(LangResourceKeys.welcome)} v${extensionSettings.version}`,
+    vscode.window.showInformationMessage(i18nManager.getMessage(LangResourceKeys.welcome),
       { title: i18nManager.getMessage(LangResourceKeys.activate) },
       { title: i18nManager.getMessage(LangResourceKeys.aboutOfficialApi) },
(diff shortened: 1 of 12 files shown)


src/init/welcome.ts
```diff
@@ -9,21 +9,22 @@ import { activationCommand } from '../commands';
 const i18nManager = new LanguageResourceManager(vscode.env.language);
 
 export function manageWelcomeMessage(settingsManager: ISettingsManager): void {
-  const themeName = getConfig().inspect(constants.vscode.iconThemeSetting).globalValue;
-  if (!settingsManager.getState().welcomeShown || themeName !== constants.extensionName) {
-    showWelcomeMessage();
```
JimiC: Consider the case where `welcomeShown` for any reason is set to `false` but `vscode-icons` is already set as the `theme` in settings (which means that it's already activated). I still believe the proper way is to switch the condition from disjunction to conjunction.

robertohuertasm: You've got a point!

JimiC:
Pardon me, but showing the version in the welcome message is excessive. We already have the version message for that. I believe the `welcome` message itself is sufficient enough.

src/init/welcome.ts
```diff
@@ -9,21 +9,23 @@ import { activationCommand } from '../commands';
 const i18nManager = new LanguageResourceManager(vscode.env.language);
 
 export function manageWelcomeMessage(settingsManager: ISettingsManager): void {
+  const extensionVersion = settingsManager.getSettings().extensionSettings.version;
```
JimiC: Isn't `settings.version` sufficient?

robertohuertasm:
I don't think it is excessive. We provide information about the product that they have installed, the same that we do when the extensions gets updated. I mean, I think it makes sense to inform the user about what's the version of the product. 

Anyway, this should be a one-time in extension's lifetime message.

JimiC:
> Anyway, this should be a one-time in extension's lifetime message.

Exactly why I believe it's excessive.

src/init/welcome.ts
JimiC: Discard that. It's late and my eyes are playing tricks on me.

robertohuertasm:
It's funny that the same fact is making us think totally opposite 😂 We'll have to untie this!

@jens1o what's your opinion? if you don't have one then bet it a coin flip!

JimiC:
@robertohuertas Don't get me wrong. I'm fine with it at the end if you strongly believe it should be this way.

robertohuertasm:
@JimiC, the same happens to me. Don't worry. I don't have really a strong point about this. It's just what came to me as something I would see fit. Similarly to what happened to you.

Let's wait for @jens1o to say what he thinks and let's discuss it again with another input.

robertohuertasm:
As we seem to agree in the messages I will really thank if our precious translators will help us again! ❤️ 

French: @MonsieurMan
German: @jens1o
Italian: @mariotsi @sabinrimni
Russian: @ksvitkovsky
Simplified Chinese: @iFun @g-plane

g-plane:
感谢安装 ${constants.extensionName}

jens1o:
Maybe we should have waited before we ask translators to translate phrases when we don't know whether this is going to be implemented at all. (and why you don't want to sleep at all?)

> @jens1o what's your opinion? if you don't have one then be it a coin flip!
Personally, I agree with @JimiC, because imagine: Every extension would initiate a welcome and a message about a new version.. That's pretty spoky to me.

jens1o:
Translation: `Danke für das Installieren von ${constant.extensionName}`

JimiC:
Just to rephrase myself. I don't see any added value in adding the version number to the message. Let's say I'm the user. I install the extension and it shows me the message with the version number. What does the number suppose to mean to me? Does it have a special meaning? That's my reasoning.

src/init/welcome.ts
```diff
@@ -5,25 +5,28 @@ import { getConfig } from '../utils/vscode-extensions';
...
+import { extensionSettings } from '../settings/extensionSettings';
 
 const i18nManager = new LanguageResourceManager(vscode.env.language);
 
 export function manageWelcomeMessage(settingsManager: ISettingsManager): void {
   const themeName = getConfig().inspect(constants.vscode.iconThemeSetting).globalValue;
-  if (!settingsManager.getState().welcomeShown || themeName !== constants.extensionName) {
+
+  if (!settingsManager.getState().welcomeShown && themeName !== constants.extensionName) {
+    settingsManager.setWelcomeShownToTrue();
```
JimiC: Can you explain the case for this?

robertohuertasm: Of course. Imagine this case:

- Some user install the extension.
- The sees the welcome message but he decides not to activate the extension yet.
- The closes the IDE.
- The user opens the IDE and receives again the welcome message.

Basically, this assures that this message is only shown once in the extension's lifetime.

Just to be clear, I decided to call for the translator friends because the message was not in discussion and the change was going to be implemented in a way or another. On the other hand, @jens1o, I'm under the impression that I didn't explain myself well enough as it was never intended to show two messages but only one with the version information in it. We all agree with one and only message, the thing that we're discussing is if we add the version to it, as we are already doing when any update happens. This is a one time message in extension's lifetime so no spoky things here! :wink: 

Finally, and following @JimiC reasoning about the utility that the version number has for the final user, then we could also apply this to the update message. It would have no meaning for the user either. I introduced this as a matter of consistency and because I feel that the version number can matter to some users.

Anyway, as I said yesterday, if you both feel that way I'm happy with our collective decision. I just wanted to present my case again for the last time.

robertohuertasm:
Just to be clear, I decided to call for the translator friends because the message was not in discussion and the change was going to be implemented in a way or another. On the other hand, @jens1o, I'm under the impression that I didn't explain myself well enough as it was never intended to show two messages but only one with the version information in it. We all agree with one and only message, the thing that we're discussing is if we add the version to it, as we are already doing when any update happens. This is a one time message in extension's lifetime so no spoky things here! :wink: 

Finally, and following @JimiC reasoning about the utility that the version number has for the final user, then we could also apply this to the update message. It would have no meaning for the user either. I introduced this as a matter of consistency and because I feel that the version number can matter to some users.

Anyway, as I said yesterday, if you both feel that way I'm happy with our collective decision. I just wanted to present my case again for the last time.

src/init/welcome.ts
JimiC: The thing is that the code flow already updates the status on first installation (https://github.com/vscode-icons/vscode-icons/blob/master/src/index.ts#L30). On first installation `settingsManager.isNewVersion()` is always `true`.

robertohuertasm: Let me check this out because this was not happening to me. Maybe because while testing I didn't changed the version number.

JimiC: And that's exactly the case. On first installation `getState()` return the default state and `version` is `0.0.0`.

robertohuertasm: You're absolutely right! It was because of the temp data file had the same version number. I'll roll it back.

JimiC:
You see the version number in the update message has a purpose as it informs the user which version is installed. 

Imagine we use in real life welcoming messages such as "Hi friend, thanks for coming to visit me at [street No]" (where 'street No' is your address). Your friend already knows your address.

robertohuertasm:
@JimiC, clearly we see this differently so it's no use to discuss it further. I'll get rid of the version info so we can move on to more important matters! 😉

jens1o:
Okay, please flip a coin for me. As far as I understood, we're discussing about whether in this new message, we'd like to show the actual version number. When that is true, I would only show that on major changes(`8.x.x`, `9.x.x`...)?

JimiC:
@jens1o Don't get me wrong but sometimes you make me laugh so hard. We are discussing to add or not a simple string and you are proposing to add more complexity (showing only major version is more complex than just showing the setting version). 😆

jens1o:
Now or never :D

src/init/welcome.ts
```diff
@@ -5,25 +5,26 @@ import { getConfig } from '../utils/vscode-extensions';
...
-    showNewVersionMessage(settingsManager);
+    showNewVersionMessage();
   }
 }
 
 function showWelcomeMessage(): void {
   const displayMessage = () => {
-    vscode.window.showInformationMessage(
-      i18nManager.getMessage(LangResourceKeys.welcome),
+    vscode.window.showInformationMessage(i18nManager.getMessage(LangResourceKeys.welcome),
```
JimiC: I wonder why `tslint` is not complaining that the parameters are not inline.

JimiC:
Still waiting for the rest of the translations.

sabinrimni:
Italian Translation: `Grazie per aver installato ${constants.extensionName}`

robertohuertasm:
Only French and Russian missing. Let's hope we can have them soon so we can publish this and make users happier.

robertohuertasm:
By the way, many thanks to our kind translators!! We really appreciate your collaboration!!

MonsieurMan:
Merci d'avoir installé ${constants.extensionName}

ksvitkovsky:
Sorry for late reply 😊 

```
welcome: `Спасибо, что установили ${constants.extensionName}`
```

robertohuertasm:
@MonsieurMan, @ksvitkovsky thanks for your unvaluable help!

PR #1192, opened 2017-09-25
version bump

robertohuertasm:
How do you feel about the installations number? I would like to know your thoughts before merging this PR. :wink:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-icons",
-    "version": "7.12.0",
+    "version": "7.14.0",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
     "name": "vscode-icons",
     "displayName": "vscode-icons",
     "description": "Icons for Visual Studio Code",
-    "version": "7.13.0",
+    "version": "7.14.0",
     "publisher": "robertohuertasm",
     "license": "MIT",
     "author": {


JimiC:
Putting it politely. :-1:
But if you feel like adding them I would add the installs per version.

robertohuertasm:
The thing is that I don't know the installs per version. We need the difference to know it. That's why I thought that this could be useful but I also have my doubts about it.

robertohuertasm:
I'll write this down somewhere and leave this for future consideration.

JimiC:
The thing is that you are the only one that knows those numbers and if anyone else wants to compile a new changelog entry will not know that info.

robertohuertasm:
It's public information. At least I thought it was. I get the information from [here](https://marketplace.visualstudio.com/items?itemName=robertohuertasm.vscode-icons)

JimiC:
Oh. Then the number will not be accurate. How do you know the exact number of the version until you actually push the new one? I mean from the time you add the info, commit, publish some more installations have been done.

robertohuertasm:
It's not accurate, that's for sure. It's an orientative figure.  I get the number just before using the `vsce` tool. Normally, after a week of a release, the number of installs slows down so it's not a huge error margin. This would be for statistical purposes only so we could know the growing trends of the extension and evaluate if people is adopting or running away whenever we change something.

robertohuertasm:
I'll keep this info noted outside the repo for our information for the moment.

jens1o:
We may round the numbers?

JimiC:
Insights Analytics.

JimiC:
I used to use GA for monitoring in another project.

robertohuertasm:
Never mind guys. I didn't want to be invasive by installing analytics. Let's leave it like it is.

JimiC:
Yeah, that's a tricky thing.

PR #1193, opened 2017-09-25
Change locale structure to support translation service.

JimiC:
This PR refactors the language resources files to support upcoming translation service integration.


JimiC:
I have tested it by uninstalling the extension, creating a package and installing again from VSIX.

PR #1231, opened 2017-10-09
Small update for VSCode Icons

torn4dom4n:
- Updated urlOfficialApi link.


src/constants/index.ts
```diff
@@ -11,5 +11,5 @@ export const constants = {
   },
   urlReleaseNote: 'https://github.com/vscode-icons/vscode-icons/blob/master/CHANGELOG.md',
   urlReadme: 'https://github.com/vscode-icons/vscode-icons/blob/master/README.md',
-  urlOfficialApi: 'https://code.visualstudio.com/docs/customization/themes#_select-an-icon-theme',
+  urlOfficialApi: 'https://code.visualstudio.com/docs/getstarted/themes#_selecting-the-file-icon-theme',
```
robertohuertasm: Thanks for this

README.md
```diff
@@ -1,4 +1,4 @@
-# vscode-icons
+# VSCode Icons
```
robertohuertasm: The name was correct, so the change has to be undone.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-icons",
-    "displayName": "vscode-icons",
+    "displayName": "VSCode Icons",
```
robertohuertasm: Same here

torn4dom4n: I think to use **VSCode Icons** is better. **vscode-icons** is only an id name. I can see you use **VSCode Icons** to `Preferences: File Icon Theme`:
image

torn4dom4n: BTW, other File Icon Themes often don't use `name`  for `displayName`.

image

JimiC: @torn4dom4n We have already discussed this as a team and we decided to go with @robertohuertasm reasoning of keeping the name "dev" related. For the history, I have already raised the naming issue some months ago.

robertohuertasm: Yes, this was already discussed. We would have to dig the closed issues to get that old debate. 

@JimiC was in favour of changing it and I have to say that I've come to abandon the initial idea that I held to when we discussed this. So... if you want to open the debate again, I'm already open to discuss it. Even I've come to think that now it's really the time to get rid of the `preview` tag. 

@JimiC , @jens1o what is your opinion on this?

JimiC: I'm in awe 😮

robertohuertasm: That's a good or bad awe?

JimiC: Neutral.

torn4dom4n: For me, I want to use **VSCode Icons** because **vscode-icons** is an id name. And I favor removing `preview` tag, **VSCode Icons** is good enough to use.

JimiC: If used, `VSCode Icons` will still be just the display name. `vscode-icons` **is** the extensions name and will remain so even if the displayed name changes.

torn4dom4n: VSCode has `displayName` and Atom is not. See https://code.visualstudio.com/docs/extensionAPI/extension-manifest#_fields

JimiC: @torn4dom4n Exactly what I'm saying.

JimiC: To not make this PR comment lengthy I suggest to keep only the link update and move the name change to an issue for further discussion.

torn4dom4n: @JimiC Should I revert commit `Using VSCode icons name` and open an issue?

JimiC: You will have to keep only `Update link`. And yes opening a new issue is the way to go.

torn4dom4n: I opened an issue to discuss at https://github.com/vscode-icons/vscode-icons/issues/1232

PR #1235, opened 2017-10-12
Docker Images

robertohuertasm:
Just improved the docker images:

- reduced file size by using `node:alpine`.
- fixed issue with `vscode.d.ts` error when using some linux docker images.
- added a docker image in order to get the latest artifact from the published version, not the local. This may be useful for cases like [vscode-icons for Github](https://github.com/dderevjanik/github-vscode-icons) extension where the author requires the latest `icons.json` file.

//cc @dderevjanik

### DIFF
--- a/docker/development/README.md
+++ b/docker/development/README.md
@@ -1,0 +1,34 @@
+# development
+
+If you want to **build** the project you can do this:
+
+```sh
+# First, build the image if not already build.
+# This should be a one time command.
+docker build -t dev docker/development
+
+# Then run it.
+# Note that we're using a volume to have access to the artifact in our host.
+# Here we're using a Docker for Windows mapping. If you're using other OS then take a look at https://docs.docker.com/engine/admin/volumes/volumes/.
+# Once the container is done it will automatically be removed and the artifact will be in your mapped folder.
+docker run --rm -it -v d:/GIT/github/robertohuertasm/vscode-icons:/app dev
+
+# Finally, if you want to clean the stale volumes you can run this.
+docker volume rm $(docker volume ls -qf dangling=true)
+```
+
+If you want to use `Docker` in order to **develop** you can do this:
+
+```sh
+# First, build the image if not already build.
+# This should be a one time command.
+docker build -t dev docker/development
+
+# Then run it.
+# Note that we're using a volume to have access to the artifact in our host.
+# Here we're using a Docker for Windows mapping. If you're using other OS then take a look at https://docs.docker.com/engine/admin/volumes/volumes/.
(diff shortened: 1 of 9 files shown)


JimiC:
I don't know much about `docker` (I know that this might come as a surprise to you) so I guess my approval is just typical.

dderevjanik:
Hi,
thank you very much for providing `remote-builder`. This will be super-helpful for `vscode-icons for github`, no more manual downloading and building `vscode-icons` :+1:

docker/development/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+```sh
+# First, build the image if not already build.
+# This should be a one time command.
+docker build -t dev docker/development
+
+# Then run it.
+# Note that we're using a volume to have access to the artifact in our host.
+# Here we're using a Docker for Windows mapping. If you're using other OS then take a look at https://docs.docker.com/engine/admin/volumes/volumes/.
+docker run -it -v d:/GIT/github/robertohuertasm/vscode-icons:/app dev sh
```
JimiC: Shouldn't this be `d:/GIT/github/vscode-icons/vscode-icons:/app` ?

docker/development/README.md
```diff
@@ -0,0 +1,34 @@
...
+```sh
+# First, build the image if not already build.
+# This should be a one time command.
+docker build -t dev docker/development
+
+# Then run it.
+# Note that we're using a volume to have access to the artifact in our host.
+# Here we're using a Docker for Windows mapping. If you're using other OS then take a look at https://docs.docker.com/engine/admin/volumes/volumes/.
+# Once the container is done it will automatically be removed and the artifact will be in your mapped folder.
+docker run --rm -it -v d:/GIT/github/robertohuertasm/vscode-icons:/app dev
```
JimiC: Shouldn't this be d:/GIT/github/vscode-icons/vscode-icons:/app ?

docker/remote-builder/README.md
```diff
@@ -0,0 +1,26 @@
...
+```sh
+# First, build the image if not already build.
+# This should be a one time command.
+docker build -t vsi docker/remote-builder
+
+# Then run it.
+# Note that we're using a volume to have access to the artifact in our host.
+# Here we're using a Docker for Windows mapping. If you're using other OS then take a look at https://docs.docker.com/engine/admin/volumes/volumes/.
+# Once the container is done it will automatically be removed and the artifact will be in your mapped folder.
+docker run --rm -it -v c:/Users/Roberto/Desktop/s/shared:/app-out vsi
```
JimiC: How about replacing `Roberto` with a placeholder?

PR #1266, opened 2017-10-15
Add Crowdin config file.

JimiC:
**_Fixes #1194_**

**Changes proposed:**

- [x] Add

**Things I've done:**


JimiC:
This PR just enables the usage of the service. It doesn't change the translations as of yet.

robertohuertasm:
@JimiC, it seems that I have to provide the `source file` and then a path to different supported langs.

image

I guess we should separate `locale` into different folders? Let's say `locale\source` and `locale\translations`? I'll wait for your instructions before proceeding 😉

PR #1283, opened 2017-10-22
Map `language` and `lang`  to `locale` folder icon

jens1o:
**Changes proposed:**

- [X] Add
- [ ] Prepare
- [ ] Delete
- [ ] Fix

https://github.com/vscode-icons/vscode-icons/issues/650#issuecomment-338482447

### DIFF
--- a/src/icon-manifest/supportedFolders.ts
+++ b/src/icon-manifest/supportedFolders.ts
@@ -43,7 +43,7 @@ export const extensions: IFolderCollection = {
     { icon: 'include', extensions: ['include', 'includes', 'incl'], format: FileFormat.svg },
     { icon: 'less', extensions: ['less', '_less'], format: FileFormat.svg },
     { icon: 'library', extensions: ['lib', '.lib'], format: FileFormat.svg },
-    { icon: 'locale', extensions: ['locale', 'locales', 'i18n', 'g11n'], format: FileFormat.svg },
+    { icon: 'locale', extensions: ['lang', 'language', 'locale', 'locales', 'i18n', 'g11n'], format: FileFormat.svg },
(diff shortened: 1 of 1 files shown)


src/icon-manifest/supportedFolders.ts
```diff
@@ -43,7 +43,7 @@ export const extensions: IFolderCollection = {
     { icon: 'include', extensions: ['include', 'includes', 'incl'], format: FileFormat.svg },
     { icon: 'less', extensions: ['less', '_less'], format: FileFormat.svg },
     { icon: 'library', extensions: ['lib', '.lib'], format: FileFormat.svg },
-    { icon: 'locale', extensions: ['locale', 'locales', 'i18n', 'g11n'], format: FileFormat.svg },
+    { icon: 'locale', extensions: ['lang', 'language', 'locale', 'locales', 'i18n', 'g11n'], format: FileFormat.svg },
```
robertohuertasm: @jens1o, what about `languages`?

jens1o: I didn't saw that ever in the wild. Do you have an example for it?

robertohuertasm: Not really, but I see it more likely that someone names their language folder with `languages` instead of the singular name. Usually, one tends to have more than one language and I'm just following the same logic we applied with `locale` and `locales`. If I were to name a folder with languages I would definetely use the plural.

JimiC: Point

jens1o:
@vscode-icons/development What do we do with `bitHound - Dependencies — 3 failing dependencies. `? I think that the problem should be solved in another pr?

PR #1312, opened 2017-10-26
Fixed pad disable detection flow

robertohuertasm:
**_Fixes #1311_**

**Changes proposed:**

- [ ] Add
- [ ] Prepare
- [ ] Delete
- [x] Fix

**Things I've done:**


JimiC:
This happened because in the latest iteration I remove `results.length` check in `detectAngular` and then I forgot to change the return value of `detectProject`. This is the price we pay for not completely covering the project with BDD test cases. I'm to blame for both.

test/init/projectAutoDetection.test.ts
```diff
@@ -67,7 +67,9 @@ describe('AutoDetectProject: tests', function () {
         function () {
           userConfig.projectDetection.disableDetect = true;
           return pad.detectProject(null, userConfig)
-            .then(res => expect(res).to.be.an('array'));
+            .then(res => {
```
JimiC: @robertohuertasm Is this your prefered coding style?

robertohuertasm: No. I was using this to debug and later I forgot to go back to a one line style. It was late... 😉

PR #1329, opened 2017-11-04
Update dependencies

JimiC:
This PR also switched to `maintainers` instead of `contributors` in `package.json`.

@robertohuertasm  This should be merged just before the next release.


JimiC:
1. Cause he's the author. He wasn't in the `contributors` either.
2. Sure.

PR #1330, opened 2017-11-04
Replace ${workspaceRoot} with ${workspaceFolder}

torn4dom4n:
VSCode is deprecating `${workspaceRoot}` and have introduced a new substitution variable `${workspaceFolder}`. This variable will get substituted to the absolute path of the workspace folder containing the appropriate `launch.json` or `tasks.json`.


JimiC:
@torn4dom4n Please sync your `master` branch before submitting a PR.

JimiC:
Please revert.

JimiC:
@robertohuertasm Did you test it?

torn4dom4n:
@JimiC You can read about it at [here](https://code.visualstudio.com/updates/v1_17#_workspacefolder-in-launchjson-and-tasksjson).

JimiC:
@torn4dom4n I know about the deprication but changes as this should be tested never the less.

JimiC:
And after all, there is no hurry. `workspaceRoot` will be kept for backwards compatibility.

torn4dom4n:
I see VSCode 1.18 (Next version) also announced [it](https://github.com/Microsoft/vscode-docs/blob/vnext/release-notes/v1_18.md#variable-names).

robertohuertasm:
I've tried it and it seems to be working as expected. No issues with this change as far as I can see.

JimiC:
OK, then. There is also a change needed in `tasks.json` while we are on replacing deprecated keys.

PR #1342, opened 2017-11-10
Add ability to hide folder arrows in explorer.

JimiC:
**_Fixes #1337_**

**Changes proposed:**

- [x] Add

**Things I've done:**


JimiC:
Upon merging the PR `Crowdin` should inform our translators that new translations are needed.

src/icon-manifest/manifestReader.ts
```diff
@@ -0,0 +1,30 @@
...
+      .filter(key => key.toString().startsWith(`_${isFile ? 'f' : 'fd'}_${name}`)).length;
+  }
+
+  public static folderIconsDisabled(func: (iconsJson: models.IIconSchema) => boolean): boolean {
+    const iconManifest = this._getIconManifest();
+    const iconsJson = iconManifest && parseJSON(iconManifest) as models.IIconSchema;
+    return !iconsJson || !func(iconsJson);
+  }
+
+  private static _getIconManifest(): string {
```
robertohuertasm: I know this is an existing function but I didn't see the underscore until now. Why the use of this syntax if we already have the private modifier?

JimiC: This is what happens when you write code in many projects with different coding styles. I'll fix it as soon as I got the time or you can do that too if everything else is fine.

JimiC: Fixed.
@robertohuertasm In order to avoid same cases in the future how about using naming convention `tslint` rule like they do in [tslint-microsoft-contrib](https://github.com/Microsoft/tslint-microsoft-contrib)?

robertohuertasm: We're already using `tslint:latest`. I guess this library you propose is composed by custom rules. I've never tried it so I don't have any opinion.

JimiC: You are right. Those are custom rules. Tslint latest doesn't have naming convention rules. On the other hand, maybe this is too much of a hassle for such a small group of developers.

jens1o:
@robertohuertasm When's the release? It's scheduled to today, I'm counting! :D

PR #1359, opened 2017-11-16
Update Yarn version in CI.

JimiC:
`Travis CI` pre-build containers use `Yarn v0.27.5` which is outdated by far.
This PR forces `Travis CI` to use version `1.3.2` which by the time this is written is the latest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ node_js:
 matrix:
   allow_failures:
     - node_js: node
+before_install:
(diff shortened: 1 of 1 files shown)


jens1o:
Can we still hit up travis, so they can update itself?

JimiC:
@jens1o Travis uses containers with preinstalled software. They will have to make a new container image just to upgrade `Yarn`. Probably they use a stable version of Yarn that they are happy with. After all the ability to update it is there.

PR #1371, opened 2017-11-22
Support for cargo

robertohuertasm:
**_Fixes #1367_**

**Changes proposed:**

- [x] Add
- [ ] Prepare
- [ ] Delete
- [ ] Fix

**Things I've done:**


JimiC:
I don't know where you got it but it's good. Saved me the trouble of creating one.

robertohuertasm:
I've created it from scratch using Illustrator. I'm not so `Rusty` as I thought! I'm glad you like it!

JimiC:
I'm impressed. 😮

robertohuertasm:
The pity is that when viewed in `vscode` it loses a lot of detail and it's not really as good as it may seem at first sight... but at least is recognizable. 😉

JimiC:
I knew this would be the case from the start and thought of proposing to use the `crate` icon but it doesn't represent `Cargo`, so I hashed.

robertohuertasm:
Yep, that's the problem with complex icons. I think that the result is not optimum in this case also but I guess that is fairly acceptable given the circumstances.

JimiC:
You know that those icons shine on a 4K 32" monitor, right?

robertohuertasm:
😂 I don't happen to have a 4K 32" monitor... for the moment! 😂 I guess you are one of the lucky ones?

JimiC:
You kidding me?!? It's not `Black Friday` yet.

PR #1372, opened 2017-11-23
Update submodule wpilgenerator.

JimiC:
This PR fixes the following issues with `wpilgenerator`:
* Fixes incorrect path to the code repo directory.
* Fixes logging inconsistencies when logging on a non-TTY terminal (like in `Travis-CI`).


JimiC:
It looks like it's now working as intended. At least when there are no changes to the icons definitions the tool detects that correctly. Next test case would be to detect with changes.

PR #1392, opened 2017-12-03
Update dependencies.

JimiC:


JimiC:
@robertohuertasm The build is failing for node `6.5.0` because `streamfilter v1.0.6` requires node engine to be `>=6.9.5`.
We have two options:
1. Revert the dependency update that uses that package
2. We drop support for `vscode  < 1.12.0`.

JimiC:
Closing in favor of  #1394. There is something obviously wrong here but I can't figure out what.

PR #1393, opened 2017-12-03
Add script to initialize submodules.

JimiC:
This PR adds a `preinstall` script to initialize the `submodules` for those that don't use a git client that already does that.

This script does not impact our `CI` service.


JimiC:
@robertohuertasm Please restart `appveyor` job.

PR #1406, opened 2017-12-22
Fix project autodetection on standalone files

zackschuster:
**Changes proposed:**

- [ ] Add
- [ ] Prepare
- [ ] Delete
- [x] Fix

**Things I've done:**

- [ ] My pull request fixes an issue, I referenced the issue.

currently throws here with "path must be a string or Buffer" error.

this PR introduces a simple `Array#filter` call to guard against that.

(as an aside, this could be caught with `--strictNullChecks` enabled :) )

### DIFF
--- a/src/init/projectAutoDetection.ts
+++ b/src/init/projectAutoDetection.ts
@@ -54,12 +54,14 @@ export class ProjectAutoDetection {
 
   public static getProjectInfo(results: models.IVSCodeUri[], name: models.Projects): models.IProjectInfo {
     let projectInfo: models.IProjectInfo = null;
-    results.some(result => {
-      const content = fs.readFileSync(result.fsPath, 'utf8');
-      const projectJson = parseJSON(content);
-      projectInfo = this.getInfo(projectJson, name);
-      return !!projectInfo;
-    });
+    results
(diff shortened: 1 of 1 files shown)


zackschuster:
note: i'll add a test later to make codecov happy. popped this up because i noticed the error when i was checking on something else from Running Extensions (which i guess displays uncaught errors now -- handy!)

jens1o:
lgtm, waiting for @JimiC 's response

JimiC:
@zackschuster Thanks for the PR but I have my doubts about it. `IVSCodeUri` is a reflected interface of `VSCodeUri` which is already strongly typed, meaning that `fspath` is always of type `string`. So your guard clause doesn't make sense. On the other hand, a `null` check guard clause makes more sense, considering also your mention about `--strictNullChecks` compiler option.

zackschuster:
ah, my apologies -- the stack trace pointed to `readFileSync` throwing, so i assumed that was the culprit. it's not throwing on my machine anymore either, so problem solved? enough to close the PR, anyway.

that being said, i think `--strictNullChecks` would be great to add to the codebase. would you mind a PR at a later time attempting that?

thanks for your time :)

Issue #1416, opened 2018-01-08
question about variable substitution in `vsicons.customIconFolderPath`


gerardmrk:
I have a custom icon folder in my project directory that I would like to add into version control, but it seems I'm unable to utilize variable substitution in the `vsicons.customIconFolderPath` setting.

I've tried **${workspaceRoot}**, **${workspaceFolder}** and even a relative path (relative from `.vscode/settings.json`), and none of them works. But specifying the full absolute path to the directory works (using `pwd`). I would like to know if variable substitution is possible, thanks

JimiC:
Have you read https://github.com/vscode-icons/vscode-icons/wiki/Custom?

gerardmrk:
@JimiC yes I have, the doc implies absolute path has to be specified. The configuration is not shareable across multiple machines that way, e.g. `/Users/gerard/Documents/DevProjects/src/github.com/gerardmrk/xo` isn't the same for other users' OS if they're using the same repo.

JimiC:
Have you tried using a path relative to the project's root directory?

JimiC:
@gerardmrk So it turns out that this feature indeed works only with an absolute path to the `vscode-custom-icons` folder. ANd in your case, yes it will not work.
I'll see what I can do to add something that uses relative paths.

JimiC:
@gerardmrk I'm gonna close this issue as it's out of my current abilities to make this work with relative paths. You see the extension needs to know the absolute path where the custom icons folder is in order to use the custom icons. Searching from where the extension is to where the custom icons folder is, in order to figure out the relative path, means that we must use an algorithm that scans your entire file system to locate it, something that isn't efficient at all.

If anyone has another idea, pleae share.

JimiC:
I'm experimenting to see if I can come up with something, using `workspaceRoot` and `workspaceFolder`.
Will post back if I have something worth mentioning.

radziksh:
@gerardmrk I added in `projectFolder/.vscode/settings.json` these lines:
```
"vsicons.customIconFolderPath": "./.vscode",
 "vsicons.associations.files": [
        { "icon": "formRelated", "extensions": ["form-container.tsx"], "filename": true, "format": "png", "overrides": "form-container.tsx" },
        { "icon": "formRelated", "extensions": ["form-view.tsx"], "filename": true, "format": "png", "overrides": "form-view.tsx" }
    ],
```

[...]

PR #1422, opened 2018-01-14
Add support to specify a relative path for custom icons folder path.

JimiC:
_**References #1416**_

**Changes proposed:**

- [x] Add

This PR adds the functionality to set the `vsicons.customIconFolderPath` with a relative path to the root of the opening project. This is handy when distributing custom icons with the project source code.

I have also done some code refactoring which normally should have been submitted in another PR but I'm getting lazy lately.

Also, once this PR gets merged the [wiki page](https://github.com/vscode-icons/vscode-icons/wiki/Custom) needs to be updated to include the added ability.

### DIFF
--- a/src/icon-manifest/iconGenerator.ts
+++ b/src/icon-manifest/iconGenerator.ts
@@ -80,6 +80,18 @@ export class IconGenerator implements models.IIconGenerator {
     }
   }
 
+  public getCustomIconFolderPath(folderPath: string): string {
+    if (!folderPath) { return this.settings.vscodeAppData; }
+
+    const fPath = folderPath.trim();
+    if (path.isAbsolute(fPath) || !this.settings.workspacePath || !this.settings.workspacePath.length) {
+      return fPath;
+    }
+
+    const absPath = this.settings.workspacePath.find(wsp => fs.existsSync(wsp));
+    return utils.pathUnixJoin(absPath, fPath);
+  }
+
   public hasCustomIcon(customIconFolderPath: string, filename: string): boolean {
     const relativePath = utils.getRelativePath('.', customIconFolderPath, false);
     const filePath = utils.pathUnixJoin(relativePath, filename);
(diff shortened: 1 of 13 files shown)


src/icon-manifest/iconGenerator.ts
```diff
@@ -80,6 +80,18 @@ export class IconGenerator implements models.IIconGenerator {
     }
   }
 
+  public getCustomIconFolderPath(folderPath: string): string {
+    if (!!!folderPath) { return this.settings.vscodeAppData; }
```
robertohuertasm: One `!` should be enough.

src/icon-manifest/iconGenerator.ts
```diff
@@ -80,6 +80,18 @@ export class IconGenerator implements models.IIconGenerator {
...
+
+    const fPath = folderPath.trim();
+    if (path.isAbsolute(fPath) || !!!this.settings.workspacePath || !this.settings.workspacePath.length) {
```
robertohuertasm: the same here

src/icon-manifest/iconGenerator.ts
JimiC: I'm converting them to `boolean`.

robertohuertasm: I know, that's what I say you only need one

JimiC: Actually one is a null check. Three is a boolean negative.

robertohuertasm: There's no such thing as a null check in `JavaScript`. `!` is a logical operator and also implies a casting to `bool` due to the concepts of falsy and truthy. `!!!` is not idiomatic JavaScript.

robertohuertasm: https://stackoverflow.com/posts/25318045/revisions

JimiC: It's just my preferred way of coding, expressing the fact that I'm doing `!!(!a)`.

robertohuertasm: It's ok but it's not really necessary and thus it's not used in JavaScript. I would prefer to get rid of this if possible as I think we should stick to common practices. 😉

PR #1430, opened 2018-01-27
Use bundler icon for files with .gemfile extension

artplan1:
**Changes proposed:**

- [x] Add

---

Useful for files with names like `spree_master.gemfile` / `spree_master.gemfile.lock`


src/icon-manifest/supportedExtensions.ts
```diff
@@ -91,6 +91,7 @@ export const extensions: IFileCollection = {
     { icon: 'bolt', extensions: [], languages: [languages.bolt], filename: true, format: FileFormat.svg },
     { icon: 'bower', extensions: ['.bowerrc', 'bower.json'], filename: true, format: FileFormat.svg },
     { icon: 'buckbuild', extensions: ['.buckconfig'], filename: true, format: FileFormat.svg },
+    { icon: 'bundler', extensions: ['gemfile', 'gemfile.lock'], format: FileFormat.svg },
```
jens1o: it is already defined a line below?

artplan1: it's defined below with `filename: true` and works only for files `gemfile` and `gemfile.lock`

JimiC:
My only question is how common is the use of such files. Any docs?

artplan1:
@JimiC 
for example gem `appraisal` generates files with this extensions:
https://github.com/thoughtbot/appraisal#under-the-hood

PR #1446, opened 2018-02-09
Updated Icon yaml to support .yaml extension

dstrebel:
_**Fixes #IssueNumber**_

**Changes proposed:**

- [X] Add
Added support for .yaml extension, as currently yaml icon only support .yml
image
- [ ] Delete
- [ ] Fix
- [ ] Prepare

### DIFF
--- a/src/icon-manifest/supportedExtensions.ts
+++ b/src/icon-manifest/supportedExtensions.ts
@@ -608,7 +608,7 @@ export const extensions: IFileCollection = {
     { icon: 'xliff', extensions: ['xliff', 'xlf'], format: FileFormat.svg },
     { icon: 'xml', extensions: ['pex', 'tmlanguage'], languages: [languages.xml], format: FileFormat.svg },
     { icon: 'xsl', extensions: [], languages: [languages.xsl], format: FileFormat.svg },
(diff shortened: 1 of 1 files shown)


robertohuertasm:
`.yaml` extension is already supported via language id. Indeed I cannot reproduce your issue even with all my extensions disabled. That leads me to think that you may have  some kind of custom mapping for your .yaml extensions. Can you create a .yaml file and check what language is enabled? 

image

JimiC:
Check your settings.

JimiC:
ANd according to [this](https://github.com/Microsoft/vscode/blob/master/extensions/yaml/package.json#L11) we can lso drop the  manual declaration of `yml` in our extension. Can't remember why we left it in there...

robertohuertasm:
I couldn't also figure out why we left it... I guess maybe there was a time when it was not supported by the language.

JimiC:
Yeah, that's the case. Looking at the [source code](https://github.com/Microsoft/vscode/blob/8f35cc4768393b25468416829e980d7550619fb1/extensions/yaml/package.json#L10) it has been always there.

JimiC:
@dstrebel Can you please provide feedback?

JimiC:
Closing this.
@dstrebel If you require additional support please open an issue.

PR #1467, opened 2018-02-17
Update dependencies.

JimiC:
**Note:** I switched `@types/node` back to `7.0.54` as we should keep in line with the `node` version the editor is been built against, which is `7.9.0`.


JimiC:
@robertohuertasm This PR SHOULD be merged before the next release.

PR #1468, opened 2018-02-17
Fix package managers lock files.

JimiC:


JimiC:
I don't know why but the lock files didn't update correctly.

PR #1487, opened 2018-02-26
Fix CI stage for list of icons wiki pages

### LINKED ISSUE #1486: [Task] Fix CI stage for list of icons wiki pages
The CI stage currently fails due to the fact that the extension doesn't get built, thus it can't find the necessary `supported*.js` files to import.

JimiC:
_**Fixes #1486**_

**Changes proposed:**

- [x] Fix


JimiC:
@robertohuertasm This PR has high priority.

PR #1523, opened 2018-03-18
Support for windows and x64 arch

robertohuertasm:
The `wallaby.js` file specifies the path to the node version assuming the user is using `nvm` and a non Windows machine. This PR avoids fixing the node version and relies on user's executing node version, provided that users and mantainers should know that our supported node version is 7.9.0.


JimiC:
I have a solution. 
```
env: {
    type: "node",
    runner: process.platform === 'win32'
      ? `${process.env.APPDATA}\\nvm\\v7.9.0\\node`
      : `${require('os').homedir()}/.nvm/versions/node/v7.9.0/bin/node`
  },
```

robertohuertasm:
@JimiC, I've used your code and added support for x64 architectures.

JimiC:
We also need to change the title of the PR.

robertohuertasm:
Done!

JimiC:
@robertohuertasm Unfortunately, the `process.arch` switch isn't working when you have already used that `node` version (`nvm use <version>`). `node64.exe` gets renamed to `node.exe`.
Consider using the above-proposed code snippet.

Issue #1524, opened 2018-03-18
Get rid of preview tag


robertohuertasm:
- [x] I'm sure this issue is not a _duplicate_.

I'd suggest getting out of the `preview` phase. The extension is in a very stable phase so I don't see any reason to keep this tag

jens1o:
Originally, I wanted to do that :D

PR #1525, opened 2018-03-18
no more preview tag

robertohuertasm:
Fixes _**#1524**_

No more preview tag


jens1o:
Oh, forgot to commit that :/

PR #1537, opened 2018-03-22
Add uninstall hook script (cleanUp)

### LINKED ISSUE #1499: [Task] Perform cleanup when extension gets uninstalled

We should explore the ability to add a cleanup process when the extension gets uninstalled.
Mainly what needs to be cleaned is the `vsicons.settings.json`.

Info: https://code.visualstudio.com/updates/v1_21#_extension-uninstall-hook

JimiC:
_**Fixes #1499**_

**Changes proposed:**

- [x] Add

### DIFF
--- a/package.json
+++ b/package.json
@@ -235,6 +235,7 @@
         "posttest": "nyc report -r lcov",
         "test:vs": "node ./node_modules/vscode/bin/test",
         "vscode:prepublish": "npm run build",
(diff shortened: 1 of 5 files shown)


JimiC:
On an upcoming time, we should also add a handler to handle failed updates and clean up the mess. This will reduce our support load for such cases.

Issue #1271, opened 2017-10-15
[Icon Request] Support for LaTeX Beamer, LaTeX Log, LaTeX Memoir


torn4dom4n:
- [x] I'm sure this issue is not a duplicate.

- [x] I want to create an icon request:
 - Type: `extension`
 - Icon Name: `tex`
 - Sample Icon: https://github.com/vscode-icons/vscode-icons/blob/master/icons/file_type_light_tex.svg
 - Language ids: `latex beamer`, `latex log`, `latex memoir`

Language Extensions: https://marketplace.visualstudio.com/items?itemName=torn4dom4n.latex-support and https://marketplace.visualstudio.com/items?itemName=James-Yu.latex-workshop

JimiC:
@torn4dom4n Is it OK with to associate those with the `tex` icon?

JimiC:
Looking at the `package.json` of both extensions I notice that there are no extensions defined to those ids. So the issue is kinda invalid.

torn4dom4n:
I am not sure about extensions for this. Can you enable icons without extensions defined?

JimiC:
No. For icons to light up they need to be associated with a file, either whole filename or extension.

torn4dom4n:
@JimiC Thank for that. I will close this issue!

PR #1546, opened 2018-03-25
Add support for LaTeX Expl3 and DocTeX

Stone-Zeng:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Recently, [LaTeX-Workshop](https://github.com/James-Yu/LaTeX-Workshop) has added support for DocTeX and LaTeX Expl3. So it's time to add the icon support too.

I use the same icon for them. Maybe we need something else, e.g. the hummingbird in <https://www.latex-project.org/latex3> for LaTeX Expl3? (personally, I think the original tex icon is enough).

By the way, #1271 requires support for LaTeX Beamer, LaTeX Log, LaTeX Memoir. It seems that LaTeX Log has been removed, but should I support Beamer and Memoir too? I think there are not so many people use them.


Stone-Zeng:
So how to add icon for the expl3 files? At present they are labelled with the default icon.

JimiC:
No file extension, no icon, no support, simple.

JimiC:
@Stone-Zeng Can you please follow up with the requested changes?

Stone-Zeng:
By the way, I'm still confusing about the relation of languages, extensions and icons. For example, language `json` and `jsonc` share the same extension `.json`. For latex and latex-expl3, the situation is the similar. Could you give some explanations in detail? Thanks very much.

JimiC:
@Stone-Zeng The `defaultExtension` attribute in https://github.com/vscode-icons/vscode-icons/blob/master/src/icon-manifest/languages.ts#L88 does not correctly reflect the file extension(s) of the `language id` used. That attribute is just a helper for our `exampleGenerator` tool.

The correct file extensions and filenames are those declared here: 
https://github.com/Microsoft/vscode/blob/master/extensions/json/package.json#L43

P.S.: While mentioning it, the `defaultExtension` for `jsonc` should have been `.jsonc` but it slipt me by when reviewing @robertohuertasm  PR at that time.

Stone-Zeng:
Can this PR be merged?

jens1o:
cc @robertohuertasm

PR #1548, opened 2018-03-28
Lint wallaby config file

JimiC:


jens1o:
Is it really worth it?

JimiC:
From my POV, no, but I had to ask.

PR #1550, opened 2018-04-04
Update yarn in CI

JimiC:


JimiC:
I don't know but because `yarn` moved 2 versions ahead it's time to update it.

PR #1582, opened 2018-05-17
support for @types directory

dojineko:
image

add `@types` support as type-definitions directory

ref: https://github.com/vscode-icons/vscode-icons/wiki/ListOfFolders

**Changes proposed:**

- [x] Add


JimiC:
@vscode-icons/development What do you think? Is `types` a unique folder name specific to `TypeScript`?

robertohuertasm:
I'm not really sure about this. `types` are not `typings`, so I don't see this folder as a proper candidate. In any case, I will say that this kind of namings should be managed via custom icons. @dojineko can you provide some reasoning behind this name folder? Otherwise, I'd vote for not approving this PR.

JimiC:
@robertohuertasm The thing is that `types` is a valid `TypeScript` compiler [option](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types) for defining the types to include. That my reasoning at least.

robertohuertasm:
Indeed, you have that compiler option but the reference to the `typeRoots` options says `typings` not `types`:

```
{
   "compilerOptions": {
       "typeRoots" : ["./typings"]
   }
}
```
Anyway, this is a minor issue.

I think that we should change from `types` to `@types`, given that this is really the name of the folder that it's being used by default in order to seek for typings. Furthermore, by making this change we reduce the risk of overlapping with any other users using `types`, which is really a name that could be used outside the TypeScript domain and we also comply with the TypeScript defaults.

dojineko:
@robertohuertasm 

Your suggestion is indeed right.
However, routinely place type definition files for the `types` directory. (It is not managed by typings)
And typings are already deprecated today and are now using npm's `@types` or `types` packages.

- https://github.com/typings/typings
- https://www.npmjs.com/~types
- https://github.com/types

So I think it is desirable to rename typings to typescriptdef just like typescript(as extension) and typescriptdef.
And after that, I think that it is better to include `typings` `@types` and `types`.

robertohuertasm:
@dojineko I agree in all your other suggestions but still I feel like maybe going for `types` is a little bit too much. Yes, the `Vue` repo is using it but I don't think is like something that is massively used and I do think that is really a common name that could be used for any other purposes not really related to TypeScript, so that's why I'm being defensive here. I think that even they decided to go for a `@types ` folder (I guess to emphasise the scope effect) instead of `types`.  

Maybe we can go with what we agree on and keep an eye on the `types` request. See if it's widely adopted or not. My reasoning is that we already implemented customizations specifically to deal with these blurry cases where it can make sense to implement it but still feels like we're not really into the 
mainstream adoption phase.

dojineko:
@robertohuertasm

> See if it's widely adopted or not.

I agree. And in my research, at least the `types` directory is used for storing type definitions in the following projects.

- DefinitelyTyped
  - https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types
- Vue official projects
  - https://github.com/vuejs/vue
  - https://github.com/vuejs/vuex
  - https://github.com/vuejs/vue-router
  - https://github.com/vuejs/vue-meta
- ava
  - https://www.npmjs.com/package/ava
  - https://registry.npmjs.org/ava/-/ava-0.25.0.tgz
- chalk
  - https://github.com/chalk/chalk/tree/master/types
- sequelize-typescript
  - https://github.com/RobinBuschmann/sequelize-typescript/tree/master/lib/types
  - https://registry.npmjs.org/sequelize-typescript/-/sequelize-typescript-0.6.4.tgz
- fast-glob
  - https://github.com/mrmlnc/fast-glob/tree/master/src/types
  - https://registry.npmjs.org/fast-glob/-/fast-glob-2.2.1.tgz

For that reason I do not think there is anything unnatural to add types as part of the behavior of typings.
This is similar to storing the FlowType type definition in the flow directory, and an icon is already set in the flow directory.
(However, in some packages like mime, I saw that `types` is not used as a directory to store type definitions.)

The idea that the directory name `@types` is derived from Scoped Packages of npm and provides a namespace for the package derived from DefinitelyTyped.
And according to your suggestion, I understand that it is ok to add `@types` to the list so I will add a commit.

robertohuertasm:
> For that reason I do not think there is anything unnatural to add types as part of the behavior of typings.

@dojineko, I don't see anything unnatural neither. My reasoning is not about that. It's about specificity. We have to only use an icon if we're sure that the community uses a name for a specific thing. In case of use collision, we always favor the most used one.

To summarize. The accepted ones:

- typings
- @types

In doubt:
- typescriptdef: where have you seen this?
- types: I'm a little bit reluctant but if @JimiC and @jens1o are ok with it I will vote in favour, too.

dojineko:
@robertohuertasm 

i added `@types`. 
(it seems that past @JimiC and @jens1o approves has been automatically dismissed as changes have been added. 😇)

> typescriptdef: where have you seen this?

this is here
https://github.com/vscode-icons/vscode-icons/blob/master/src/icon-manifest/supportedExtensions.ts#L547-L548

so, my proposal is how about this. (however, it has the problem of being incompatible.)

```diff
-    { icon: 'typings', extensions: ['typings'], format: FileFormat.svg },
+    { icon: 'typescriptdef', extensions: ['typings', '@types', 'types'], format: FileFormat.svg },
```

robertohuertasm:
@dojineko, just change the name of the icon to typescriptdef instead of typings and everything will be fine.

Regarding @Jimic and @jens1o 's approval I would like to have their feedback to see if they may have changed the way they see this.

dojineko:
@robertohuertasm
> just change the name of the icon to typescriptdef instead of typings and everything will be fine.

i changed `typings` to `typescriptdef`! is this ok?

JimiC:
Sorry for the late reply but I was out of town.
@robertohuertasm 
- By renaming the `icon` name we are breaking the API.
- Icons should also be reverted to their original state.
- The only addition should be the `@types` in `extensions`, as we should go with what is ensured to not cause any inconvenience.

dojineko:
@robertohuertasm revert it 😇

JimiC:
@dojineko Not quite. You still have to revert the changed icon.

dojineko:
@JimiC oops, i reset and force pushed it. 😭

jens1o:
@robertohuertasm I'm sorry, but I already stated
> I'm fine with this, as I have never seen this folder with another use.

JimiC:
@dojineko Let's just drop `types` from the extensions for the time being.

dojineko:
@JimiC sorry for late, i dropped `types`, and i will use fine tune for it. 😇

elongl:
@JimiC
I think `types` is a valid and expected name for a folder in which you define *types*.
I strongly recommend adding it to the extension.

PR #1584, opened 2018-05-20
Associate "firestore.rules" filename with Firebase Firestore icon

donmckenna:
**Changes proposed:**

- [x] Add


JimiC:
Can you please post a link so we can get more info on the filename? I suspect that this may need its own icon.

donmckenna:
Cloud Firestore (Firestore) is a NoSQL database under the Firebase family of products.
https://firebase.google.com/products/firestore/
`firestore.rules` are the security rules for Firestore database access.

I agree with you and have made a Firestore-specific icon based on the existing `firebase.svg` icon.
It retains the Firebase root but includes the child Firestore logo in the corner like the TS or JS on Angular icons.

Hope this is more useful.

src/icon-manifest/supportedExtensions.ts
```diff
@@ -226,6 +226,7 @@ export const extensions: IFileCollection = {
     { icon: 'favicon', extensions: ['favicon.ico'], filename: true, format: FileFormat.svg },
     { icon: 'fbx', extensions: ['fbx'], format: FileFormat.svg },
     { icon: 'firebase', extensions: ['.firebaserc'], filename: true, format: FileFormat.svg },
+    { icon: 'firebase_firestore', extensions: ['firestore.rules'], filename: true, format: FileFormat.svg },
```
JimiC: Better use simply `firestore` as the `icon` attribute value.

donmckenna: Changed :)

donmckenna: Not that it needs to change again, but I'll just say there is no Firestore on its own without Firebase, which is why I included `firebase_` in the initial `icon` attribute value.
I also prefer the more succinct `firestore`, but just wanted to make the distinction that Firestore is a sort of child of Firebase, if that affects any existing patterns `vscode-icons` may be already using.

donmckenna: Fair enough.

donmckenna:
Thank you for doing that. I will be more observant next time.

donmckenna:
Thank you! And thank you for a great plugin :)

PR #1602, opened 2018-05-28
Remove 'wpilgenerator' from dependencies

JimiC:
This RP removes the dependency of the `wpilgenerator` from the project.

**Reasoning:**
`wpilgenerator` was created to be an automation script to publish the files and folders icons list, to our wiki from the CI. It shouldn't be part of the project in the first place, simply because, as it turned out, it was causing too much of a hassle to the average user, when trying to build the project. Not to mention the issue that arose on the CI builds for node v10.x, the security issues of the `nodegit` version used and the excessive build time of all jobs.

**Resolution:**
- `wpilgenerator` build and run scripts, on `v0.4.0` , have been reworked so that it can be installed as an `npm` package, without been publish to the `npmjs` registry.
- The `Travis-CI` config was adjusted, so to install `wpilgenerator` only on the build stage that it's needed. This resulted in resolving the build issues for node v10,x and reducing the build time of all jobs.

Additionally, `wpilgenerator` [v0.4.0](https://github.com/vscode-icons/wpilgenerator/tree/v0.4.0) has some extra goodies, like a fix when logging on non-TTY console and self-anchors at each file and folder name, so that we can now link directly to that icon when commenting in issues.

**Important Notice:**
As my experience with Node increases, I look back at code I have written and facepalm myself very often, realizing that although I know enough, I just know too little.

I like to apologize to all devs of this repo for the inconvenience I may have caused them.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
 [submodule "submodules/examples-generator"]
 	path = submodules/examples-generator
 	url = https://github.com/vscode-icons/examples-generator
-[submodule "submodules/wpilgenerator"]
-	path = submodules/wpilgenerator
-	url = https://github.com/vscode-icons/wpilgenerator
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,50 +14,47 @@ matrix:
   allow_failures:
     - node_js: node
   fast_finish: true
-addons:
-  apt:
-    sources:
-      - ubuntu-toolchain-r-test
-    packages:
-      - libstdc++-4.9-dev
 cache: yarn
 before_install:
   - curl --compressed -L https://yarnpkg.com/install.sh | bash
   - export PATH=$HOME/.yarn/bin:$PATH
 script: npm run build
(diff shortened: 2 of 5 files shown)


JimiC:
@vscode-icons/development Not all vulnerability issues have been resolved with `nodegit`, simply because the `nodegit` team dropped support for `node v7` since `0.21.2`. I was forced to use the latest version that supported `node v7` which unfortunately includes security issues. As soon as `vscode` moves to `node v8` we will be able to address those.

jens1o:
I think they switch very soon: https://github.com/Microsoft/vscode/issues/45542

Electron:   2.0.2
Node:       8.9.3
Chromium:   61.0.3163.100

I wonder why they use this version and not the LTS version..

JimiC:
They are forced to use whatever `Node` version comes bundled with `Electron`.

jens1o:
I know. I wonder why electron uses this version...

PR #1621, opened 2018-06-10
Fix babelrc extension

zalishchuk:
**Changes proposed:**
Rename `babelrc.js` to `.babelrc.js`, because `baberc.js` filename is not supported according to [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) specification.

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


robertohuertasm:
Maybe it would be better to keep the former extension for retro-compatibility's sake. What do you think @JimiC, @jens1o?

robertohuertasm:
By the way, thank you @zalishchuk for this PR! 😉

zalishchuk:
As it turned out, Babel is not using cosmiconfig to load configuration files. According to [this](https://github.com/babel/babel/blob/master/packages/babel-core/src/config/files/configuration.js#L19) file, icon should be supported on these filenames:
- `.babelrc`
- `.babelrc.js`
- `.babelignore`
- `babel.config.js`

JimiC:
Backwards compatibility should be used whenever possible.

zalishchuk:
@JimiC check this out - https://github.com/vscode-icons/vscode-icons/pull/1621/commits/3747ce011d63885cfd16494a1c33893b859b9f9d

PR #1622, opened 2018-06-10
Add better support for Django

batisteo:
**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

Add the file type `django` for `django-html`.

The file type is introduce by the/my [extension vscode-django](https://github.com/vscode-django/vscode-django) which avoid shadowing the `html` file type.


JimiC:
Why don't you add [django-txt](https://github.com/vscode-django/vscode-django/blob/master/package.json#L41) as well, while you are at it?

batisteo:
@JimiC I actually wanted to, and then reverted my changes fearing that would slow the merging process.
I’ll happily add it!

JimiC:
You do not need to declare new `ILanguage`. A simple array of `ids` on the existing declaration will suffice.

JimiC:
Make 
```
django: { ids: 'django-html', defaultExtension: 'html' },
```
to
```
django: { ids: ['django-html', 'django-txt'], defaultExtension: 'html' },
```
and you are donw. Don't worry about `defaultExtension`. It's a helper attribute for our examples generator tool. Doesn't play any role in the extension.

batisteo:
Oh I see, cool thanks!

batisteo:
@JimiC Is it good like that?

_(Is the last check hung?)_

batisteo:
@robertohuertasm I hope I’m not pushy if I ask what’s blocking for a new release. :slightly_smiling_face:

robertohuertasm:
Don't worry! We were just discussing that. That's been my fault for this release (too much things going on outside of Github) but hopefully this will be solved with the new changes introduced by @JimiC allowing new releases from any team member.

The new version will be released today or tomorrow!

PR #1663, opened 2018-08-01
support for wasm

robertohuertasm:
_**Fixes #1641**_

**Changes proposed:**

- [x] Add

Adds support for WebAssembly. 

I've added `.wasm` as a extension not entirely controlled by language Id because not all the vscode extensions support it (they mainly support `.wat`) and I've this kind of files as a subproduct during my Rust + WASM development. Even though I don't care enough about those files to install a new extension I think is nice to see the icon there.

PR #1664, opened 2018-08-01
Preparing 7.25.0

robertohuertasm:
This PR is build on top of #1663 so we'll need to merge & approve it before doing the same thing with this.


JimiC:
There is already a branch called `prepare-next`. Why create another one?

PR #1667, opened 2018-08-02
Add prettier to keep a uniform format in our code

robertohuertasm:
I added a precommit hook so `prettier` is automatically applied to all staged files before pushing any code. By doing this we don't have to care anymore about code format.

I've also applied `prettier` to our current `*.ts` files.


JimiC:
Not a fan of `Prettier`. Creates more issues than it solves.

robertohuertasm:
Why do you say so? I've been using a lot this last year and I haven't found any issue.

JimiC:
Tried to use it and my experience was negative. But I don't object in using it if this is what you prefer,

robertohuertasm:
I would encourage you to use it and give it a try again! 😉 Let's try it. If it presents an issue I will be the first to remove it.

JimiC:
Also, these lines (https://github.com/vscode-icons/vscode-icons/blob/master/.editorconfig#L19-L21) can be removed as `vsce` now uses `npm version` when updating the `package.json` (which preserves the used indentation).

PR #1684, opened 2018-08-16
Add .mjs extension for testjs icon

zo:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

### DIFF
--- a/src/icon-manifest/supportedExtensions.ts
+++ b/src/icon-manifest/supportedExtensions.ts
@@ -2540,7 +2540,7 @@ export const extensions: IFileCollection = {
       icon: 'testjs',
       extensions: [],
       filenamesGlob: ['test', 'spec'],
-      extensionsGlob: ['js', 'jsx'],
+      extensionsGlob: ['js', 'jsx', 'mjs'],
       light: true,
       format: FileFormat.svg,
     },


JimiC:
I was just reading about `[mjs](https://medium.com/passpill-project/files-with-mjs-extension-for-javascript-modules-ced195d7c84a)` and thinking maybe it's best to assign them a different icon, so to make them distinguishable from the vanilla `js` ones. Something in the line of the official `JS` icon but with a different background color, like [this](https://cdn-images-1.medium.com/max/2000/1*siP0U2MiFbV0teXwSKyuRw.jpeg).

JimiC:
And additionally `test.mjs` file extension should get its own icon too.

robertohuertasm:
TBH, I'm not really sure about creating a new icon for `.mjs`. In the end is just plain `js` and it's indeed supported by the `js` language id. I mean, in `.jsx` for instance we have some sort of DSL, so it makes sense to put a specific icon. But in ES6 modules, we're just writing vanilla JavaScript. We don't have icons for CommonJS, AMD or UMD modules, for example.

zo:
I'm glad to do a small contribution here ;)
I agree that for now modules icons should look like `js` ones, the appeal to use `.mjs` is already too low to also give away the familiar "javascript yellow"

PR #1714, opened 2018-09-12
.bazel.rc and .bazel support

### LINKED ISSUE #1713: .bazel.rc and .bazel support
[bazel](https://bazel.build/) files are already supported but `.bazel.rc` and `bazel.rc` are missing.

wayou:
resolve #1713


wayou:
hi @JimiC, thanks for the review.
did you mean this:
```diff
+   extensions: ['.bazelrc', 'bazel.rc','bazel.bazelrc'],
-      filenamesGlob: [
-       'bazel',
-    ],
-   extensionsGlob: ['bazelrc'],
```

JimiC:
Yes. Since `filename` attribute is set to true all entries in `extensions` will be considered as filenames. In order to support a filename extension, you will have to add another `bazel` icon declaration without the `filename` attribute. We have plenty of other cases in there.

robertohuertasm:
Can you ellaborate please

JimiC:
#3 of [comment](https://github.com/vscode-icons/vscode-icons/pull/1714#pullrequestreview-154660521) is misseing

robertohuertasm:
It's a filename, isn't it? 
image

JimiC:
I'm talking abiut `#2` of your screeny.

robertohuertasm:
I remove it, sorry, bad example. I think #2 is not a extension but a filename.  But I'm not sure.

JimiC:
And who knows, maybe they use also something like `baze.*.bazelrc` (where * can be one of: 'dev', 'prod', 'debug' etc).

JimiC:
And that's why I wanted the extension author to jump in and take care of this.

robertohuertasm:
I've reviewed the docs and can't see a mention of a file extension. I've even searched through github repos and no clue about this as a file extension.  According to what I can see in the docs it still looks good to me.

robertohuertasm:
[#1714 comment](https://github.com/vscode-icons/vscode-icons/pull/1714#issuecomment-421101958) is a pure conjecture and you cannot infer from the quote that the second one is a file extension. Provided that it's repeated also in the third line I think it's not very wild to guess that it's just a filename. 

I prefer to publish this as it is, even at the risk of being restrictive (if later on we discover that this was, indeed, a file extension) than the other way around.

robertohuertasm:
The only think I can see clearly is that we should have to learn `Bazel` to be 100% sure 😜

JimiC:
OK. I rest my case.

PR #1737, opened 2018-10-01
Update Dart language icon per brand refresh

timsneath:
The Dart language has a new brand in honor of Dart 2 (c.f. https://www.dartlang.org/). This PR updates the Dart icon with the new logo. The primary difference is a color change, but there are also some minor geometric changes.


JimiC:
Nice, though we will have to adjust the icon to our standards.

PR #1741, opened 2018-10-05
add Azure Pipelines correct icon

### LINKED ISSUE #1725: update azure-pipelines logo

update the azure-pipelines icon to the one microsoft uses.

[Icon](https://docs.microsoft.com/en-us/azure/devops/_img/index/devopsiconpipelines96.svg?view=tfs-2018)

vtbassmatt:
_**Fixes #1725**_

**Changes proposed:**

- [x] Fix

I'm a PM on Azure Pipelines. Here is a corrected icon for our product, which is distinct from Azure and only tangentially related to `azcli`.


src/icon-manifest/languages.ts
```diff
@@ -16,6 +16,7 @@ export const languages: ILanguageCollection = {
   autoit: { ids: 'autoit', defaultExtension: 'au3' },
   avro: { ids: 'avro', defaultExtension: 'avcs' },
   azcli: { ids: 'azcli', defaultExtension: 'azcli' },
+  azure_pipelines: { ids: 'azure-pipelines', defaultExtension: 'azure-pipelines.yml' },
```
JimiC: Can you point us to the `vscode` extension that uses `azure-pipelines` as `languageId`?

vtbassmatt: It hasn't shipped yet, and the repo isn't public. I can add you or someone else who needs to verify: https://github.com/Microsoft/azure-pipelines-vscode

If you want to wait until it ships before merging this PR, I'm 👌 with that. I expect to ship it imminently.

JimiC: Nope. We are fine to early support it. Thanks for letting us now.

vtbassmatt: (Repo will be public when we ship!)

vtbassmatt:
If you describe the changes, I can ask one of our designers to do it. I saw the 2px margin requirement but possibly screwed that up. Anything else?

JimiC:
It will be easier and quicker if I do it. No worries.

JimiC:
If `azure-pipelines.yml` is declared as a filename in the `languageId` then you don't need to declare it also in the `extensions` attribute. `vscode` is going to pick that up from the contribution.

vtbassmatt:
Gotcha. It is declared that way in the extension.

JimiC:
Come to think of it again, until the `vscode` extension gets released let's support the icon with the explicit extensions declaration. We will remove it once the `vscode` extension is published (assuming we will remember to do it :smile:)

PR #1765, opened 2018-11-01
Add alternate extension-based support for ActionScript files

admvx:
There was language-based support for ActionScript files already, via the [nextgenas extension](https://marketplace.visualstudio.com/items?itemName=bowlerhatllc.vscode-nextgenas), but this adds a second version (disabled by default) for `.as` files, which aligns more with Adobe's branding.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


JimiC:
You should add the `languageid` and remove the file extension. This way someone can still benefit from the syntax highlighting and the alternative icon.

admvx:
Ah, sorry for missing that! I switched out the `languageid` as I personally don't use any specific VSCode extension with ActionScript, but I can remedy that separately.

PR #1770, opened 2018-11-06
Add new supported folders aliases

mgcrea:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Adds several useful folder icon aliases.

### DIFF
--- a/src/iconsManifest/supportedFolders.ts
+++ b/src/iconsManifest/supportedFolders.ts
@@ -11,7 +11,7 @@ export const extensions: IFolderCollection = {
     { icon: 'app', extensions: ['app', '.app'], format: FileFormat.svg },
     {
       icon: 'asset',
-      extensions: ['assets', '.assets'],
+      extensions: ['assets', '.assets', 'files'],
       format: FileFormat.svg,
     },
     {
@@ -57,7 +57,7 @@ export const extensions: IFolderCollection = {
     { icon: 'cypress', extensions: ['cypress'], format: FileFormat.svg },
     {
       icon: 'db',
-      extensions: ['db', 'database', 'sql'],
+      extensions: ['db', 'database', 'sql', 'data'],
       format: FileFormat.svg,
     },
     { icon: 'debian', extensions: ['debian'], format: FileFormat.svg },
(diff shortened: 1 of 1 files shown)


JimiC:
I'm gonna leave the approval of this PR to @robertohuertasm.

src/icon-manifest/supportedFolders.ts
```diff
@@ -11,7 +11,7 @@ export const extensions: IFolderCollection = {
     { icon: 'app', extensions: ['app', '.app'], format: FileFormat.svg },
     {
       icon: 'asset',
-      extensions: ['assets', '.assets'],
+      extensions: ['assets', '.assets', 'files'],
```
robertohuertasm: I'm not sure of how `files` alias relates to `assets`. What's the rationale? I see this more of a subfolder of assets.

src/icon-manifest/supportedFolders.ts
```diff
@@ -206,7 +206,7 @@ export const extensions: IFolderCollection = {
     { icon: 'ravendb', extensions: ['ravendb'], format: FileFormat.svg },
     {
       icon: 'route',
-      extensions: ['route', 'routes', '_route', '_routes'],
+      extensions: ['route', 'routes', '_route', '_routes', 'routers'],
```
robertohuertasm: What about this? Can you provide a reference to a renowned project using this? I cannot see why someone would use a `routers` folder. Is it because you use more than one `router implementation` in your client applications?

src/icon-manifest/supportedFolders.ts
mgcrea: On some projects I used to use files in place of assets, but I agree it can/should be a subfolder of a more generic assets folder.

src/icon-manifest/supportedFolders.ts
mgcrea: Main use case for me is for backend [express](https://expressjs.com)-powered applications when you often have a lot of [routers](https://expressjs.com/en/4x/api.html#router) that you attach to your main express app.

src/icon-manifest/supportedFolders.ts
robertohuertasm: I mean, don't get me wrong, it can be whatever you want, but for that we have customization support. We only tend to implement generalized usages.

src/icon-manifest/supportedFolders.ts
robertohuertasm: I see, I was only taking into account client apps and didn't thought about `express`. You totally have a valid point there. 😉

src/icon-manifest/supportedFolders.ts
robertohuertasm: @mgcrea if you change this, we'll include it in the next release for tomorrow.

robertohuertasm:
Please resolve the conflicts if possible. For next collaborations I'd suggest you to don't do your changes on the `master` branch. It gives unnecessary amounts of works when trying to modify your local changes from our side.

PR #1810, opened 2018-12-16
Disable 'before_install' step from docker stages

JimiC:
Minor change/fix.


JimiC:
Bollocks, forgot to compile `wpilgen`. 3rd time is the charm.

PR #1820, opened 2018-12-27
Update devDependencies

WarenGonzaga:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

**Note**: Just update the devDependencies :)  I would like to add an action to the "Changes Proposed" for "Update" so it would like this...

 - [x] Update

That's it, thanks!


robertohuertasm:
Sorry but these dependencies were OK as they were. Thanks anyway for this PR.

PR #1838, opened 2019-01-20
Add support for .webp files

uu1t:
**Changes proposed:**

- [x] Add

Add support for .webp image files, which have been officially supported (Microsoft/vscode#35502).


JimiC:
I believe that it would be best to assign that to a new icon ([icon](https://developers.google.com/speed/webp/images/webp-devsite-icon.png)) since `png`, `jpg` files and `webp` ones can reside side by side.

uu1t:
Sorry for replying late.
I added a WebP icon which is downloaded from [Wikipedia](https://en.wikipedia.org/wiki/File:Webp-logo-wordmark.svg) and modified by myself.

I found that there is transparent blur around the icon and it seems smaller than other icons when resized to 28x28 px.
jpg-vs-webp
Should I fix it?

JimiC:
I will review the icon and fix it if need be.

uu1t:
Thanks, but I recreated a little bigger one whose content width is almost 28 px.
Please review new one.

PR #1843, opened 2019-01-24
Make greenkeeper ignore `vscode` dependecy

JimiC:
No comment


JimiC:
I'll try in the upcoming days to resolve `sinon` update issues.

PR #1845, opened 2019-01-25
Fix Jenkins default extension

tempora-mutantur:
Change this to Jenkinsfile.
This is the default extension for Jenkins multibranch pipelines.
see:
https://jenkins.io/doc/book/pipeline/jenkinsfile/

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


tempora-mutantur:
@JimiC Thank you, eventually I was thinking that it started working, but now I found out that the other file extension was set.

PR #1863, opened 2019-02-04
Added Support for AngularDart plus extensions

### LINKED ISSUE #1861: [Icon Request] Support for AngularDart plus extension files
- [X] I'm submitting an icon request:
  - Type: `filename`
  - Icon Name: `angulardart, ng_*_dart`
  - Sample original Icon: logo
  - Filenames: `ng_component_dt, ng_guard_dt, ng_service_dt, ng_directive_dt, ng_pipe_dt, ng_routing_dt/ng_routes_dt`

- More info: [AngularDart](https://webdev.dartlang.org/angular)

You guys kinda did the same with the Angular files at this issue: #1016. I am making the custom AngularDart svg and will post soon an image with all the icons following the guidelines. 😄 

Note: There is no **Module** file at AngularDart since all the providers, imports, and so on are already placed at the component.

KingDarBoja:
_**Fixes #1861**_

**Changes proposed:**

- [X] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Just added the Dart icon at the right bottom side of the current angular icons. I still provided all of them just in case. The little white line behind the Dart icon was to give it a slighty difference with the big Angular one when having same colours as happens with the ng_service one.


JimiC:
As @robertohuertasm mentioned [here](https://github.com/vscode-icons/vscode-icons/issues/1861#issuecomment-460091534) AngularDart icon is not necessary and better remove it.
The icons need adjustments to the `dart` icon. Specificaly, the guideline is that the icon should not pass over the horizontal line of the letter `A` but it extend to the bottom-right up the 'shield' outer limits.

JimiC:
Also to prevent future conflicts the names should be using the `dart` word instead of `dt` (i.e. `ng_component_dart`). The reasoning is that the file extension is `dart` and not `dt`.

KingDarBoja:
> Also to prevent future conflicts the names should be using the `dart` word instead of `dt` (i.e. `ng_component_dart`). The reasoning is that the file extension is `dart` and not `dt`.

Okay, I will rename it as soon as I get time, same goes for removing the AngularDart one :+1: 

> As @robertohuertasm mentioned [here](https://github.com/vscode-icons/vscode-icons/issues/1861#issuecomment-460091534) AngularDart icon is not necessary and better remove it.
> The icons need adjustments to the `dart` icon. Specifically, the guideline is that the icon should not pass over the horizontal line of the letter `A` but it can extend to the bottom-right up to the 'shield' outer limits.

Ah, basically place it at the limits of the shield without hidding the horizontal line. Gonna try shrinking the Dart logo to see how it looks but I am afraid of it being too small to be distinguishable.

robertohuertasm:
Great job @KingDarBoja with those `dart` icons!

PR #1868, opened 2019-02-10
Add support for reasonML

### LINKED ISSUE #1854: [Icon Request] Support for ReasonML
  - Type: `.re`, `reason`
  - Icon Name: `reasonml`
  - Sample original Icon: logo
  - Extensions: `re`
  - Filenames: `somefilename.re`
  - Language ids: `reason`

- Language Extensions: vscode-reasonml, reason-vscode

KingDarBoja:
_**Fixes #1854**_

**Changes proposed:**

- [X] Add
- [ ] Delete
- [X] Fix
- [ ] Prepare

Added the reason interface icon for `.rei` extension and fixed the extension for `.re`.


JimiC:
https://github.com/vscode-icons/vscode-icons/issues/1854#issuecomment-462169121

robertohuertasm:
@JimiC, I'm not sure if this PR is completely unusable taking into account the support for `Reason Interfaces` it provides. 

It seems that the extensions are only providing a common `language id` for these two kind of files thus making both having the same icon.

I think that having a separate icon would make a lot of sense for this case. Instead of changing the  `Reason` icon with the `REI` extension in it I think it would be better to slightly change the background color. After implementing this, we could maybe contact the `Reason` extension authors so they could provide support for a second `language id`.

robertohuertasm:
I'm going to reopen this until we arrive to a final conclusion. :)

JimiC:
My opinion is to contact the `Reason` extension author first, await the decision and then act accordingly. I don't want to have to deal with two PRs for one thing.

JimiC:
Btw this should be tracked on the issue and not on the PR. It's best to reopen #1854 and close this one.

robertohuertasm:
Well, we have some resolution at least. 😉 I'll put all this in the issue.

PR #1870, opened 2019-02-12
Add support for MySQL and MariaDB

### LINKED ISSUE #1822: [Icon Request] Support for MySQL
  - Type: `filename`
  - Icon Name: `MySQL`
  - Sample original Icon: mariadb
  - Extensions: `.mysql`

KingDarBoja:
_**Fixes #1822**_

**Changes proposed:**

- [X] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

To be honest, I am not sure about this one since MySQL official logo is a dolphin while this one is a Sea Lion (which I do prefer).

The good thing is that MariaDB Foundation has the same founder as MySQL, also being compatible with MySQL.


JimiC:
Check how the icon looks on light coloured themes. I suspect you will have to provide a darker icon version for those.

JimiC:
I like to apologize cause I didn't notice it earlier (didn't read the OP, mea culpa) but if we are to support this we have to go with the dolphin logo.

JimiC:
The sea lion can be used for `MariaDB`. Don't through the icons.

robertohuertasm:
I agree with @JimiC. In order to keep this, given that the job has already been done, and given that someone maybe prefers the `MariaDB` flavor, we can keep an empty entry for this logo (no file or language association) so the users can choose it and use it via customization.

jens1o:
I agree with you both.

KingDarBoja:
@JimiC @robertohuertasm So should I rename the current ones mysql icons (sea lions) to mariaDB so you guys can do the file association part, right?

robertohuertasm:
I've made a [quick search in the marketplace](https://marketplace.visualstudio.com/search?term=mysql&target=VSCode&category=All%20categories&sortBy=Relevance) for extensions supporting `msyql` as a language and I've found none. On top of that, I've never seen any file with the  `.mysql` extension. Normally people name them `*.sql`, regardless of the sql flavor.

This means that associating this icon to a language id which currently does not exist leaves this icon orphan an only usable from a customization perspective. 

Is it ok for you? @JimiC, @jens1o any thought about this?

I'm ok with adding these icons in case someone wants to use them or for further usage in case a new extension appears in the scene but I'm not sure of the premature `language id` association.

JimiC:
@robertohuertasm My thoughts exactly but I didn't want to discourage @KingDarBoja .

JimiC:
I want to take a look at the PR again tomorrow from work where I have a better workstation environment.

JimiC:
I changed the icons for `mysql` and `mariadb` and also added relative folder icons.
I also removed the `mysql` `language id`.
I believe this is the best way to support them.

PR #1879, opened 2019-02-17
Update dependencies

JimiC:
As promised, just in time for the new release.

Also resolves new `SinonJS` type definition issues in our tests.

Note: I reverted `@types/node` to `v7` as we are still supporting `vscode` versions with `7.9.0`.


JimiC:
@vscode-icons/development 
For the next release, suggesting to be a major increment "v9", we can drop the support of `vscode` versions with `node v7.9.0` and also drop `yarn`. Thoughts?

robertohuertasm:
I'm ok with your suggestion. I also sent an email to the team responsible of the marketplace to see if there's any option to change the `publishId` without losing our users.

PR #1886, opened 2019-02-19
Add support for QML & Qbs

oltodo:
Fixes https://github.com/vscode-icons/vscode-icons/issues/1887

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


JimiC:
Please open an issue before submitting a PR (https://github.com/vscode-icons/vscode-icons/wiki/Issue)

oltodo:
@JimiC Done

JimiC:
There is no extension that supports `qbs` as language id. `qbs` should be supported via `extension` reference.

oltodo:
Like this?

JimiC:
Still we need to do some adjustments to the icons but the code part is fine. Thanks.

oltodo:
I think I forgot something. In QML we generally create [`qmldir`](https://doc.qt.io/Qt-5/qtqml-modules-qmldir.html) files and it can be nice the `Qt` icon for them. Could I add something like below in the `supportedExtensions.ts` file?

```js
{
  icon: 'qml',
  extensions: ['qmldir'],
  filename: true,
  format: FileFormat.svg,
},
```

JimiC:
Do those serve a purpose like `h` (header) in `c++`? If so then make an icon with the same logo and different colour to distinguish them from `qml`.

oltodo:
Not exactly because it isn't linked to one specific file. It's just used to declare what modules/components can be found in a directory.

oltodo:
I've just added `qmldir` file with the `Qt` icon but using an orange color on background

AnabelSalomone:
Any news concerning this PR? It would be really helpful for my team, working with QML :)

robertohuertasm:
I cannot promise you anything but we'll probably have a new release this weekend or the next one.

JimiC:
I updated the icons to meet our art guidelines. I changed the letters in `qbs`, using the official font (which is [Titilium Web](https://fonts.google.com/specimen/Titillium+Web)), to be similar to those of `Qt`.
I also made them pixel perfect.

I hope this doesn't bother you.

oltodo:
@JimiC Thanks for updating SVG files, you did well 👍

PR #1897, opened 2019-02-26
Nestjs detection

### LINKED ISSUE #1324: [Icon Request] Support for Nestjs
Nestjs is a nodejs *framework* based on angular modular architecture, _(also built with ts ❤️)_, so it would require the same work as Angular projects as there are kind of the same extensions.

- [x] I want to create an icon request
  - Type: `extension`
  - Icon Name: `name`
  - Sample original Icon:  nest

  - Extensions: 
    - *.controller.ts
    - *.module.ts
    - *.service.ts
    - *.guard.ts
    - *.pipe.ts
    - *.decorator.ts
    - *.interceptor.ts
    - *.filter.ts
    - *.gateway.ts
    - *.adapter.ts
    - *.middleware.ts
  - Filenames: `nestconfig.json`
  - Language ids: `*ts* or *js*`

- More info: [docs]( https://docs.nestjs.com/)
> @kamilmysliwiec I think this could be nice for vscode user. Also, did I missed / added extensions ? _e.g. I was unsure about .entity.ts as this is more coupled to typeorm_
> edit: Also do you have svg available ?

KingDarBoja:
_**Fixes #1324**_

**Changes proposed:**

- [X] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

*NOTE:* I am missing the test files, I will updated it as soon as possible. Also, notice that all the Nest Icons are the same because I would like to hear the team opinion about the color palette & logo design  to be used in this case.


src/app/extensionManager.ts
```diff
@@ -39,14 +39,8 @@ export class ExtensionManager implements models.IExtensionManager {
...
 
-    // Check if Angular 2+ Project
     this.projectAutoDetectionManager
-      .detectProjects([models.Projects.angular])
-      .then(detectionResult => this.applyProjectDetection(detectionResult));
-
-    // Check if NestJS Project
-    this.projectAutoDetectionManager
-      .detectProjects([models.Projects.nestjs])
+      .detectProjects([models.Projects.angular, models.Projects.nestjs])
```
JimiC: Exactly what I wanted to suggest. Glad to see you figured it out by yourself.

KingDarBoja:
@JimiC I think eveyrthing is okay already but not sure why codecov is "failing" on checks.

Also, regarding the NestJS file icons, I will stick to the Angular colors used for the `ng_*` icons but with a gradient in order to keep the same style as the Nest logo.

JimiC:
Personally, I like the icons but those submitted are all the same and lacking the language type overlay icon.

Regarding `codecov` it fails because you introduced code that you have not covered with tests. Still, from a quick look, some code is duplicated in a way and it can be generalized (like `toggle*Preset` and `get*Preset`). I'll give it a more thorough look on the weekend if time permits.

KingDarBoja:
@JimiC Thank you for the feedback and yes, I didn't put the TS/JS overlay icon because I wanted to suggest the colors to avoid the extra work when saving the icons. I will submit a new image as soon as I get time to add the overlay. 

As stated on the original issue, there are like nine `nest_*` type files (adpater, middleware, module, pipe, etc) so that's why picking a color was a bit harder and instead used gradients.

Cheers!

JimiC:
1. I'm confused. I don't see the correct icons committed with the PR. I'm expecting to see the gradient versions with the `js` and `ts` overlay.
2. I won't be able to review the PR until the weekend of 17-18 March, so don't get discouraged.

KingDarBoja:
@JimiC Arghhh, my fault, forgot to push the changes into my branch when finished the icons, give me a few minutes 👍 

And no problem reviewing the PR, I can wait.

JimiC:
What concerns me about those icons now is their size. It exceeds the 2KB limit.

JimiC:
Hopefully, I'll review this in the morning.

JimiC:
The last commit concludes the work on this PR.
@KingDarBoja Thank you very much for the tremendous work you did. You truly made my life easier on this one.
If the rest of the @vscode-icons/development team is OK with the icons, we can proceed with the merging.

TristanMarion:
Thanks for this feature 😄

PR #1902, opened 2019-03-06
Add support for WPML config files

displaynone:
**Changes proposed:**

- [x] Add

Some WordPress Themes and Plugins include an [wpml-config.xml](https://wpml.org/documentation/support/language-configuration-files/) file so it works with WPML


src/iconsManifest/supportedExtensions.ts
```diff
@@ -3118,6 +3118,16 @@ export const extensions: IFileCollection = {
       extensions: ['doc', 'docx', 'docm', 'dot', 'dotx', 'dotm', 'wll'],
       format: FileFormat.svg,
     },
+    {
+      icon: 'webpack',
```
JimiC: You do realize that the `icon` name does not conform with the file name of the icon file.

displaynone:
@JimiC where can I find your art guidelines? I couldn't find it

I realized SVG width and height are wrong, but is there any other issue?

JimiC:
The art guidelines are scarred here and there I'm afraid. But by observing the existing icons you can get a general idea.

displaynone:
Thanks @jimic, crazy last weeks and I wasn't able to fix it

JimiC:
NP

PR #1978, opened 2019-04-18
Add support for APL

### LINKED ISSUE #1963: [Icon Request] Support for APL
  - Type: `languageId`
  - Icon Name: `apl`
  - Sample original Icon: N/A
  - Language ids: `apl`

- Language Extensions: https://marketplace.visualstudio.com/items?itemName=OptimaSystems.vscode-apl-language
- More info: https://en.wikipedia.org/wiki/APL_(programming_language)

KingDarBoja:
_**Fixes #1963**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


src/iconsManifest/supportedExtensions.ts
```diff
@@ -406,6 +406,22 @@ export const extensions: IFileCollection = {
       languages: [languages.apib],
       format: FileFormat.svg,
     },
+    {
+      icon: 'apl',
+      extensions: [
```
robertohuertasm: You don't need to specify the extensions as we're relying on the language id, which is already supporting these extensions.

icons/file_type_apl.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 32 32"><title>file_type_apl</title><path d="M30,28.275,16,2,2,28.275H14.162V30h3.676V28.275ZM17.838,24.826V13.161l6.215,11.665Zm-9.891,0,6.215-11.665V24.826Z" style="fill:#d2d2d2"/></svg>
```
robertohuertasm: Is this the proper color for the icon? It seems to be black although I guess you did it to support dark themes. If that's the case, I think it would be nice to add the black icon for the light themes.

src/iconsManifest/supportedExtensions.ts
KingDarBoja: Argh, I always forget that thing x.x! I will fix it as soon as possible.

icons/file_type_apl.svg
KingDarBoja: Yup. If that is the case, I will provide the black one for light themes within the next commit 🤙

Alhadis: Hey guys, just an FYI when using my icons: they're *all* black on purpose because their colours are assigned with CSS classes (each SVG gets packaged into an icon-font, so it's pretty much styled like ordinary text).

I vaguely recall some uncertainty about the Husky icon's colours as well. Just wanted to make sure that you know our icons are "colourless" on purpose. =)

(APL doesn't have an obvious colour, but GitHub [arbitrarily assigned](https://github.com/github/linguist/blob/9116c90fcbb82ac03b4b33c58cfbde1fcf745e99/lib/linguist/languages.yml#L102) `#5A8164`, which we use for the icon as well).

JimiC: We are aware. For `husky` we went with that color because mainly huskies are grey. For `APL` we couldn't find any official coloring, so we left it on purpose black.

Issue #1022, opened 2017-06-17
[Icon Request] Support for 'modules/models/controllers/router' folder


1312682:
- [x] I'm sure this issue is not a *duplicate*?
- [x] I want to create an icon request(if not, remove these lines below):
      - Type: [folder]
      - Icon Name: [modules, models, controllers, router]
      - Sample original Icon: 
https://maxcdn.icons8.com/Share/icon/Programming//module1600.png
https://openclipart.org/image/2400px/svg_to_png/202776/pawn.png
http://www.endlessicons.com/wp-content/uploads/2013/05/continuous-play-icon-solid.png
http://www.iconarchive.com/download/i91409/icons8/windows-8/Network-Router.ico
      - Extensions: modules/module, model/models. controllers, router
Thank you.

JimiC:
Please see #650.

PR #1985, opened 2019-04-19
Add support for modules folder

### LINKED ISSUE #650: [Request] Specific folder icons

KingDarBoja:
_**References #650**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

This is a proposal for `modules` folder using the icon provided at #1022. Thoughts?


JimiC:
I'm not 100% OK with the icon or the colors (something is bothering on it but I can't pinpoint it yet) but if the team agrees I'm OK with it.

KingDarBoja:
> I'm not 100% OK with the icon or the colors (something is bothering on it but I can't pinpoint it yet) but if the team agrees I'm OK with it.

Feel free to change the colors. Regarding the icon, I think it was a proper choice because couldn't think anything for modules besides multiple boxes linked together... unless it is better for shared folder icon?

robertohuertasm:
Personally, I don't dislike the icon. What I think is that the color combination is not the normal kind of combinations that we use. Normally we grab the color of the icon as a base for the color of the folder, but in this case I guess you chose a complementary. 

I would make it more monochromatic and choose whether purple or yellow as the base color for everything.

KingDarBoja:
@robertohuertasm I am working on it right now using the steps provided by JimiC on the other PR.

JimiC:
Looking at this in detail I believe the best course it to bind `modules` with the existing `components` folder icon. After all `modules` and `components` are similar in concept.

robertohuertasm:
Certainly, the components icon could work also for modules. But since we already have this cool icon I don't see a reason not to use it. Components folder has a specific meaning in the UI world and modules seems to me more a backend construct, so I wouldn't mind having two separated icons.

JimiC:
OK. I'm going to fix the contrast a bit. Don't merge yet.

JimiC:
I decided to leave the contrast as is. Will iterate on them if any issue arises.

PR #2004, opened 2019-04-21
Add support for certificate folder

JimiC:
_**References #650**_

**Changes proposed:**

- [x] Add


robertohuertasm:
What about changing the background color? Red seems strange given that we're talking about certificates and normally green is the color of choice for this.

JimiC:
It's difficult now for me to do any changes. Let it out of this release and we can iterate on it later.

robertohuertasm:
I will make a test

JimiC:
Use the `src` colors. Just copy paste the values in `fill`.

JimiC:
The `typings` folder colors are also a good choice.

robertohuertasm:
image

After testing with green colors, they seem too similar to `src` folder, and considering that nowadays lots of repos have this folder I thought it would be better to go with a more common color for locks.

JimiC:
I'm OK with it

PR #2008, opened 2019-04-21
Add support for mock folder

JimiC:
_**References #650**_

**Changes proposed:**

- [x] Add


KingDarBoja:
It looks scary if you ask me 💀 But works as intended (as `mock` reference).

JimiC:
We discussed it with @robertohuertasm over at `slack` and decided to go with this icon. Wasn't easy to find an icon that references `mock` in a clear way.

JimiC:
The idea behind the choice of this particular icon is that the `mask` is used to fool someone, just like `mock` does on the unit testing flow.

PR #2011, opened 2019-04-21
Add support for theme folder

KingDarBoja:
_**References #650**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


JimiC:
Actually, @robertohuertasm has a point here. The guideline is that the overlaying icon should not be larger than 24px in width and 21px in height.

PR #2013, opened 2019-04-21
Add support for interfaces folder

KingDarBoja:
_**References #650**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

The `link` icon fits really well since interfaces enforces the correct type of any variable/property, hence, you are "linked" to it.


KingDarBoja:
> I think this color is risky in dark themes. We should provide a light and dark version. @KingDarBoja have you tried this with dark themes?

I always use vscode in dark theme mode, so the icon looks okay to me. I used the same folder color as the [script](https://github.com/vscode-icons/vscode-icons/wiki/ListOfFolders#script) one.

interfaces

JimiC:
Nope. Too much gray. And the overlaying icon doesn't bring to mind `interfaces`.

JimiC:
I just came up with an icon concept. How about using an icon something similar to a train wagon connector.

JimiC:
@KingDarBoja In general, try to avoid the gray for folder color. Use only when there is no alternative.

robertohuertasm:
image
I prefer @JimiC's idea, too. It's what everyone familiar with `UML` would expect.

robertohuertasm:
I created a new version of the icon: 
image

JimiC:
Exactly my concept

KingDarBoja:
@JimiC @robertohuertasm I am not familiar with the UML concept since it is out of my university studies, instead got the idea from Typescript Interfaces definition to came up with the link icon. 

Anyway, glad we got another folder icon done :D

JimiC:
UML is great for architecture designing. It's worth exploring and learning it even for just the heck of it.

PR #2014, opened 2019-04-21
feat(azure-pipelines): add support for azure pipelines folder

robertohuertasm:
**Changes proposed:**
Azure pipelines supports templating and it's a common practice to store them in a separate folder.

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


JimiC:
Wouldn't it be prudent to add also the namings without the dash or even with a leading dot?

robertohuertasm:
I already added the one with a leading dot. Regarding the without dash naming I'll wait for such a request as the official naming is with the dash.

PR #2015, opened 2019-04-21
feat(coverage): support for coverage folder

robertohuertasm:
**Changes proposed:**
I've based the icon in this kind of coverage report for libraries:
image

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


JimiC:
I'm still not convinced that we should cover this folder name.

robertohuertasm:
That doesn't mean it doesn't exist. I mean, I don't usually delete it after running the tests so it's always there even it's not committed. Lots of people don't commit vscode or idea folders and we support it.

JimiC:
Not the best point but in order to not get into a meaningless debate, go ahead.

PR #2028, opened 2019-04-25
Update folder coloring

JimiC:
This PR improves the color contrast for some folders. In some cases, the icons have been changed. I have to admit that `mysql` is still not perfect but for dark themes, it's an improvement to the current.


KingDarBoja:
Love the new color design for most of them, the only one I am not so happy is the service one, I feel like there are too many folders using the gear icon already. 👍

robertohuertasm:
@KingDarBoja do you have any alternative?

KingDarBoja:
@robertohuertasm Perhaps something like these ones:
[Service option 1](http://cdn.onlinewebfonts.com/svg/img_362532.png)
[Service option 2](https://www.gardnerdenver.com/-/media/images/nash/icons/factory-service-icon-blue.ashx?h=360&la=en&mh=360&mw=360&w=360)

JimiC:
OK. Let's put here my argument for changing that icon.
1. It now has better contrast.
2. We already had an icon for `services` (the one I used) and the spirit is to use icons we already have instead of introducing new ones.

PR #2071, opened 2019-05-30
Added 'db' icon folder names

grantcarthew:
**Changes proposed:**

- [x] Add

I use a directory called `store` in my projects that is related to database storage.

This PR adds the following folder names mapping to the db icon:
* storage
* store
* dataset
* datasets

I don't know this project well so two questions worth noting:

* How do you get this approved? Is this change ok?
* Have I changed the right file in the right way?

Thanks for your great work.


JimiC:
The main concept is to support folder icons for folder names that are widely used in projects or frameworks. For personal use (meaning folder names that are particular to a project) it's suggested to use the customization feature of this extension. Check the wiki for more info.

If you can justify the addition of the folder names (besides the fact that you are using them), we are more than willing to add them.

grantcarthew:
Hi @JimiC 

I am well aware of that. I would imagine that `dataset`, `datasets`, `store`, and `storage` are all very common names for directories.

Do you disagree?

JimiC:
It's not that I'm disagreeing per say. It's just that I haven't seen those folders used. But that's just me. That's why I asked for any justification by providing any link to projects that use them. @robertohuertasm can give some insights.

robertohuertasm:
The way I see it, we only tend to incorporate widely spread usages. For instance, I wouldn't support adding `store` as a database icon as it is used in a different ways in other communities. It has other kind of usage for people using `redux` and alike state management patterns in the front end development. For me, that's a clear candidate for custom icons or for our Project Auto Detection functionality.

Regarding the rest, if you can somehow argue that these are used by someone aside your team, we could consider them as long as they don't conflict with other usages.

grantcarthew:
That's fine guys. I thought they would be useful for the wider community.

grantcarthew:
Just now I added a `blueprints` directory into a project and up comes a beautiful blue directory.
This is great. I love this extension.

Yet you guys think a database icon for a directory called `store` is not?

Scenario:

* I'm working on some random project.
* Need a directory to put some json files that hold some common datasets.
* Decide to call it `store` (or `datasets` etc) because `db` is being used elsewhere.
* BAM up comes a database icon (drum).
* Sweet

Google definitions:

* `store`: a quantity or supply of something kept for use as needed.
* `storage`: the action or method of storing something for future use.
* `dataset`: a collection of related sets of information that is composed of separate elements but can be manipulated as a unit by a computer.
* `database`: a structured set of data held in a computer, especially one that is accessible in various ways.

I'm opening this again because it just seems logical to me that if I add a store folder, something beautiful happens in my explorer tree.

JimiC:
@grantcarthew For `dataset` I'm imagining an icon with a grid in it and for `storage` something that resembles a drive. `Store` is really a vague word to use as a folder name in programming (personally I haven't stumbled upon it on any project) but that doesn't mean that it's not used somewhere. But still, I can't think of an icon that can describe its usage.

tl;dr I'm afraid I'm going to disapprove this PR for several reasons.

PR #2089, opened 2019-06-11
Haxelib icon should only be used for .haxelib

Gama11:
\+ add haxe_libraries (used by a Haxe package manager called lix)

### DIFF
--- a/src/iconsManifest/supportedFolders.ts
+++ b/src/iconsManifest/supportedFolders.ts
@@ -180,7 +180,7 @@ export const extensions: IFolderCollection = {
       ],
       format: FileFormat.svg,
     },
-    { icon: 'haxelib', extensions: ['haxelib'], format: FileFormat.svg },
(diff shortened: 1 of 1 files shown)


Gama11:
Urm.. well, the relevant docs section doesn't seem to mention what the folder is called exactly.

https://lib.haxe.org/documentation/per-project-setup/

But I can point you to the relevant sources:

https://github.com/HaxeFoundation/haxelib/blob/56390adc9c804463cda1dff279bf70be08743432/src/haxelib/client/Main.hx#L160

`haxelib newrepo` creates a folder named `.haxelib`, and it definitely needs the dot to be considered a library repository.

JimiC:
Looks like @robertohuertasm and you missed the dot when the PR was submitted (https://github.com/vscode-icons/vscode-icons/pull/224/files)

robertohuertasm:
I think at that initial stage in the plugin history the dot was automatically supported.

PR #2092, opened 2019-06-15
Updated vscode file icons with stable and insider versions (1.35)

### LINKED ISSUE #2087: [Icon Change] VSCode Icon

The latest version of VSCode (v1.35) has been released today and with it a new icon for the software has arrived as shown on the [changelog notes](https://code.visualstudio.com/updates/v1_35#_updated-visual-studio-code-icon).

Gonna make a PR as soon as I can. :small_airplane:

KingDarBoja:
_**Fixes #2087**_

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

Since the new `vscode` icon is the same for all platforms (check [this response](https://github.com/microsoft/vscode/issues/71827#issuecomment-495167013)), I replaced the `vscode2` icon with the insiders version just in case someone wants to use it. 

Also, this PR is not ready yet because I have to change the folder icon too.

I would like to ask if I could change the `vs` icon too for the new one.

**UPDATE:** Gonna need some help with the folder color for the `vscode` and `vscode2` @JimiC .


KingDarBoja:
> Do not replace. Add new.

So just append another number for the new versions like 3 & 4 or what? I don't get the idea.

JimiC:
Just like a list. New gets added on top. Old gets pushed one place down the list.

JimiC:
@robertohuertasm Don't merge until I resolve the icons. I'm out of town and it's difficult for me to coordinate this by phone.

KingDarBoja:
@robertohuertasm Let's wait for the folder background to be fixed by JimiC.

And of course, I would fix the `vscode2` icons to not use the insiders version and instead, provide them under a different key, like as `vscode-insiders`. If the `vscode-team` is not okay with that, I would just wipe them out and just keep the vscode ones.

Cheers!

JimiC:
I properly updated the `vscode` to the new icon.
@KingDarBoja I love the idea of adding the `insiders` version but now it's not the time to do it. I kept the icon though in case someone wants to use it via the customization feature. I intend (after the release of `v9`) to add a feature that detects the editor used and load the corresponding `vscode` icon. Until then we keep it as it is.

PR #2097, opened 2019-06-16
Add support for Jest Snapshot

### LINKED ISSUE #2091: [Icon Request] Support for Jest Snapshot
  - Type: `extension`, `languageId`
  - Icon Name: `Jest Snapshot`
  - Sample original Icon: logo
  - Extensions: `.js.snap`, `.jsx.snap`, `.ts.snap`, `.tsx.snap`
  - Language ids: `jest-snapshot`

- Language Extensions: https://marketplace.visualstudio.com/items?itemName=Orta.vscode-jest
- Language Definition: https://github.com/jest-community/vscode-jest/blob/master/package.json#L34-L46

KingDarBoja:
_**Fixes #2091**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


src/iconsManifest/supportedExtensions.ts
```diff
@@ -1664,6 +1664,12 @@ export const extensions: IFileCollection = {
       filename: true,
       format: FileFormat.svg,
     },
+    {
+      icon: 'jest_snapshot',
+      extensions: ['js.snap', 'jsx.snap', 'ts.snap', 'tsx.snap'],
```
JimiC: Why are you adding the extensions when they are already supported by the `language id` (https://github.com/jest-community/vscode-jest/blob/master/package.json#L37)?

robertohuertasm: I have mixed feelings here. By always using `language ids` were forcing people to install some extensions supporting this `language id`. Most of the times it makes sense, but many others the language id just serves the purpose of the extension. In this case, for example, this extension does something similar to what wallaby does. What if I don't want it because I use my own solution with tests but I still want to have my icons?

JimiC: Enable them with the custom feature. My reasoning is to keep the consistency to the rule we have kept all these years.

robertohuertasm: The consistency makes sense up to a certain level. What I mean is that it makes sense for extensions basically meant to serve the purpose of enabling syntax highlighting and so on, but just because a random or not so random extension has a language id I don't think we should automatically rely on this.

Basically, what I'm advocating for here is being practical and pragmatic. Me as a user I wouldn't understand having to decide between setting up a custom configuration or install an extension that provides something that I don't need and want. And I do think many others will think alike too.

The thing is that having rules is ok but there are cases like this one where we may reconsider how we apply them and whether it makes sense to follow them blindly. New scenarios are opened every day, and we must update our rules according to them (or at least open a debate around it :wink:).

JimiC: Then we should support it one way OR the other, NOT both.

robertohuertasm: I agree.

JimiC: I'm in favor of keeping a strict rule case. If we are going to have special cases then we should document them. Having some cases breaking the rules creates gray areas and blurry situations for any newcomer that wants to contribute to this extension.

robertohuertasm: @jens1o, @KingDarBoja what are your thoughts on this matter?

KingDarBoja: I do prefer keeping the same rule about installing a extension or using the custom feature to keep consistency. Saldy, most users doesn't know about this feature or get confused about how implement it, at least that's the idea I get by looking at related issues.

Cheers!

robertohuertasm: Thanks @KingDarBoja for your input. I'm not sure what you mean but "most users don't know about this feature", what are you referring to?

robertohuertasm: @JimiC regarding your concern about gray areas, I don't think that's an issue at all. We may have a general rule, which I'm not against it but that's why the PR processes are for. People may follow the general rule and then we can decide if a specific case falls into the gray area according to some very specific parameters. I mean, special cases may follow a rule, too. 

As an example, if the extension doesn't provide a core functionality (let's say syntax highlighting) and just adds an `optional` functionality (as testing niceties) I wouldn't rely on the language id for this. Even more if we consider that the snapshots are `Jest` files. Every user using Jest should be able to have Jest icons even if they decide not to use this particular testing extension, because they may use other tool for this, like Wallaby. In my opinion, we shouldn't be forcing users to install tools that they probably don't need just to get an icon.

JimiC: Regarding the PR process, this is exactly what I'm trying to avoid. Support on PRs (and in general) isn't my favorite thing to do. I like coding more. So I'm trying to avoid it in any way I can.

Regarding this case (Jest snapshot), you made your point from your first comment and I wasn't against it at all. I'm in favor of using one or the other (extensions or language id) to support it. And the main reason is my above statement. So let's go with extensions declarations and ditch the language id.

KingDarBoja: I came with a idea regarding @robertohuertasm opinnion but I am not sure if it is possible right now, here is:
Provide some kind of boolean to toggle between enabling the use of the icons if you want to use them without installing an extension. Something like the custom feature but automatically set it to every icon instead of letting the configuration to the end user.

JimiC: @KingDarBoja  That's not feasible. We rely on `language id` so that we don't have to deal with the file extensions and/or filenames it supports. Your proposal is out of the question.

JimiC: @KingDarBoja In order to not delay this any longer, can you please get rid of the `language id` implementation?

KingDarBoja: It's done, sir :dancer:

JimiC: Thanks.

robertohuertasm:
Changed it accordingly.

JimiC:
Why is `package-lock.json` in this PR?

robertohuertasm:
I did `npm -i` in order to test the changes. It happens that this is not my usual workstation so I had to download all the dependencies. I guess it was not up to date?

JimiC:
The is an inconsistency with `npm i` when you do it on a `Windows` machine and on a `Unix` one. If it's possible to revert the addition we will be forced to inspect it before the release.

robertohuertasm:
Done!

PR #2161, opened 2019-07-23
Support `.dockerignore` via `languageid`

### LINKED ISSUE #2104: [Icon Change] Support dockerignore via languageid
  - Language ids: `ignore`

- Language Extensions: https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker
- More info: https://github.com/microsoft/vscode-docker/blob/master/package.json#L726

JimiC:
_**Fixes #2104**_

**Changes proposed:**

- [x] Add


src/models/language/nativeLanguageCollection.ts
```diff
@@ -11,6 +11,7 @@ export interface INativeLanguageCollection {
   css: ILanguage;
   diff: ILanguage;
   dockerfile: ILanguage;
+
```
jens1o: what does this empty line mean? I thought like a separation(because `dockerfile` is the last entry starting with a `d`) however then it would not be semantic.

JimiC: First I added dockerignore in that file but since it's not natively supported I moved it and the extra line is a leftover.

PR #2166, opened 2019-07-29
Improve support for Sequelize

papb:
Improve support for [Sequelize](https://github.com/sequelize/sequelize). Currently only  `.sequelizerc` is supported, but `.sequelizerc.js` and `.sequelizerc.json` should also be supported.

**Changes proposed:**

- [X] Add


papb:
@JimiC Done! Thanks for the fast response! Hopefully it is correct now.

PR #2186, opened 2019-07-31
chore(release): prepare 9.2.0

robertohuertasm:


JimiC:
Since there are no new features shouldn't this be `9.1.1`?

robertohuertasm:
I thought so, but there are new translations, which are a new feature, the same as adding support for more extensions in an existing file extension. That's why the minor update.

JimiC:
Well, I can go with that.

PR #2189, opened 2019-08-03
Add support for Json folder

KingDarBoja:
_**References #650**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


JimiC:
We need to make two versions for this. As we have with the icon. The active icon should be the `{}` one.

JimiC:
Better test the icon on the editor. I suspect that the braces need to be thicker.

KingDarBoja:
> Better test the icon on the editor. I suspect that the braces need to be thicker.

I do see it fine, didn't touch the bracket besides resizing.

image

JimiC:
OK. Let's go with that and see.

robertohuertasm:
Why is the folder blue if the icons are yellow and gray? Normally we use the same color or pretty similar in a lighter tone. It's just my opinion, don't take it personally, but I don't like this color combination at all.

KingDarBoja:
> Why is the folder blue if the icons are yellow and gray? Normally we use the same color or pretty similar in a lighter tone. It's just my opinion, don't take it personally, but I don't like this color combination at all.

I don't remember why I picked this folder color, I think it was related to avoiding gray folder due to the second json icon being gray too.

Of course any update or suggestion is welcome, just not sure if yellow or gray would contrast very well with those icons.

robertohuertasm:
@KingDarBoja I've used the same colors that other icons with similar color. For `json` I used the `JS` backgrounds (it also kind of make sense). For the `json_official` I've used the same background as `include`, which sports a gray icon. There are similar ones like `mac` and the like which also use similar color palette. What do you think?

robertohuertasm:
We can make a release once we merge this one 🚀. I'll begin preparing the changelog!

KingDarBoja:
> We can make a release once we merge this one 🚀. I'll begin preparing the changelog!

LGTM too 🚀

PR #2222, opened 2019-08-31
Update C family icons

KingDarBoja:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

This PR add new icons for the C family. The ones being used are based on the official cpp logo being used at [Standard C++](https://isocpp.org/). You will recognize them as they have been around for several years and being used at most websites (C# I'm looking at you).

The C one was made using the old C logo at *The C Programming Language* book.


robertohuertasm:
I'm not particularly sure of this change. Specially for `C#` and `C++`. In my opinion, the former icon is way clearer. In the new ones I barely can see the `#` or the `++`. 

On the other hand, usually all `c#` devs are used to the green icon and not those new and bulky ones.

@JimiC, @jens1o, what do you think?

image

JimiC:
As I was the one that designed the icons, my main focus was the clarity of the representation of the icon, that what others are describing as `visual hook`.

The proposed set produces the opposite.

robertohuertasm:
If I'm not wrong, the old icons are the same or similar to the ones that are used in `Visual Studio` (it's been a while since I opened it but probably they remain the same). One of the rationales was to make the users coming from `Visual Studio` feel at home when dealing with `c#` projects and the like.

I propose to not get rid of these set but make it the second version instead of the principal version. Would that be ok?

KingDarBoja:
@robertohuertasm Makes sense to me to keep the same feeling for users coming from `Visual Studio` as the current icons are the same being used on that IDE.

My PR purpose is to add the official icon for the cpp one but as I mentioned before, the C# one is widely used on the web as _"official"_ icon (even if it isn't). For that reason, I also made the C one to keep the same styling as the others but was afraid of the `visual hook` as @JimiC said.

They look great to me to be honest, but he is right about the visual, their symbols are harder to seen if zoomed out unless you increase the font or windows size on `VSCode`.

Anyway, I do agree with roberto proposal to keep them as second version instead of principal one as everything comes to user taste.

Cheers!

JimiC:
@vscode-icons/development You do realize that with this change you are introducing breaking changes. If we are to keep backward compatibility we need to add the new icons last and not replace the existing icons.

JimiC:
@robertohuertasm I hope you are not merging it as it is.

robertohuertasm:
No, don't worry. I checked out if the changes were done and I see it's still in the same state.

KingDarBoja:
> No, don't worry. I checked out if the changes were done and I see it's still in the same state.

Done! Now the changes have been made as requested.

JimiC:
Nope. Still issues. `c`, `c2` and `csharp` should not have been changed. And this time it's not the 🍻  talking.

Issue #828, opened 2017-03-21
Support .jestrc extension for Jest


Thiht:
- [x] I'm sure this issue is not a *duplicate*?

Hi,

Following the issue https://github.com/vscode-icons/vscode-icons/issues/496 the extensions "jest.json" and  "jest.config.json" are supported for Jest config files. In the code: https://github.com/vscode-icons/vscode-icons/blob/95aadbbd92949a94af372effeb502d47f6821fb1/src/icon-manifest/supportedExtensions.ts#L178

Could you also support ".jestrc"?

Some real examples of existing .jestrc files for reference:

* https://github.com/gilbarbara/react-joyride/blob/master/.jestrc
* https://github.com/kuitos/es6-http-utils/blob/master/.jestrc
* ...

Thanks!

JimiC:
I'm not sure how we should proceed with this.
According to `Jest` [`.jetrc` is not yet officially supported](https://github.com/facebook/jest/pull/2445).

@robertohuertasm You have worked with `Jest` more than me. I leave this to you.

Thiht:
Officially Jest doesn't support a dedicated config file, other than package.json. It's possible to give it a custom config file but the name can be anything, hence the "jest.json" and "jest.config.json" currently supported by vscode-icon. I'm pretty sure these two filenames are not more standard than .jestrc.

JimiC:
@Thiht You are wrong. In early `Jest` docs the proposed way was the `package.json` and alternative `jest.json` or `jest.config.json`, hence the support.

Thiht:
I didn't know that!

I'm encline to believe you but I can't find [any reference on that](https://www.google.fr/search?q=jest.config.json&rlz=1C1CHBF_frFR734FR734&oq=jest.config.json&aqs=chrome..69i57j0.2895j0j4&sourceid=chrome&ie=UTF-8#q=%22jest.config.json%22&*), do you have one?

JimiC:
Unfortunately, I can't find the reference on the `Jest` docs. The page has been changed and there is no evidence anymore to support my claim.

robertohuertasm:
In my opinion, we could add this filename. Although not official there are some projects that are already using it and it may become eventually the preferred configuration file name. I see no harm in including it given that many people is using this convention.

Thiht:
Also I guess it's worth mentioning that it's very unlikely ".jestrc" will ever conflict with something else :)

PR #2237, opened 2019-09-10
Add more jest extensions

9KwHCUDP:
**Changes proposed:**

- [x] Add


robertohuertasm:
@9KwHCUDP can you provide more information about this addition? Are this supported or common `jest` config files?

robertohuertasm:
A quick search at Google for these extensions show almost no significant results at all. I'm under the impression that this may be a use case of custom extensions and we should close this. Waiting for @9KwHCUDP to make his/her case.

9KwHCUDP:
@robertohuertasm These are common jest config files, which are currently used by community. Since it doesn't support `.jsonrc` because config files should end with either `js` or `json`. Links: [#1](https://github.com/search?q=%22.jestrc.json%22&type=Code), [#2](https://github.com/search?q=%22.jestrc.js%22&type=Code)

JimiC:
https://jestjs.io/docs/en/configuration
>Jest's configuration can be defined in the package.json file of your project, or through a jest.config.js file or through the --config <path/to/js|json> option.

robertohuertasm:
These links you provided gave only 283 hits vs a total of 24484 for the official `jest.config.js|json` oficial extension, which roughly means that these extensions represent just the 1.15% of the total usage. It doesn't seem to me like a widely used extesions at all. 

I understand that some people are using it but, as @JimiC pointed out, it's not even the official extension. What do you think about this @vscode-icons/development ?

robertohuertasm:
@9KwHCUDP don't get me wrong. I really appreciate the time you took to do this PR. It's just that we have a policy of only allow documented or widely used extensions. If not, we have support for custom extensions so people can use the icon and extension combination they want.

9KwHCUDP:
Why does it include `.jestrc` then, it's also not official extension, and no one can use it, since jest can't even read it.

robertohuertasm:
This was a legacy thing. We kept it just because of backwards compatibility.

robertohuertasm:
Just for the sake of history #828. Here we were already debating if we should introduce or not the unofficial `.jestrc`. Back in the day, 2017, `Jest` was not so widely used but there was a significant amount of repos using this extension. That's why I voted in favor of including it. This just doesn't seem the same case to me. But let's hear what the other team members have to say if they want to comment on this. 😊

JimiC:
Basically, I'm in favor of: Since it's a custom config name it falls under the customization feature support. But this is so trivial that I don't mind merging it even if it's for the 1.15% of the repo base.

robertohuertasm:
Ok then. It goes in.

PR #2249, opened 2019-09-22
README.md: abbreviate "installs" badge

bmalehorn:
Kind of random, but saw this on the README and wanted to fix it.
"3644194 installs" is hard to parse; abbreviate to "3.64M installs".

### Before

image

### After
image


JimiC:
Glad to see I'm not the only one with O.C.D. Personally the long version doesn't bother me.

PR #2253, opened 2019-09-28
Add support for Gradle folder

KingDarBoja:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm:
@KingDarBoja I guess you've tested it but I'm wondering that giving that the folder color is light, the same icon may work both for light and dark themes. I'm referring to the light one with the elephant in dark green.

KingDarBoja:
> @KingDarBoja I guess you've tested it but I'm wondering that giving that the folder color is light, the same icon may work both for light and dark themes. I'm referring to the light one with the elephant in dark green.

I just want to avoid some elephant parts not being seen on dark themes, like the paws for example.

I will give a try and post a screen of how it looks.

Cheers!

KingDarBoja:
@robertohuertasm I am still not confident about how the dark gradle logo looks on dark themes even with the light folder as shown below:

Gradle folder dark theme

JimiC:
Use the light version of the elephant

PR #2268, opened 2019-10-14
Update Vlang logo

Delta456:
Update V lang logo as per [now](https://github.com/vlang)

Site would be updated with the new logo soon. 

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

PR #2269, opened 2019-10-14
icons: update Vlang logo

donnisnoni95:
> Update V lang logo as per [now](https://github.com/vlang)
> 
> The site would be updated with the new logo soon.
> 
> **Changes proposed:**
> 
> * [ ]  Add
> * [ ]  Delete
> * [x]  Fix
> * [ ]  Prepare

FIX #2268


donnisnoni95:
You're welcome ... Yes, please ... I also considering removing the shadow too ... but still looking at the community response.  👌 🙂

KingDarBoja:
> You're welcome ... Yes, please ... I also considering removing the shadow too ... but still looking at the community response. 👌 🙂

My opinion is that even though the shadow looks pretty neat on the icon, it will be harder to notice on smaller sizes (i.e 32x32). Maybe you can do this:

- Leave the shadow for bigger sizes.
- Drop the shadow for smaller ones (as I did for this extension).

Also can I provide some changes to it? I mean, the drop shadow can be optimized by using clipping mask on Illustrator instead of relying on a trapeziod shape hidden behind the left size as currently being done.

donnisnoni95:
@KingDarBoja Thank you... sure... PR is welcomed  👌 ✌️

PR #2270, opened 2019-10-15
Add file type icon for Systemd .service files

adamzerella:
Systemd is a popular Linux system management process system, it's a bit debatable if it technically has an offical icon. This icon was inspired by the image on the offical website: https://freedesktop.org/wiki/Software/systemd/.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


JimiC:
@KingDarBoja Support via https://github.com/bearmini/vscode-systemd-unit-file/blob/master/package.json#L20

adamzerella:
@JimiC Good find, I didn't notice that. I'll update the logo for this PR :)

KingDarBoja:
> @JimiC Good find, I didn't notice that. Should we update the logo in this MR?

Don't worry, I will do that 🍰

PR #2282, opened 2019-11-03
Add support for Python folder

KingDarBoja:
_**References #650**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

I added a general Python folder icon to support any kind of related folder like as `.venv` or `.virtualenv`.


robertohuertasm:
@KingDarBoja great work! Have you tried to play with one of the colors of the `python` logo? Usually we pick a variation of the hovering logo. The green color seems strange to me.

KingDarBoja:
> @KingDarBoja great work! Have you tried to play with one of the colors of the `python` logo? Usually we pick a variation of the hovering logo. The green color seems strange to me.

Yes, I did. I didn't use them at the end because the logo has gradients which doesn't play well with certain ranges of blue and yellow color. Still, any change on the folder color can be made but surely the green looks better in this case as some IDEs use that for icon overlay on top of the Python logo.

PR #2298, opened 2019-11-28
Add support for Nuxt folder

KingDarBoja:
_**References #650**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


JimiC:
Could the folder color get a little lighter?

KingDarBoja:
> A color change of the closed folder icon pending.

Need to re-install my editing software, not sure If I would make it before new year eve 🙈

robertohuertasm:
After testing with several green tones I think it's better to take the blue-gray color as a base:

image
image
image
image

The problem is not completely solved, as the line of the icon is really thin and not enough thick to get noticed in a clear way. 

Making the green lighter is not enough to make the great triangle visible. Darker works better for that triangle.

Let me know what do you think.

KingDarBoja:
@robertohuertasm Maybe we could slightly increase the thickness of each triangle in order to improve readability. I do feel like any folder color couldn't satisfy at all both dark and light versions with smaller lines as this icon does.

robertohuertasm:
I tested it and it wasn't really good either... I'll keep this PR and the one for `PERL` ([#2305](#2305 )) for the next release when we all come back from Christmas holidays so we can discuss it. 

In the meantime I'll prepare the changelog and prepare for the last 2019 release.

JimiC:
@vscode-icons/development I'll have a look at this today.

JimiC:
Doing various tests the only solution is to add an outline on the `nuxt` icon to make it more distinguishable. @KingDarBoja can you take of it?

KingDarBoja:
@JimiC Okie dokie, a white outline should do the job. I will get some free time to submit the new version today.

JimiC:
I would have done it myself but had to leave work early today

PR #2305, opened 2019-12-05
Add light version of Perl camel

### LINKED ISSUE #2300: Icon for perl is too dark for a dark theme.
- [ ] I'm submitting an icon request:

- Language Extensions: pl, pm, t
- More info: The icon is too dark for for a dark theme 
Screen Shot 2019-11-28 at 12 30 01 PM

KingDarBoja:
_**Fixes #2300**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


KingDarBoja:
@JimiC Any suggestion on how to add the mask? I do feel it follows the guidelines provided by the Perl / O'Reilly Foundation.

JimiC:
It requires me to fire up my design software and as I don't have a working machine atm this will have to wait.

robertohuertasm:
@JimiC could you be more specific about what exactly doesn't feel right with the light version? Is it the thickness of the contour line?

JimiC:
Basically we don't need a light version. The icon is visible enough in light themes. The issue reported is with dark themes. And since we adding the mask it can be used in any theme.

robertohuertasm:
Sorry, I said `light` instead of `dark`. So, if I'm getting this right, your suggestion would be to use the same icon (the one with the outline) for both kind of themes?

JimiC:
Yes. The concern is that some details of the icon are getting overlapped by the mask. But I could revise the icon on a later time (if I remember to do it).

KingDarBoja:
I could try to revise the icon as soon as possible, probably on this weekend.

And you're right, I should not have provided both dark and light version as the outline does the job. However, the outline was already part of the icon but filled with the purple color.

KingDarBoja:
@JimiC @robertohuertasm  Just to save you some time, here is the Perl icon on full size

Perl Camel

The gray outline was the same color as the camel on the original icon.

PR #2314, opened 2019-12-20
Webpack: add support for electron-forge

goosewobbler:
Just adding some additional webpack config file permutations to support the default config files provided by https://www.electronforge.io/.  

A rule like `webpack.*.js` would be ideal to cover these permutations and others, but I am assuming that this has been discussed before and is not possible or desired - since if it were, such rules would be prevalent across the project.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


JimiC:
Some info or justification is highly appreciated. Don't just drop a PR and think that we magically know your intentions.

goosewobbler:
added info

PR #2319, opened 2019-12-26
Added babel & gulp ESM files

ManzDev:
Added 'babel.config.esm.js' (babel es modules support) & gulpfile.esm.js (gulp es modules support)


xiaoxiangmoe:
@ManzDev What's `'babel.config.esm.js'`? I only find `'babel.config.mjs'` in babel's release note.

PR #2328, opened 2020-01-06
support mjs/cjs config file extension for eslint

xiaoxiangmoe:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

---

see https://github.com/eslint/rfcs/pull/43 and https://github.com/eslint/eslint/pull/12321


KingDarBoja:
From the above PRs, specifically at **configuring.md** changes on [eslint/eslint#12321](https://github.com/eslint/eslint/pull/12321):

> * **JavaScript (ESM)** - use `.eslintrc.cjs` when running ESLint in JavaScript packages that specify `"type":"module"` in their `package.json`. Note that ESLint does not support ESM configuration at this time.

So I don't see `.mjs` being included as well.

xiaoxiangmoe:
@KingDarBoja .mjs will be include in the future version.

evanplaice:
FYI, ESLint currently only supports CommonJS configs. `.mjs` would only be necessary if ESLint adds support for ESM configs.

PR #2329, opened 2020-01-07
Update actionscript language entry to match extension's new ID

admvx:
**Changes proposed:**
- [x] Fix

This adds a new `actionscript` language ID (cf. the existing one for `nextgenas`), to allow the secondary (more generic) AS icon to be used with this extension: https://marketplace.visualstudio.com/items?itemName=admvx.as2-language-support.


admvx:
Could you be specific about your preferred approach?

This follows the same pattern currently used by the three versions of gamemaker icons, so “totally wrong” seems to be at odds with the existing code base.

JimiC:
Open an issue and let us handle it.

robertohuertasm:
Hi @admvx, thanks for you PR. For what I've seen, your extension declares `actionscript` as the `languageId` which poses a problem, as `actionscript` is associated to the first set of icons, as you already spotted.

The problem here is that if you want to associate your extension to the secondary `actionscript` set of icons, you'll have to create a totally different `languageId` in your extension and add the whole entry again. Note that if you modify `languages.actionscript` for `languages.actionscript2`, people who were already using this secondary icon set will suddenly not see it (they should have to reconfigure their settings).

So, although you put `actionscript2`, this language id doesn't seem to match with the one in your extension so it will never work.

Let me know if that was your intention. The problem of changing your language id to something else is that it may affect how other icon extensions will deal with your extension. I think you can provide maybe more than one languageId but I ignore in case of multiple match which will prevail.

We can test it though.

admvx:
Hi @robertohuertasm, many thanks for your thoughtful response, and sorry for the confusion.

>So, although you put actionscript2, this languageId doesn't seem to match with the one in your extension so it will never work.

https://github.com/vscode-icons/vscode-icons/blob/c331d24d8c91b8c64a687961b44a3e00e0e510d3/src/iconsManifest/languages.ts#L5

Would I be correct to say that the outer identifier (e.g. `actionscript2`) is merely an internal reference for vscode-icons, while the actual comparison is based on the value of the `ids` field – specifically, it must match the 3rd party extension's languageId? This would explain why my PR appears to be functional in my local tests, due to the match with `actionscript` (as opposed to `actionscript2`).

I believe this means that the published version of vscode-icons has not been compatible with the 'ActionScript & MXML' extension for some time, since the author updated the languageId there from `nextgenas` to `actionscript` 8 months ago: https://github.com/BowlerHatLLC/vscode-as3mxml/commit/2e3fb2688202314ee74377659b797a3d80c4369e (When I test both extensions now, I no-longer see the associated icon.)
https://github.com/vscode-icons/vscode-icons/blob/26c4a0ae02043900ca9935491d31718f8636e52e/src/iconsManifest/languages.ts#L4

I mistakenly assumed the latter still used `nextgenas` as its languageId, so I wasn't anticipating a conflict between the two versions. Thanks also for pointing out the breaking change my PR would introduce for anyone already using the secondary icon.

In light of the current compatibility issue with 'ActionScript & MXML', I feel there are a few possible options, so I won't assume; please let me know how you'd like to proceed. My own extension is less than a week old, so I'm comfortable modifying its languageId at this point, if it would be useful to differentiate.

Thank you!

KingDarBoja:
@admvx Thanks for pointing out the `languageId` change on 'ActionScript & MXML' extension. I am pretty sure both extension can use the same language id as long as the end user only install one of them.

admvx:
Thanks @KingDarBoja - good point :)

Now that 'ActionScript & MXML' has the a generic language ID, maybe it's best if I keep the singular entry in `language.ts` (and update `ids` to `actionscript`) then remove my other changes?

While it wouldn't make the secondary icon to be default for my own extension, I'm happy to go with the simplest solution. I'll update my PR to that effect meanwhile.

KingDarBoja:
Thanks @admvx for the contribution, this is getting merged in #2336.

PR #2349, opened 2020-01-27
Add kitchen.yml as a Test Kitchen file

tas50:
The default config for Test Kitchen is no longer .kitchen.yml. We changed it to kitchen.yml about a year ago and those aren't picked up by this extension.

_**Fixes #IssueNumber**_

None

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


JimiC:
Thanks for letting us know (though a year delayed 😄)

PR #2350, opened 2020-01-27
Drop support for node v8

### LINKED ISSUE #2345: Drop Support for NodeJS 8

Node.js 8 has reached EOL as stated on the [official website](https://nodejs.org/en/about/releases/).

I think we should drop support for this version in order to avoid extra work on supporting old vscode versions.

JimiC:
_**Fixes #2345**_

**Changes proposed:**

- [x] Prepare

- Update dependencies
- Fix issues

@vscode-icons/development  Upon release, a major version bump is required.


JimiC:
@KingDarBoja Unfortunately more commits will be pushed as dependencies get updated often. Better approve when we are closer to release.

KingDarBoja:
> @KingDarBoja Unfortunately more commits will be pushed as dependencies get updated often. Better approve when we are closer to release.

No worries! As long as this is done in a standalone release (major version).

PR #2368, opened 2020-02-13
Fixes 'husky' with 'nvm' and 'Gitkraken'

JimiC:
Add `node` PATH set by `nvm` to work with `GitKraken`


JimiC:
Closing as false positive.

PR #2389, opened 2020-02-24
Add support for Nimble

### LINKED ISSUE #2387: Add Nimble package icon
- [x] I'm submitting an icon request

- Language Extensions: [vscode-nim](https://github.com/pragmagic/vscode-nim)
- More info: The nim language has a `.nimble` file which is indeed the package manager for it. It does has its own file icon and language id as stated on the above extension.

The icon can also be seen at [nimble directory](https://nimble.directory/) along with the [svg](https://nimble.directory/img/nimble_logo.svg).

KingDarBoja:
_**Fixes #2387**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


JimiC:
I don't like the black one. Better keep the gold-ish for all themes.

KingDarBoja:
The funny thing is that they don't have a gold-ish version for this icon, only the black one as the nimble website only supports light background 😆

JimiC:
Black is used as neutral.

KingDarBoja:
> Black is used as neutral.

No problem with that, I will keep the gold-ish for both themes.

PR #2399, opened 2020-03-04
Lossless image optimization

FlipFloop:
_**Fixes #IssueNumber**_

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


KingDarBoja:
@FlipFloop Did you just reduced the image size of every icon? 😮

FlipFloop:
@KingDarBoja Using SVGcleaner yes!

KingDarBoja:
> @KingDarBoja Using SVGcleaner yes!

Impressive, never heard of it. You mean this [svgcleaner](https://github.com/RazrFalcon/svgcleaner)?

@robertohuertasm Look at this! 🥇

FlipFloop:
@KingDarBoja Yes! Using software called ImageOptim (macOS only)

FlipFloop:
@KingDarBoja Thanks! Just tell me if you need my help with anything! Happy to help with whatever

JimiC:
Having a look at it the only thing that it does is to remove `xmlns`, `title` and use `lowercase` instead of  `uppercase`. `xmlns` is the namespace declaring that the content is `svg`. `title` is something we use when previewing the icon in a browser (the title is used in the tab title). Using `lowercase` instead of `uppercase` seems really trivial to me.

Concussion, I don't see any benefit in doing this. It would have been prefered to discuss this before proceeding submitting any PR.

robertohuertasm:
Just make a quick review of the changes but I saw that the `xmlns` is being kept:

image

Regarding the title, I guess we could change the way we get the title by using the title of the file?

If we end up reducing a little bit the size of the icons by applying this small set of measures I think it's a good trade-off.

JimiC:
Ahhh right, it changes the position of the namespace. The title is automatically read by the browser it's nothing we can do about it.

Now, comparing icons folder size before and after, here are the results:
Before: 3.5MB (6.1MB on disk)
After: 3.3MB (6.0MB on disk) 

5.71% size reduction (1.64% on disk)

In my opinion still not forth it.

JimiC:
Not to mention that we use the exact `export for screens` from Illustrator, which means that if we accept the PR we will have to add another tool in the pipeline.

robertohuertasm:
> Not to mention that we use the exact export for screens from Illustrator, which means that if we accept the PR we will have to add another tool in the pipeline.

That's another question, which we should discuss. I agree. 

What we're discussing here is what we do with this effort (@FlipFloop's PR). In my opinion, having a 5,71% reduction of our icons just for free it's acceptable if it only affects the `title` of the previews, which to be honest, I'm not sure were to find. is it here somewhere?:

image

As I'm not sure of what exactly the title issue implies I'm not "all in" to accept this change, but if it's a trivial thing, then my vote is a `yes`.

That being said, I wouldn't enforce the usage of yet another tool as a mandatory thing but as a recommendation. At least we must use  Illustrator's `export for screens`. Anything else, in my opinion, should be optional.

JimiC:
I mean this title usage.

Illustrator already exports the `svg` optimized for screens when using the `Export for screens` selection.

Bottom line, I'm leaving the final decision up to you. My vote is `No`.

KingDarBoja:
Arrghhh I got lazy and expected the savings were way more than I thought but looking at JimiC analysis, I have changed my mind and feel like it isn't worth the reduction unless we provide some way to load the title from the icon itself.

My vote is `No`.

FlipFloop:
Whatever y'all want!

robertohuertasm:
Ok then. Just for the record and in case any of you wants to reconsider :stuck_out_tongue_winking_eye:, personally I'd rather have this reduction than having a title in a browser. The reason is fairly simple: the cost in terms of effort is 0, it's already done and the reduction has a direct impact in the loading times of the tree view or our users while the existence of this title is something probably very few of our users will ever notice. Our icons are not there to be opened with a browser, and you always have the path with the filename in the url bar anyway.

That being said, feel free to close this. Thanks @FlipFloop for your time and dedication. :+1:

FlipFloop:
No problem! Please just tell me if you need anything. I’m super bored right
now and want to code stuff.

JimiC:
@robertohuertasm You do know that I'm in favor of any improvement as long as it's worth it. This improvement will not make a noticeable impact on the UIX of the user. We already have rules to keep the icons at a bare minimum size. If we could reduce a 2-3KB icon to 1KB that would be something worth adopting.

KingDarBoja:
It is possible to achieve a higher size reduction with the current icon settings? I don't think so...

FlipFloop:
If you want I can try again without the SVG compressions. However, i have
no idea if any image files are png, jpg or else

Le jeu. 5 mars 2020 à 10:37, Manuel Bojato <notifications@github.com> a
écrit :

KingDarBoja:
@FlipFloop Most of the icon files are `svg`, the only 3 icons are related to the extension docs.

JimiC:
I don't think there is anything to be done here. Icons are already optimized. One thing I could see of interest would be the support of images sprites but still, it won't add any significant benefit.

JimiC:
Closing as non beneficial.

PR #2445, opened 2020-04-08
Renovate / Add support for renovate.config.js

Djiit:
As the preferred way for using Renovate with their official Github Action is to use a `.js` file instead of a `json`, we should add `renovate.js` and `renovate.config.js`.

**Changes proposed:**

- [x] Add

### DIFF
--- a/src/iconsManifest/supportedExtensions.ts
+++ b/src/iconsManifest/supportedExtensions.ts
@@ -3147,7 +3147,7 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'renovate',
-      extensions: ['.renovaterc'],
+      extensions: ['.renovaterc', 'renovate.js', 'renovate.config.js'],
       filenamesGlob: ['renovate', '.renovaterc'],
       extensionsGlob: ['json'],
       filename: true,


JimiC:
This also adds support from `renovate.config.json`. Don't know if this is wanted.

Djiit:
Nope. How can I prevent it ?

JimiC:
Both need to be added in `extensions` instead.

JimiC:
But it conflicts with https://docs.renovatebot.com/configuration-options/.
Do you have any link that supports your claim?

Djiit:
Yes sure. A .JS file is used by their official Github Action : https://github.com/vidavidorra/github-action-renovate (strange it is not yet under their renovate org, but anyway they sent me here in this comment : https://github.com/renovatebot/renovate/issues/5881#issuecomment-609385924)

Djiit:
Updated the commit with your suggestions

JimiC:
I ended up digging into `Renovate` code (something I didn't want to do).
According to their code-base:
These are the supported config file names: https://github.com/renovatebot/renovate/blob/master/lib/config/app-strings.ts
which are checked here: https://github.com/renovatebot/renovate/blob/master/lib/workers/repository/init/config.ts#L38

But they also use several other `config` parsers, one of which is a file parser -> https://github.com/renovatebot/renovate/blob/master/lib/config/index.ts#L50

Now if you follow the code you will end up in the file config parser -> https://github.com/renovatebot/renovate/blob/master/lib/config/file.ts which in this line https://github.com/renovatebot/renovate/blob/master/lib/config/file.ts#L15 they use `require` (`commonjs` module loader) to load a `json` from a `js` file.

This is how they do it which is unconventional if you ask me.

Bottom line, either ask them to officially support `js` config files or change the `GitHub Action` implementation.

Djiit:
Allright :). Thanks for your complete anwser. I'll close this.

JimiC:
Furthermore, the `renovate Github Action`  should not use that "hack" and let `renovate` load the configuration from an official supported file, in my opinion.

JimiC:
Reading the `entrypoint.sh` of https://github.com/renovatebot/github-action/blob/master/src/entrypoint.sh the action doesn't work if you don't specify a configuration path. Totally unconventional.

JimiC:
This https://github.com/renovatebot/github-action/blob/master/src/entrypoint.sh#L21 should definitely be changed to support the official config filenames.

Djiit:
Hey, yes I totally agree with you. nvm, I'll reopen a PR here if things change on their side :)

PR #2454, opened 2020-04-11
10.1.1

JimiC:
:rocket: Prepare release 10.1.1


KingDarBoja:
Would be good to add #2455 after getting merged otherwise LGTM 🚀

JimiC:
@KingDarBoja #2455 is non-essential for the release.

PR #2473, opened 2020-04-29
added new folder extensions variants

3GOMESz:
added new folder extensions variants: 'application', 'webapp', 'dev' and 'development'

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

### DIFF
--- a/src/iconsManifest/supportedFolders.ts
+++ b/src/iconsManifest/supportedFolders.ts
@@ -8,7 +8,11 @@ export const extensions: IFolderCollection = {
   supported: [
     { icon: 'android', extensions: ['android'], format: FileFormat.svg },
     { icon: 'api', extensions: ['api', '.api'], format: FileFormat.svg },
-    { icon: 'app', extensions: ['app', '.app'], format: FileFormat.svg },
+    {
+      icon: 'app',
+      extensions: ['app', '.app', 'application', 'webapp'],
+      format: FileFormat.svg,
+    },
     {
       icon: 'arangodb',
       extensions: ['arangodb', 'arango'],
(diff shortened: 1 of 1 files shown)


src/iconsManifest/supportedFolders.ts
```diff
@@ -431,7 +435,7 @@ export const extensions: IFolderCollection = {
     },
     {
       icon: 'private',
-      extensions: ['private', '.private'],
+      extensions: ['private', '.private', 'dev', 'development'],
```
robertohuertasm, line 438: Sorry but I don't agree with having `dev` or `development` under the private icon. This is a very opinionated decision and should follow the `customization` [approach](https://github.com/vscode-icons/vscode-icons/wiki/Customization).

JimiC:
I too agree that `dev` and `development` are not represented correctly by the `private` icon. Furthermore, I also disagree with the `app` extensions. Especially `webapp` deserves its own icon in my opinion. 

In order to avoid such cases in the future, it would be best to first open an issue and discuss any proposal rather than submitting a PR that will end up rejected.

I'm closing this PR as "non-acceptable".

3GOMESz:
Sure, next I'll open an issue, sorry

> I too agree that `dev` and `development` are not represented correctly by the `private` icon. Furthermore, I also disagree with the `app` extensions. Especially `webapp` deserves its own icon in my opinion.
> 
> In order to avoid such cases in the future, it would be best to first open an issue and discuss any proposal rather than submitting a PR that will end up rejected.
> 
> I'm closing this PR as "non-acceptable".

PR #2527, opened 2020-07-04
Update Bloc folder icon

KingDarBoja:
**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

The [bloc library](https://marketplace.visualstudio.com/items?itemName=FelixAngelov.bloc) updated its logo last week.


felangel:
Thanks so much for doing this! One question, does it matter that the size of the icon was modified? Previously it was 40x40 but the new one is 34x34. I just want to make sure that was intentional 👍

KingDarBoja:
@felangel it is intentional, all the icons must be 28x28 whereas the viewport is 32x32 (2px of margin on each side) and the small icon inside the folder must not be greater than 20x20.

PR #2537, opened 2020-07-11
reduce bosque icon size

KingDarBoja:
**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


robertohuertasm:
Test fail but they will be fixed once we release as the prepare PR contains the fix for this issue.

PR #2557, opened 2020-07-18
Add support for AWK

### LINKED ISSUE #2411: [Icon Request] Support for awk icons
  - Type: `extension`
  - Icon Name: `awk`
  - Sample original Icon: Same as shell icon? Can't find original icon
  - Extensions: `awk`
  - Filenames: `awk`

KingDarBoja:
_**Fixes #2411**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

I used a png I found on the web and simplified it.


KingDarBoja:
> Remove leading dot of extension. (Why ppl keep adding that?)

I don't know, I think it is automatic because you thinking on extensions, which always are listed with the leading dot on other websites 😆

PR #2577, opened 2020-08-02
Add support for ReasonML BuckleScript

### LINKED ISSUE #2576: [Icon Request] Support for ReasonML BuckleScript 8 files
- [X] I'm submitting an icon request:
  - Type: `.res, .resi`
  - Icon Name: `reason`
  - Sample original Icon: logo
  - Extensions: `.res, .resi`
  - Filenames: `example.res, example.resi`
  - Language ids: `reason`

- Language Extensions:
- More info:

There's a new syntax for ReasonML that uses files ending in .res and .resi instead of .re and .rei. The existing ReasonML icon is fine, it just needs to be used for '.res and '.resi files as well as .re and .rei (at least for now, all four file extensions are supported in ReasonML).

Details at https://reasonml.org/docs/reason-compiler/latest/new-bucklescript-syntax

KingDarBoja:
_**Fixes #2576**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

### DIFF
--- a/src/iconsManifest/supportedExtensions.ts
+++ b/src/iconsManifest/supportedExtensions.ts
@@ -3222,7 +3222,7 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'reason',
-      extensions: [],
+      extensions: ['res', 'resi'],
       languages: [languages.reason],
       format: FileFormat.svg,
     },


robertohuertasm:
What should we do with this after #2600 ?

KingDarBoja:
I think we could close this pull request and also we need to remove the `res` extension from the `reason` declaration to avoid any conflict with new ReScript support.

robertohuertasm:
I think we don't have any extension in the reason declaration, so we could safely close this and move on.

KingDarBoja:
> I think we don't have any extension in the reason declaration, so we could safely close this and move on.

Oops, you are right, I didn't noticed this PR was adding both `res` and `resi` 🙈

robertohuertasm:
Feel free to close it. 👍🏻

PR #2603, opened 2020-09-01
support cjs config file extension for rollup

sant123:
related issue: https://github.com/rollup/rollup/pull/3204/files

_**Fixes #IssueNumber**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


src/iconsManifest/supportedExtensions.ts
```diff
@@ -3342,7 +3342,7 @@ export const extensions: IFileCollection = {
         'rollup.config.dev',
         'rollup.config.prod',
       ],
-      extensionsGlob: ['js', 'mjs', 'coffee', 'ts'],
+      extensionsGlob: ['js', 'cjs', 'mjs', 'coffee', 'ts'],
```
robertohuertasm, line 3345: Sorry but I'm not sure why we should support `cjs` as an extension. According to the PR you refer to I can only see that the output uses the `cjs` format but it doesn't mention the extension. 

Can you elaborate a little bit more?

robertohuertasm (changes requested):

src/iconsManifest/supportedExtensions.ts
sant123: Yes sure, please take a look at this link https://rollupjs.org/guide/en/#configuration-files

It says you can create a file config with `.js`, `.cjs` and `.mjs` extensions.

Referring to `.cjs` extension `will prevent Rollup from trying to transpile the file` and read it directly as a Node.js script.

robertohuertasm: Thanks!

robertohuertasm (approved):

PR #2605, opened 2020-09-04
add .bazel extension support for Bazel

CaringDev:
e.g. [googletest](https://github.com/google/googletest/blob/master/BUILD.bazel) uses the .bazel extension and [many other projects](https://github.com/search?l=&q=extension%3Abazel+load&type=Code) seem to do so as well.

**TODO:**
- [x] fix filename
- [x] add `starlark` language support

**Changes proposed:**

- [x] Add

### DIFF
--- a/src/iconsManifest/languages.ts
+++ b/src/iconsManifest/languages.ts
@@ -252,6 +252,7 @@ export const languages: ILanguageCollection = {
   sql: { ids: 'sql', defaultExtension: 'sql' },
   squirrel: { ids: 'squirrel', defaultExtension: 'nut' },
   stan: { ids: 'stan', defaultExtension: 'stan' },
+  starlark: { ids: 'starlark', defaultExtension: 'bazel' },
   stata: { ids: 'stata', defaultExtension: 'do' },
   stencil: { ids: 'stencil', defaultExtension: 'stencil' },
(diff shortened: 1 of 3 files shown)


robertohuertasm (changes requested):
Hi @CaringDev, thanks for this PR. Just a couple of comments:

- first of all, the change you made won't work as you would expect. That is a `filename` declaration, so it will only add the icon to a file named `.bazel`. I guess you wanted to have `.bazel` as a extension, like in `*.bazel`.
- secondly, the convention is to have a file named `BUILD.bazel` as you can see in your provided links and in the official [vscode bazel extension](https://github.com/bazelbuild/vscode-bazel/blob/master/package.json#L217).

My suggestion would be to modify `.bazel` for `BUILD.bazel` in ln 632 and probably add support for the `starlark` language as stated in [here](https://github.com/bazelbuild/vscode-bazel/blob/master/package.json#L195). This would involve creating a new `languages` property and add it to the array of languages. By doing that, anyone having the bazel extension installed would get the icons for the file extensions defined by the bazel extension.

Let me know if you need more help.

CaringDev:
Not sure my changes would achieve this:
> ... anyone having the bazel extension installed would get the icons for the file extensions defined by the bazel extension.

@robertohuertasm thanks for your support

robertohuertasm (approved):
Great job! Thanks for your PR!

PR #2606, opened 2020-09-04
Added wildcard .eslintrc.*

MerrickKing:
I use different eslintrc files for different file types (e.g. eslintrc.js.json, eslintrc.ts.json).

- [ x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


MerrickKing:
Testing with fine tuning my settings.json and the wildcard doesn't seem to work (doesn't follow proper glob rules...)

robertohuertasm:
Wildcards are not supported by vscode file extension system.

PR #2630, opened 2020-09-26
Update TSConfig with official alternative

### LINKED ISSUE #2626: Update TypeScript Icons for tsconfig.json

Since we've got the cool changes from https://github.com/vscode-icons/vscode-icons/issues/2581
It would be better if we have the same changes for tsconfig.json while enabling TS Official Icons:

Screen Shot 2020-09-22 at 16 49 09

KingDarBoja:
_**Fixes #2626**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Also updates the `tsconfig` extension with latest Angular 10 changes (`tsconfig.lib.json` and `tsconfig.lib.prod.json`).


JimiC:
Why?

JimiC:
@KingDarBoja If you want to implement this, then you need to introduce an `official` icon with the ability to switch between the two, as we do with the official `TS` icon.

robertohuertasm:
What do you mean? Sorry, you closed the PR before I could even pull the changes to test it locally but my understanding is that the `TS Config` icon will change with the TS switch. I don't see why would we introduce another switch just for this. Or am I misunderstanding something?

JimiC:
The `tsconfig` is not tied with the switching mechanism.

JimiC:
In this PR is gets tied but it needs to be separated.

JimiC:
I'll try to be more clear. This icon will have to have its own switching mechanism to suit all use cases.

robertohuertasm:
Let's discuss this in the issue, but I don't agree with what you say here. I explain my reasons there.

PR #2657, opened 2020-12-03
jest.config.ts configuration file in Typescript

carlocorradini:
**Jest Typescript**

**Jest** now supports *jest.config.ts* as configuration file.

See [https://github.com/facebook/jest/issues/9636](https://github.com/facebook/jest/issues/9636) for more information.

**Changes proposed:**

- [X] Add
- [ ] Delete
- [X] Fix
- [ ] Prepare


KingDarBoja (dismissed):
It only applies to `jest.config` but should not worry about `jest.config.babel` at all.

PR #2683, opened 2021-01-14
Update Rust icon

mehcode:
**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

Update Rust icon. The official Rust icon is no longer colorized and is a flat black color. 

https://www.rust-lang.org/

Source: https://www.rust-lang.org/static/images/rust-logo-blk.svg


KingDarBoja:
This will require some changes according to guidelines, no worries, I can take that.

robertohuertasm:
Hi @mehcode, we are aware of it, but the problem here is that we have to find a way to make the icon work with different background colors. Defaulting to black won't work for most dark themes. So maybe we can use the black one for the light themes for the moment. 

Btw, I'm a big fan of `sqlx`! Keep up the good work!

robertohuertasm:
Just curious about whether you're using a dark or a light theme. In any case, you can always change the icon at will by using the custom icon functionality: https://github.com/vscode-icons/vscode-icons/wiki/Custom

KingDarBoja:
My suggestion was to give the black one in dark backgrounds a small white outline 🤔 Not sure if that's okay but could work.

robertohuertasm:
I also tried the same and wasn't good. I would keep it like it is for the moment as it works for all kind of backgrounds.

KingDarBoja:
> I also tried the same and wasn't good. I would keep it like it is for the moment as it works for all kind of backgrounds.

Based on the Rust Discussion, using the light-gray is the best choice for dark themes

rust_dark

robertohuertasm:
In the docs they use the outlined...

image 

The light gray it's too much light for my taste. The thing is that that's why I used the oxide gradient in the beginning. It was safe for all backgrounds and was a color reference to Rust. 

I see the black one working flawlessly for light themes. For dark themes, I don't see a clear winner. I prefer the one we have right now or probably the one with the white shadow. We can maybe keep both of them and let people use it via customization as we already do for some other icons (not talking about providing a menu option, just the icon in there for people to use it via json customization).

KingDarBoja:
If you need the Rust with the white shadow (not the outline) let me know @robertohuertasm so I can share the svg here so you can update the PR 😄

robertohuertasm:
I've just set up the official black one for light themes and kept the old one for dark themes. Merging it like this for the moment.

PR #1862, opened 2019-02-03
Add support for Imba

KingDarBoja:
_**Fixes #1844**_

**Changes proposed:**

- [X] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

I added the V2 provided at the issue but had no idea how to add it to the iconsManifest.

PR #2684, opened 2021-01-17
Update Imba Icon (to version 2.0)

ericvida:
_**Fixes #1862**_

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

Update Imba Icon to match the branding of the new version of the [Imba Language (2.0)](https://imba.io)
Source: https://github.com/imba/branding-imba/blob/master/imba-icon.svg


src/iconsManifest/supportedExtensions.ts
```diff
@@ -1889,7 +1889,7 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'imba',
-      extensions: [],
+      extensions: ['imba','imba2'],
```
KingDarBoja, line 1892: This is not neccessary as the language id will automatically detect the extension.

robertohuertasm: Not sure the language id works anymore as the extensions in the marketplace don't link to their repo anymore and we cannot be sure that the id is valid. To be on the safe side, I'll keep them here.

robertohuertasm (approved):

PR #2703, opened 2021-02-24
Fix Angular: Any module can have routing

### LINKED ISSUE #2702: Angular module routing icon for any module, not just main

Currently main Angular routing file `app-routing.module.ts` does have a differentiated icon (`'ng_routing_ts'`).

However, in Angular any module can have a routing file and be routed to. However the existing icon for Angular routing is not currently being assigned, thus rendering the currently visible icon without any difference to the module file.

Here you can notice the Users module routing file icon is the same as the Users module file icon itself, despite the main App module routing file icon actually being differentiated:
image

So my proposal is **to enable any module routing file in an Angular project to display the icon being currently used for main App module routing**.

- Language Extensions affected: `-routing.module.ts`, `-routing.module.js`, `-routing.module.dart`

jagomf:
Any module can have routing features in Angular, not only the main `app` module, so icon should be applied to any module's routing file.

_**Fixes #2702**_

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


robertohuertasm (commented):
Unfortunately, this doesn't work like that. We still haven't got support for wildcards in filenames or extensions and this is due to a well-known vscode api limitation, so it's not on us.

As an alternative, remember you can always use the customization functionalities and associate the icon you want to the filename of your choice. That, though, should be a manual process on your side, probably per project. No other workaround for the moment. 😞 Sorry.

PR #2720, opened 2021-03-09
Add support for Laravel Vapor

d8vjork:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


src/iconsManifest/supportedExtensions.ts
```diff
@@ -3961,6 +3961,7 @@ export const extensions: IFileCollection = {
     { icon: 'vala', extensions: ['vala'], format: FileFormat.svg },
     { icon: 'vapi', extensions: ['vapi'], format: FileFormat.svg },
     { icon: 'vash', extensions: ['vash'], light: true, format: FileFormat.svg },
+    { icon: 'vapor', extensions: ['vapor.yml'], format: FileFormat.svg },
```
robertohuertasm, line 3964: Is this intended to be an extension like `a.vapor.yaml`? Or will it be the name of a file? If it's the later, then you should add `filename: true` and that will make it (similar to [this](https://github.com/vscode-icons/vscode-icons/blob/cf315a35e926fb9f225e15406de8316f41ad7c7e/src/iconsManifest/supportedExtensions.ts#L2662-L2672))

robertohuertasm (changes requested):
Thanks for this. Just a minor doubt before merging.

src/models/language/languageCollection.ts
```diff
@@ -225,6 +225,7 @@ export interface ILanguageCollection extends INativeLanguageCollection {
   tuc: ILanguage;
   twig: ILanguage;
   typo3: ILanguage;
+  vapor: ILanguage;
```
robertohuertasm, line 228: Note that language is only needed if you want to add language support. This basically means that if there's an extension in the marketplace supporting vapor, we would use it's configuration in order to decide which extensions to use.

src/iconsManifest/languages.ts
```diff
@@ -286,6 +286,7 @@ export const languages: ILanguageCollection = {
   typescript: { ids: 'typescript', defaultExtension: 'ts' },
   typescriptreact: { ids: 'typescriptreact', defaultExtension: 'tsx' },
   typo3: { ids: 'typoscript', defaultExtension: 'typoscript' },
+  vapor: { ids: 'vapor', defaultExtension: 'vapor.yml' },
```
robertohuertasm, line 289: See my comment about the necessity of using a `language` here. I've searched the marketplace and found out about one extension. The id, though, it's not `vapor` but `vaporlang`. You can check it out [here](https://github.com/minkcv/vaporlang-vscode/blob/master/package.json#L16).

robertohuertasm (changes requested):

src/models/language/languageCollection.ts
d8vjork: Oh ok, then I'll remove it on the next PR, didn't know why was this required, just copy one of the PRs.

src/iconsManifest/languages.ts
d8vjork: No, it's Laravel's Vapor (serverless platform & framework), sorry I didn't describe it on the PR clearly. So it doesn't have an extension (yet), but it'll be cool if it has its own icon as the file is always called vapor.yml.

src/iconsManifest/supportedExtensions.ts
d8vjork: Ok, thanks for the suggestions, changing asap

PR #2722, opened 2021-03-11
Add alias for interfaces folder

george-hm:
This PR adds 'interface' as an alias for the 'interfaces' folder icon

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):
Fair enough! Thanks for this addition!

PR #2723, opened 2021-03-13
Fix license link in readme

tywmick:
Link is currently broken because the casing doesn't match the filename.

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


robertohuertasm (approved):
Thanks for spotting this!

PR #2727, opened 2021-03-15
Fix URL typo in changelog entry

kymckay:
**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

Just an incorrect URL I noticed while checking out the latest changes


robertohuertasm (approved):
Thanks!

PR #2678, opened 2021-01-02
fix(cfm): add more language ids

robertohuertasm:
It seems CFM extensions are using a new language id. I've kept the old one in case some extension is
still using it although I have reviewed all the cfm extensions and they don't seem to be supporting
anymore.

fixes #2673

PR #2728, opened 2021-03-15
fix(cfm): remove wrong language id association

robertohuertasm:
See #2678

//cc @homestar9, wait for the next release and the issue should be fixed.


KingDarBoja:
Why won't we just remove the `cfc` lang id and keep only the `cfml` one as this also covers `.cfc` extension at https://github.com/KamasamaK/vscode-cfml/blob/master/package.json#L39 ?

Makes sense as homestar9 pointed out: 
> .cfc files and .cfm files are both CFML language files

robertohuertasm:
I wasn't sure if they where indeed the same language, and that's why I left it like it was. But it probably makes sense to remove it.

robertohuertasm:
My main concern is if there's some extension using the `cfmhtml` id. I'll keep it for the moment.

PR #2731, opened 2021-03-18
Add support for `tsv` extension

mMosiur:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Adds support for `.tsv` files to be treated as `.csv`. I know that often `.csv` files are used with tab as a separator, but I myself like the verbosity of `.tsv` in those cases and use that instead. Maybe some people would find that useful as well.


KingDarBoja:
I don't see nothing wrong with this PR. 🚀

KingDarBoja (approved):

PR #2732, opened 2021-03-18
Add support for file 'tailwind.config.json' in the 'tailwind' icon

mkvlrn:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Adds support for `json` extension in the `tailwind` icon extensionsGlob, allowing `tailwind.json` or `tailwind.config.json` files to use the icon.


KingDarBoja:
Does Tailwind recognize json files as config? I can't find at the docs about supported extensions.

mkvlrn:
> Does Tailwind recognize json files as config? I can't find at the docs about supported extensions.

Yes, you can point to them the same way you can point to a non default webpack configuration, for example.

This is a sample .postcssrc.json I've been using that works just fine.

```json

// .postcssrc.json
{
  "plugins": [
    [
      "postcss-preset-env",
      {
        "browsers": "last 2 versions"
      }
    ],
    [
      "@tailwindcss/jit",
      {
        "config": "./tailwind.config.json"
      }
    ],
    "cssnano"
  ]
}
```

I admit this comes from wanting to use JSON instead of JS for most/all configuration files in a project. Figured I'd use the same format I've been using with a Jest config (jest.config.json, which is picked up by default by jest).

KingDarBoja (approved):
I have nothing against supporting `json` extensions for tailwindcss as its a matter of user choice.

robertohuertasm (approved):

PR #2733, opened 2021-03-20
Add support for Yarn Berry config file

### LINKED ISSUE #2714: [Icon Request] Yarn Berry configuration file
  - Type: `file`
  - Icon Name: `file_type_yarn`
  - Sample original Icon: logo
  - Filenames: `.yarnrc.yml`

- More info: Yarn Berry (v2) has a configuration folder that name is `.yarnrc.yml`. This file is identical `.yarnrc` file in Yarn Classic and icons should be same

KingDarBoja:
_**Fixes #2714**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


src/iconsManifest/supportedExtensions.ts
```diff
@@ -4335,6 +4335,8 @@ export const extensions: IFileCollection = {
         '.yarn-metadata.json',
         '.yarnignore',
       ],
+      filenamesGlob: ['.yarnrc'],
```
robertohuertasm, line 4338: Not sure about this change. According to #2714 we would need support for `.yarnrc.yml` as a folder. Is that right? And it seems also that we should add `.yarnrc.yml` as filename. If that's the case, why using globs and not just add ` .yarnrc.yml` to the collection? It would be clearer.

robertohuertasm: @KingDarBoja any comments about this?

KingDarBoja: I thought I answered this before... guess I never submitted anything lol

Anyway, the `.yarnrc.yml` should be used for yarn v2 and onwards, so there is no need to use the filenamesGlob as you said.

robertohuertasm (approved):

PR #2738, opened 2021-03-26
Add support for Elyra Pipelines

bourdakos1:
Add support for Elyra Pipeline files: https://github.com/elyra-ai/elyra

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm:
Hi @bourdakos1, is this icon original? I cannot see it in the Elyra documentation. It also seems like a very generic extension, which could potentially be tied to other kind of pipelines. That's why I'm asking about the origin of the icon. Thanks for the PR.

bourdakos1:
Hi @robertohuertasm, this icon was originally created for [@elyra/canvas](https://github.com/elyra-ai/canvas) which is the core of Elyra Pipelines, but also used in a couple other IBM pipeline related products. The `@elyra/canvas` package can be used by anyone to build a pipeline editor, so the icon was designed to be generic. Hope this helps :)

robertohuertasm:
Thanks @bourdakos1 for the explanation. It helps a lot!

robertohuertasm (approved):

PR #2789, opened 2021-06-12
More easy description for vscode beginner

You-saku:
1. When I install vscode-icons, I used https://github.com/vscode-icons/vscode-icons#installation as reference. 
2. But, I don't understand where ```ext install vscode-icons``` should be execute in. Actually I tried the command in the terminal(zsh) at first.
3.  I think this suggestion is better for vscode beginner. How about this change ?

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


robertohuertasm (approved):
Thanks for this ❤️

PR #2796, opened 2021-06-13
fix(vsce): fix the vsce version

robertohuertasm:
New 1.93.0 version makes the publication fail with an error saying that the The specified icon 'extension/images/logo.png' wasn't found in the extension.

I looked into vsce issues and there was one related which was closed without any answer in 2018. I'll get back to it at some point, but for the moment I will merge this to unblock our release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ jobs:
       os: linux
       node_js: 12.4.0
       before_script: skip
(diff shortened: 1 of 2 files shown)


robertohuertasm:
For further reference: https://github.com/microsoft/vscode-vsce/issues/341

gerson-henrique:
i'm lost in the same trouble in POP_OS

PR #2819, opened 2021-08-06
feat(unified): add missing file extensions

remcohaszing:
rehype, remark, and retext are all part of the unified ecosystem. They use `unified-engine` to resolve their configuration files. This now supports `.cjs` and `.mjs` files.

See https://github.com/unifiedjs/unified-engine/blob/main/lib/configuration.js#L50-L57 for a list of supported file extensions.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


remcohaszing:
The code climate failure is due to an error in code climate, not because of the changes.

robertohuertasm (approved):

PR #2827, opened 2021-08-24
feat(preact): add Preact icon

Snazzah:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


src/iconsManifest/supportedExtensions.ts
```diff
@@ -3181,6 +3181,14 @@ export const extensions: IFileCollection = {
       format: FileFormat.svg,
       disabled: true,
     },
+    {
+      icon: 'preact',
+      extensions: [],
+      filenamesGlob: ['preact.config'],
```
robertohuertasm, line 3187: This is not strictly needed as there's only one possible extension `.js` but it doesn't hurt. You can see what would be the leaner option just below in the `precommit` icon.

robertohuertasm (approved):
Aside from the minor comment I think this is good to go! Thanks for this contribution!

PR #2829, opened 2021-08-27
Add .devcontainer.json to devcontainer file extensions

willfrey:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Apparently VS Code supports both [`devcontainer/devcontainer.json` and `.devcontainer.json` in the root of the project](https://code.visualstudio.com/docs/remote/create-dev-container#_create-a-devcontainerjson-file), so I added `.devcontainer.json` to the `extensions` field for the `devcontainer` icon.

Assuming that this change is okay, I'm unsure if I should update the Wiki or if that's something you do (or if it's automated). Just let me know!

### DIFF
--- a/src/iconsManifest/supportedExtensions.ts
+++ b/src/iconsManifest/supportedExtensions.ts
@@ -1202,7 +1202,7 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'devcontainer',
-      extensions: ['devcontainer.json'],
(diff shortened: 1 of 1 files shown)


robertohuertasm (approved):
LGTM! Thanks for this! Don't worry about the wiki. That should be recreated automatically.

PR #2842, opened 2021-10-05
feat: add codeowners icon

gperdomor:
_**Fixes #IssueNumber**_

**Changes proposed:**

- [x] Add CODEOWNERS support
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm:
@gperdomor have you tested this icon in dark mode? My guess is that it won't work very well but I've not tested it either.

robertohuertasm:
I've pushed a fix to the declaration. The icon doesn't work in dark modes so you can provide a light version or think of a color that works in both kind of themes. Aside from that, I'm not sure about the icon. Where did you get it? What's the rationale behind the shield and a key hole? I would have expected some group of people and some reference to the code (html tags or something). I'm curious to know more about your thoughts.

gperdomor:
@robertohuertasm you are right about the color themes, it's solved now

Screen Shot 2021-10-31 at 2 41 47 PM
Screen Shot 2021-10-31 at 2 41 59 PM

about the icon, is the icon used by GitHub and it was taken from octicons library: https://primer.style/octicons/shield-lock-24

Screen Shot 2021-10-31 at 2 45 54 PM

robertohuertasm:
@gperdomor thanks for the explanation. It seems though that the light icon was not uploaded or at least I can't see it. I've tested your branch but it doesn't work yet.

gperdomor:
@robertohuertasm my bad... try again please

robertohuertasm (approved):
Thanks for the update! Ready to be merged!

PR #2852, opened 2021-10-31
feat(vanilla-extract): add vanilla extract icon

douglaseggleton:
_**Fixes #IssueNumber**_

**Changes proposed:**

- [x] Add Vanilla Extract Icon
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Adds an icon for https://vanilla-extract.style/ - The convention of `.css.ts` seems unique to this library.

vanilla-extract-icon-example


robertohuertasm (approved):
@douglaseggleton I've optimised the icon a little bit and removed some of the layers. Now its size is 5kb, which is still high according to our standards (we aim for 2kb icons) but I think we can work with it.

Thanks for your contribution!

PR #2857, opened 2021-11-03
Use git icon for 'ignore' language

### LINKED ISSUE #2802: ignore language should use git icon, not docker icon

1. vscode-icons incorrectly uses the docker icon for 'ignore' language.  'ignore' language represents git files and should use the git icon
    - Files named `.gitignore` are the exception, as they do correctly use the git icon
2. vscode-icons should use the docker icon for files named `.dockerignore` (instead of default file icon)

**Default VS Code Theme (Seti)**
image
Change Language Mode command:
image

**VS Code Icons**
image
Change Language Mode command:
image

baincd:
_**Fixes #2802**_

**Changes proposed:**

- [x] Fix

    - 'ignore' language uses git icon
    - `.dockerignore` file uses docker icon (whether language is 'plaintext' or 'ignore')
      - Defined as a file extension for  docker in `supportedExtensions.ts` - following the same pattern as other `.*ignore` files

image
Change Language Mode command:
image


robertohuertasm (changes requested):
hi @baincd, not sure what this PR is addressing. As far as I can see, the language id `ignore` is actually being used by the [most dowloaded docker extension in the marketplace](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker). 

I mean, you're assuming that the language id ignore belongs to git, but that's an assumption at the moment. Indeed, the [most downloaded extension with support to gitignore](https://marketplace.visualstudio.com/items?itemName=codezombiech.gitignore) declares the `gitignore` language id.

So, I would just add a new language id called `gitignore` and associate it to the git icon and leave the `ignore` language id associated to docker for the moment until we have more insights of who is the most used extension using that language id.

baincd:
Hi @robertohuertasm, 

I do think the `ignore` language id belonging to git is a better solution than it belonging to Docker (or any other technology), because:

1. Not everyone is using Docker.  Per the most recent Stack Overflow Developer Survey, [less than half of all developers are using Docker](https://insights.stackoverflow.com/survey/2021#section-most-popular-technologies-other-tools).  So a majority of developers using VS Code likely do not have any Docker extension installed or open any files related to Docker.
2. For better or worse, the default behavior of VS Code is that the `ignore` language belongs to git.  Maintaining that assumption leads provides better consistency and user experience for users
3. Consistency with other "ignore" files.  There are 31 other "ignore" files that are supported by vscode-icons by their filename.  I think `.dockerignore` files should be treated similar as these other "ignore" files

I will admit that I think the `ignore` language id belonging to git isn't a perfect solution, however it is the best option available.

robertohuertasm:
@baincd, the rationale for us is just to respect the [Known Language Identifiers](https://code.visualstudio.com/docs/languages/identifiers#_known-language-identifiers).If the language id is not there, we assume that it's a custom one and we support the most downloaded extension. 

As you can see, no `ignore` language id is declared there. And the most downloaded extension using that language id is the docker extension.

That being said, I've tested this by disabling all my extension an it seems, indeed, that the `ignore` language comes by default and it's associated to the git icon. For what I've been able to see it's the default vscode file icon theme behavior. That is not something enforced by the vscode icon theme api but just a decision from the default theme creators. 

Anyway, after looking into all this, I think you have a point. Generally speaking, I think the usage of this id it's indeed far from perfect no matter the decision we make.

So... let's do the change you were proposing. We'll deploy it in our next release.

robertohuertasm (approved):

PR #2867, opened 2021-11-28
Add support for goctl

codingfanlt:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Add support for goctl,  goctl is go-zero's API definition and generation tool, which can generate Go, iOS, Android, Kotlin, Dart, TypeScript, JavaScript code with one click based on the defined api file. For more information, please refer [here](https://github.com/zeromicro/go-zero).


robertohuertasm (changes requested):
Unfortunately, we cannot accept this icon. We don't support embedded images in svg. Maybe we could work using the go icon as the base and add something specific for this kind of extension?

robertohuertasm:
image
would that work?

robertohuertasm:
@codingfanlt as I'm preparing a new release, I'll go with that version, so you can have at least a temporary icon. We can work on this in the future.

codingfanlt:
Thank you very much for the temporary icon. I will provide the composite request icon in the future, thanks again.

PR #2882, opened 2021-12-11
feat(icon): support for .tool-versions

azat-io:
Add `.tool-versions` filetype support for [asdf](https://asdf-vm.com/) version manager 

**Changes proposed:**

- [x] Add

### DIFF
--- a/src/iconsManifest/supportedExtensions.ts
+++ b/src/iconsManifest/supportedExtensions.ts
@@ -4023,6 +4023,12 @@ export const extensions: IFileCollection = {
       format: FileFormat.svg,
     },
     {
+      icon: 'config',
+      extensions: ['.tool-versions'],
+      filename: true,
+      format: FileFormat.svg,
+    },
+    {
       icon: 'tox',
       extensions: ['tox.ini'],
       filename: true,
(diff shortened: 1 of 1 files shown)


robertohuertasm (changes requested):
Add this alphabetically, please 🙏🏻 
Thanks for the contribution!

azat-io:
@robertohuertasm Done!

src/iconsManifest/supportedExtensions.ts
```diff
@@ -1036,7 +1036,7 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'config',
-      extensions: ['plist'],
+      extensions: ['plist', '.tool-versions'],
```
robertohuertasm, line 1039: Is `.tool-versions` a extension or a file name? In case it's a extension, you should remove the `.`. If it's a filename, you should create another entry: 
```json
{

   icon: 'config',
   extensions: '.tool-versions',
   light: true,
   filename: true,
   format: FileFormat.svg
}
```

robertohuertasm (changes requested):

src/iconsManifest/supportedExtensions.ts
robertohuertasm: Note that the test are failing because you put a dot into an extension 😉

azat-io: Done

robertohuertasm (approved):
LGTM 🚀 Thanks for the changes!

PR #2887, opened 2021-12-21
Add `cuda-cpp` to `ids` item for cuda language.

SaltyChiang:
**Changes proposed:**

- [x] Add

For those who don't use [`vscode-cudacpp`](https://marketplace.visualstudio.com/items?itemName=kriegalex.vscode-cudacpp) extension like me, `vscode-icons` will not change `*.cu` and `*.cuh` files' icon. The icon for "CUDA C++" item in `Select Language Mode` dialog box is also blank.

But there exsists a default language mode for CUDA called `cuda-cpp`, adding this to `ids` for `cuda` helps a lot.


robertohuertasm (approved):

PR #2891, opened 2022-01-01
feat(icon): support for store and stores folders

azat-io:


robertohuertasm (changes requested):
I guess I cannot accept this PR as `redux` folder icon is really tied to `redux`. Store folder could be potentially related to any kind of tech. Let's find another icon for this.

N0tExisting:
Think it would make sense to use the DB folder by default and then use project detection.

robertohuertasm:
Closing this as this solution is not acceptable. Thanks for your contribution, though.

PR #2919, opened 2022-02-11
fix(constants): make link in notification point to current release notes

atiabbz:
- [x] Fix

Changed the link in `constants` so that users can go directly to the [Releases](https://github.com/vscode-icons/vscode-icons/releases) page with the current release notes since the use of the previous one was stopped in December 2021.


robertohuertasm (approved):

PR #2981, opened 2022-04-19
refactor(gnu): rename gnu logo and add .ld files

### LINKED ISSUE #2676: [Icon Request] Support for GNU Linker (ld) file support
  - Type: `extension`
  - Icon Name: `gnu_linker`
  - Sample original Icon: [Icon Example](https://cdn1.iconfinder.com/data/icons/management-business-line/25/pazel-512.png)
  - Extensions: **.ld** (Capital - **.LD**)

- More info: [GNU Linker](https://en.wikipedia.org/wiki/Linker_(computing)#GNU_linker). I chose that icon because I thought that three pieces of a puzzle that need to be put together (like linker does, takes object files and connect them together into a final executable file) is similar to the linker process. The Icon can be converted to _svg_ format.

- An LD file is a script written in the GNU "linker command language." It contains one or more commands that are used to configure how input files storing static object code are to be compiled into a single executable program or library for the GNU operating system.

[Makefile uses the GNU logo](https://github.com/vscode-icons/vscode-icons/blob/master/icons/file_type_makefile.svg) so maybe also use the GNU logo for the linker or create different for each one of them.

sasial-dev:
_**Fixes #2676**_

**Changes proposed:**

- Rename `makefile` logo -> `gnu`
- Add `.ld` (linker) files


src/iconsManifest/supportedExtensions.ts
```diff
@@ -2622,12 +2628,6 @@ export const extensions: IFileCollection = {
       format: FileFormat.svg,
     },
     { icon: 'lync', extensions: ['crec', 'ocrec'], format: FileFormat.svg },
-    {
-      icon: 'makefile',
-      extensions: ['makefile'],
```
robertohuertasm, line 2627: Why did you remove the `makefile` extension?

sasial-dev: As in the PR commment:
> Rename `makefile` logo -> `gnu` & add `.ld` extension to it

robertohuertasm: I still don't understand why the `makefile` extension has been removed. I think we should keep supporting that and don't introduce a breaking change.

sasial-dev: True! It's because the icon for `makefile` is actually the `gnu` logo, so there'd be a double-up.

robertohuertasm: Yes, but I'm talking about the `makefile` extension. Line 2638.  My guess is that it should be added to ln 1903.

robertohuertasm (approved):

PR #2994, opened 2022-04-29
ci(github): create test and WPIL workflows with github actions

Snazzah:
This should help with all the cancelled Travis CI jobs. Adds workflows for tests and the [wpilgenerator](https://github.com/vscode-icons/wpilgenerator) using GitHub Actions.

Note that ~~I haven't put in the codeclimate test reporter within the test workflow and~~ I haven't tested WPIL working, but should work fine. I haven't touched the `travis.yml` file since you may want to keep or change it.


JimiC:
@Snazzah You may wonna have a look how I'm doing it for `nsri` (https://github.com/JimiC/nsri/blob/main/.github/workflows/build.yml). 

The concept is pretty much the same

JimiC:
Hint: Basically we are testing against the node versions the `vscode` editor is using in each release. So you will have to add all those versions in the matrix, starting from the minimum up to the latest and add the `latest` `node` version (allowing it to fail of course) just to see if anything brakes in the future.

Personally I would advise you to reach out to @robertohuertasm and update the minimum supported `vscode` version and the `npm` dependencies for that version and then use only those specific node versions (for each afterward release) in the build matrix.

Though I do encourage you keep up building the `github` actions. 👍

Snazzah:
Thanks @JimiC, also would need the `CC_TEST_REPORTER_ID` secret to be set in order for codeclimate to work. Other than that, mainly just waiting for @robertohuertasm to give input on how he wants the workflow to work.

.github/workflows/wpil.yaml
```diff
@@ -0,0 +1,33 @@
+name: Update Wiki Page Icons List
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  update:
```
robertohuertasm, line 10: Probably we should consider allowing this job to fail: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

.github/workflows/test.yaml
```diff
@@ -0,0 +1,39 @@
...
+          node-version: ${{ matrix.node }}
+      - name: Installing Dependencies
+        run: npm ci
+      - name: Test
+        run: npm test
+      - name: Send coverage report
+        if: (success() && matrix.os == 'ubuntu-latest' && matrix.node == '12.4.0' && github.repository_owner == 'vscode-icons' && github.actor != 'dependabot[bot]')
+        uses: paambaati/codeclimate-action@v3.0.0
+        env:
+          CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
```
robertohuertasm, line 39: Added this one but it seems to be failing for some reason.

robertohuertasm:
Hi @Snazzah, thanks a lot for this PR and apologies for the late response. This first half of the year has been really busy and intense for me but things seems to be getting into place again so I expect to have more time to keep maintaining this repo in the coming months.

Anyway, I added the ID for code climate but the pipeline fails for some reason.

Aside from that, I'm not sure if it wouldn't be better to split this PR into two separate PRs, one for the tests and another one for the WPIL workflow. By doing this we could merge the bits that already work. 

For the tests, let's probably work on CC support in another PR. The test workflow is already adding value even without CC, so we can work on that in another specific PR.

What do you think?

Snazzah:
Hey @robertohuertasm, not sure why CC isn't working, but I've removed it for this PR so we can move that to another one. If you have a log of this failing, I can take a look.

For now, seems like this is good to merge, and you can probably remove Travis from the CI.

robertohuertasm (approved):
LGTM 🚀 Thanks @Snazzah for this PR ❤️

robertohuertasm:
@Snazzah you can see the log here: https://github.com/vscode-icons/vscode-icons/runs/7266001105?check_suite_focus=true

image

It seems that it's not getting the variable from the secrets... let me check into that...

robertohuertasm:
@Snazzah I'm going to merge this one as it will be very useful and remove some of the travis functionality as per your suggestion. Thanks again for your contribution!

PR #2998, opened 2022-05-03
feat(icon): add eleventy icon

### LINKED ISSUE #2995: [Icon Request] Eleventy (11ty)
 
**Type**: `filename`
 **Icon Name**: `11ty`
 **Filenames**: `.eleventy.js`
 
| Sample Original Icon | Other Known Variations |
|----------------------|:-------------:|
|                                      |                                       |

> I was unable to source a `.svg` for the black background logo variation.

 **More info**:
 Eleventy (also known as 11ty) is a javascript static site generator tool similar to Jekyll. The logo is not very elegant (imo) but would be nice to have it applied to `.eleventy.js` configuration files which exist in the root of projects. Given the logo base colour is black, might be worth inverting it to a white if you decide on using the _Original Icon_  reference.
  
- [Official Website](https://www.11ty.dev)
- [Documentation](https://www.11ty.dev/docs/)
- [NPM Registry](https://www.npmjs.com/package/@11ty/eleventy)

sasial-dev:
_**Fixes #2995**_

**Changes proposed:**

Adds Eleventy Icon


sasial-dev:
@panoply, I was on the fence about this one - let me know your thoughts on if I should re-design the icon a bit. Maybe just using the 11's and adding a background?

Code-pQFNpsni
Code-2Z3ZrVvl

panoply:
I feel you on this. 11's would be much more elegant, the `ty` is unsettling.

sasial-dev:
Much Better!

Code-1mdjcMX5
Code-3eKI6FBF
Code-sPX5a08k
Code-xn9SutTZ

panoply:
Much better indeed. The quality standard maintained by this project does not go unnoticed. Much appreciate the time and efforts.

robertohuertasm (changes requested):
Thanks for this PR but I have to say that I have some doubts about the icon. Normally, we always use the original icon unless there's a major issue with it. We don't care about if it's nice or not as we don't get to judge that. In my opinion, we should go with the original ones. 

What we do, sometimes, is to provide another set of icons (i.e. see `access` and `access2` [here](https://github.com/vscode-icons/vscode-icons/blob/master/src/iconsManifest/supportedExtensions.ts#L10-L53)) so users can use the icon customization setting and use the other alternatives. 

Note that the extension also supports custom icons.

sasial-dev:
I see what you mean, but the problem is that the original icon was squashed so much that it was essentially distorted (see the picture in the first comment). I see what you mean, but this needs a bit of thought.

panoply:
A solution here might be to black box it with radius?

Screenshot 2022-05-22 at 22 24 30

robertohuertasm (changes requested):
The icon margins are not respected.

robertohuertasm (commented):
@sasial-dev for some reason I cannot checkout this PR and do the fix myself. We'll have to wait for you to fix the margins to publish it :(

robertohuertasm:
@sasial-dev this PR is still pending. As I cannot check it out I guess I'll close it and fix it myself.

src/iconsManifest/supportedExtensions.ts
```diff
@@ -1522,6 +1522,19 @@ export const extensions: IFileCollection = {
       format: FileFormat.svg,
     },
     { icon: 'elasticbeanstalk', extensions: [], format: FileFormat.svg },
+    {
+      icon: 'eleventy',
+      extensions: ['eleventy.js'],
```
gradyat, line 1527: Eleventy 2.0 also supports `eleventy.config.js`, and `eleventy.config.cjs` (in addition to `.eleventy.js`). Can those extensions be added?

11ty/eleventy#2464

gradyat (changes requested):

src/iconsManifest/supportedExtensions.ts
robertohuertasm: Sure! Thanks for the suggestion!

robertohuertasm: Will tackle that in another PR as it will probably need a new icon for those files. #3222

robertohuertasm (approved):

PR #3001, opened 2022-05-03
refactor(webp): remove icon

sasial-dev:
**Changes proposed:**

Removes the `webp` icon to bring it inline with the `webv` icon (uses the video icon).
`webp` shouldn't have its own icon, and it looks out of place in folders with webp, jpeg & png.


robertohuertasm (approved):

PR #3017, opened 2022-05-15
refactor(eslint): add tsconfig.eslint.json

sasial-dev:
**Changes proposed:**

tsconfig.eslint.json is commonly used in TypeScript eslint. 
It is also mentioned [in the docs](https://typescript-eslint.io/docs/linting/type-linting#i-get-errors-telling-me-the-file-must-be-included-in-at-least-one-of-the-projects-provided)!

### DIFF
--- a/src/iconsManifest/supportedExtensions.ts
+++ b/src/iconsManifest/supportedExtensions.ts
@@ -1497,7 +1497,7 @@ export const extensions: IFileCollection = {
     {
       icon: 'eslint',
       extensions: ['.eslintrc', '.eslintignore', '.eslintcache'],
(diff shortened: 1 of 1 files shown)


sasial-dev:
True, but wouldn't it make more sense to use the eslint icon? It is after all only for eslint...

sasial-dev:
I can change it if you'd rather

JimiC:
> True, but wouldn't it make more sense to use the eslint icon? It is after all only for eslint...

No

robertohuertasm (approved):
LGTM 🚀

PR #3004, opened 2022-05-05
feat(icon): add .objidconfig icon

vjekob:
_**Fixes #3003 **_

**Changes proposed:**

- [x] Add

Adds AL Object ID Ninja configuration file (.objidconfig)

PR #3025, opened 2022-05-24
refactor(icon): fix SVG size and curve accuracy issues for objidconfig

vjekob:
**Changes proposed:**

- [x] Fix

Previous icons were created through online conversion from PNG to SVG. This resulted in large files that were optimized in size by @robertohuertasm.

However, the optimization took care only of the file size, and since the original icons were conversion from raster to vector, there were accuracy issues with curves that did not have correct shapes (especially visible around the knot in the upper-right position in the icon), and had gaps between black and teal areas (in light version of the icon, especially visible when zoomed in).

These icons are true vectors, created from scratch (rather than through conversion) and they curve accuracy issues. Also, being actual vectors, they are also smaller in size.

Finally, the new icons are also better centered (the previous ones were shifted too far to the right)


vjekob:
Hi @robertohuertasm,

following up on our conversation in PR https://github.com/vscode-icons/vscode-icons/pull/3004 from May 23. I've optimized the icon size, and I have created a new PR (this one). However it has been a long time that it hasn't been approved.

Can you please let me know if there is anything else I need to do to get these new, higher-quality and smaller-size icons included?

Thanks!

robertohuertasm:
Sorry @vjekob I've been on holidays these days and haven't had time to review pending PRs. I'll take a look during the week but at first sight, I've noticed that they seem a little bit smaller. Is that right? Aside from that, they look good to me!

vjekob:
It's not just that they are smaller. Here's what I have done:
* New are manually drawn as curves in Illustrator. Previous were converted from PNG using an online tool.
* New are just a tiny little bit smaller, better centered, to look better in AL Language workspaces (where they are used).
* New are much, much smaller in file size.

robertohuertasm:
@vjekob I meant smaller in actual size, not in file size. 😅

vjekob:
Yes I know. The second bullet point addresses that. They are tiny bit smaller as compared to previous ones, because this way they look better in AL Language workspaces, where they are meant to be used.

robertohuertasm (approved):

Issue #3041, opened 2022-06-02
[Icon Request]: Support for: Common/Shared Folder


sean-7777:
### Housekeeping

- [X] I'm sure this issue is _not_ a duplicate

### Icon Type

- [ ] File Icon
- [X] Folder Icon

### Icon Type

- [ ] Extension
- [X] Filename
- [ ] Language ID

### Original Icon

Not sure what I need to put here? 😕
Maybe something like this:
![image](https://user-images.githubusercontent.com/67909939/171702570-cef3a46c-23f6-467e-9978-a0b8e082b934.png)

### Icon Information

A shared or common folder is used in many projects for contents such as utilities.

PR #3043, opened 2022-06-02
feat(icon): add shared folder icon

PAXANDDOS:
_**References #3041**_

**Changes proposed:**

- [X] Add

Added icon for 'share' and 'shared' folders. Maybe someone won't agree with that type of icon, but I tried to make it unique but still intuitive. See the issue for more information.


robertohuertasm (approved):
Nice icons!

PR #2977, opened 2022-04-18
feat(icon): add cspell icon

sasial-dev:
_**Fixes #2963**_

**Changes proposed:**

Add's the cspell icon.

@carlocorradini - I took the (surpisingally painstaking) task of redrawing the icon by hand - please let me know if anything's off!

PR #3050, opened 2022-06-04
feat(icon): add cspell icon

### LINKED ISSUE #2963: CSpell support
Add support for [CSpell](https://cspell.org/configuration/) configuration files.

  - Type: `/ico`
  - Icon Name: `cspell`
  - Sample original Icon: 
    logo
  - Extensions: `.json`, `.js`, `.cjs`, `yaml`, `.yml`
  - Filenames: `.cspell`, `cspell`, `.cSpell`, `cSpell`, `cspell.config`

- More info: Only `.ico` file available, but I think creating an `svg` equivalent should not be too difficult.

PAXANDDOS:
_**Fixes #2963**_

_**Superior version of #2977**_ (PR meant to be closed if this one succeeds)

**Changes proposed:**

- [X] Add

**Regarding changes compared to the latest PR:**

@sasial-dev you did a good job redrawing the icon! 😃 
But still, I could find some things that could be fixed, so what I've changed:
- Decreased size from `512x512` to `32x32`, I know that's not a big number, but still too big for a simple icon like this one, which has no small details. And it's really better to stick to `32x32` since vector graphics do not require a lot of space and this produces smaller size SVGs, which is great for big projects with a lot of assets like this one
- You used *width/height*, I used **viewbox** which is more reliable and makes scaling possible
- Added rounded corners on vectors. I guess you had a clip-path for the entire figure because when I loaded it into my editor and deleted everything except vectors, there were no rounded corners
- And finally, you used `#ffffff` and I used `#d2d2d2`. Guidelines...


robertohuertasm (approved):

PR #3060, opened 2022-06-14
feat(icon): add esphome icon

glmnet:
_**Fixes #IssueNumber**_

**Changes proposed:**

- [x] Add


robertohuertasm:
@glmnet I'm not sure if this icon will work ok on dark themes as the borders won't be really visible. I think we should probably implement a dark/light version of this icon. WDYT?

glmnet:
Yes! Thanks for bringing up the suggestion, I wasn't aware this was possible.

robertohuertasm (approved):

PR #3062, opened 2022-06-18
fix(progress): prolog extension collided with progress openedge

clement-brodu:
**Changes proposed:**
add missing extensions for progress openedge

- [x] Fix


robertohuertasm:
@clement-brodu it seems that that extensions are already supported via language id. The icons should be there if you use [this other extension](https://marketplace.visualstudio.com/items?itemName=chriscamicas.openedge-abl). See how it is defined [here](https://github.com/chriscamicas/vscode-abl/blob/master/package.json#L51-L54)

Is it not working for you?

robertohuertasm:
I'm closing this unless @clement-brodu confirms this is not working with language id.

clement-brodu:
Hello, 

Even with the extension detecting the correct file type, the icon is not the correct one for `*.p` files.

image

robertohuertasm:
That’s strange. Looks like a bug. I’ll try to reproduce it.

robertohuertasm:
I see what's going on. Prolog has `P` as an extension and both are competing against that. Let me check if we can get rid of that prolog extension.

robertohuertasm:
Let's get rid of the `P` extension in prolog and leave it to language extensions implementors to decide.

robertohuertasm (approved):

PR #3069, opened 2022-06-25
feat(icon): Add additional Godot icons

officialduke99:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Adds additional Godot icons, like the `.tscn` scene files, `.tres` resource files and `.godot` project settings file. It also renames the already existing `Godot` icon to `GDScript`, as that is what it is used for. The icons were created by adjusting the hue of the original, the colors were chosen based on the Seti icon set from VSCode.


robertohuertasm (changes requested):
Hi @officialduke99! First of all, thank you very much for taking the time to create this PR! :heart: Unfortunately, I cannot accept it like that. I would need more information. 

I've seen you've created several language ids but my guess is that you haven't based that decision on a vscode extension supporting those language ids, am I right?

[These](https://github.com/godotengine/godot-vscode-plugin/blob/master/package.json#L171) are the `language ids` supported by the most downloaded extension.

The problem here would be that we could not support the same color-extension association you proposed, because both the `tres` and `tscn` extensions, along with some more, are supported with the `gdresource` language id.

Here, we can do several things, but it's vital to understand if the extensions under the `gdresource` language id should be all under the same icon or not. I'm not a Godot user myself so I don't have a clue, and basically, tend to rely on extension creators to define language ids and the dependant extensions.

If you're positive that those files should have a separated icon, then you should provide some documentation/links to back up your opinion and we should use normal icon associations instead of language id associations.

Let me know what are your thoughts and I'll be glad to help you modify this PR so we can merge it.

officialduke99:
Hi @robertohuertasm, thanks for reaching out, 
as far as the different `gdresource` file types are concerned, I guess the reason that all of the extensions are under one language ID is that I'm pretty sure they all have the exact same syntax, however these files all have wildly different use cases. 

To start with, the `.godot` extension is used (typically, I'm not sure if there are any edge cases) for just one file in an entire project, which contains the overall project settings. The `.tres` is for project specific resources and contains just one Godot "node" (basically the smallest building block, a mostly self contained object with functions and properties) with all of its settings or a specific resource object with all of its settings, which can be used in these nodes. Lastly, the `.tscn` is for entire trees built-up from Godot's nodes, becoming essentially a reusable prefab with a more complex functionality than any single node can have by itself.

(There isn't that much external documentation on all these file types, probably the closest thing is the Godot Docs' [page on the `.tscn` file format](https://docs.godotengine.org/en/stable/development/file_formats/tscn.html), but even that isn't as extensive as what I wrote here)

It's for this reason that **I believe that these file types should have different icons** and as I off-hand mentioned in the description for this PR, I'm not alone, since the default VSCode icons are color coded in much the same way. Even in the engine itself, the files are shown with different icons. (here is an example) Showcase

The other thing you've mentioned was me creating several `language ids`, and to be honest, I didn't really know what I was doing. I mostly just went off from what I saw in other PRs and wherever I found ".gd" or "Godot" in the files. That's because I didn't really find good documentation on how to add these in correctly and it may very likely be a me thing, but creating a better PR guide in the wiki for icon submissions may be something someone might try to look into. But it's also totally possible that it's just clear to everyone except for me.

And one last thing, I've debated whether to include other Godot file formats, as for example in the language ids you linked to, (excluding those file extensions which aren't exclusive to Godot, like `.res`) there are a few other files you may come across in your Godot projects, especially when you use a different programming language than GDScript through GDNative, however these files don't have special icons in the default VSCode icon set either and they are quite uncommon for most people, so I didn't bother. 

That's pretty much it, I'll be happy to fix any issues with the PR to get it up to snuff.

robertohuertasm (dismissed):
Fixed all issues

robertohuertasm (approved):

PR #3072, opened 2022-06-27
fix(webpack): Add webpack cjs and mjs extensions

MaxTechnics:
**Changes proposed:** Add webpack cjs and mjs extensions

- [x] Fix

Sometimes webpack configs are used where `type: 'module'` is in package.json. So your webpack config needs the `.cjs` extension.


robertohuertasm (approved):

PR #3075, opened 2022-07-04
feat: update vite and vitest extensions

xiaoxiangmoe:
see https://github.com/vitejs/vite/blob/d41286065d0f5b0b82c16fd97735b2cf2d744794/packages/vite/src/node/constants.ts#L33-L40


src/iconsManifest/supportedExtensions.ts
```diff
@@ -4513,15 +4513,29 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'vite',
-      extensions: ['vite.config.js', 'vite.config.ts'],
+      extensions: [
```
robertohuertasm, line 4516: It would be nice to change this similar to [this](https://github.com/vscode-icons/vscode-icons/blob/master/src/iconsManifest/supportedExtensions.ts#L528-L535)

src/iconsManifest/supportedExtensions.ts
```diff
@@ -4513,15 +4513,29 @@ export const extensions: IFileCollection = {
...
+      ],
       filename: true,
       format: FileFormat.svg,
     },
     {
       // Taken from https://vitest.dev/logo.svg
       icon: 'vitest',
       // See https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/constants.ts
-      extensions: ['vitest.config.ts', 'vitest.config.js', 'vitest.config.mjs'],
+      extensions: [
```
robertohuertasm, line 4531: Same applies here.

robertohuertasm (changes requested):
Thanks for this @xiaoxiangmoe. It looks good to me but it would be nicer if you could change it to use `filenamesGlob`.

robertohuertasm (dismissed):
I change it to be a little bit more succint.

robertohuertasm (approved):

PR #3077, opened 2022-07-08
feat(icon): add bun icons

Snazzah:
Adds icons for [Bun](https://bun.sh) including `.bun` and `.lockb` files and `bunfig.toml`.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


sasial-dev:
I wonder if `lockb` should use the `bunfig` icon?

robertohuertasm:
@sasial-dev I guess `lockb` may be similiar to `package-lock-json` according to the docs. If that's the case, then the usage of the icons would be the same as we have for `npm`. WDYT @Snazzah?

Snazzah:
@robertohuertasm Yep, the default name for Bun's lock file is `bun.lockb` and is pretty much like `package-lock.json` and `yarn.lock` but in a different format.

robertohuertasm (approved):

PR #3080, opened 2022-07-10
ci(coverage): add test coverage

robertohuertasm:
Based on previous work made by @Snazzah

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -37,3 +37,6 @@ jobs:
         uses: paambaati/codeclimate-action@v3.0.0
         env:
           CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
+        with:
+          coverageLocations: |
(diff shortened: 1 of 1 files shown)


robertohuertasm:
@Snazzah, not sure why this didn't work in #2994... I've just reused your code and now it seems to do the trick.

We were getting this warning, though:

image

Not sure if that was ok so I added the `coverageLocations` and now we're getting this:

image

I've quickly checked code climate but still cannot see any information from today. Will check in a couple of hours and let you know if something changed.

robertohuertasm:
Now I've seen that we've received a bot message stating that the coverage is 100% so I guess it works.

robertohuertasm:
Merging this for the moment. Will keep an eye on it. Thanks again @Snazzah for your help on this matter!

Snazzah:
@robertohuertasm You might need to set `coverageCommand` (which is what that warning is about) to `npm test`

robertohuertasm:
I read the actions documentation and it’s not needed. It seems to be optional.

PR #3081, opened 2022-07-10
ci(github): fix test workflow

Snazzah:
I should've probably done this before #2994 (and #3080) was merged, but didn't realize till now.

Test workflow now uses `pull_request_target`, fixing that issue about secrets not working:
> This event is similar to `pull_request`, except that it runs in the context of the base repository of the pull request, rather than in the merge commit. This means that you can more safely make your secrets available to the workflows triggered by the pull request, because only workflows defined in the commit on the base repository are run.

This would apparently have to be merged to master before it starts working, so this PR might just fail checks anyways.

Also added `coverageCommand` to the codeclimate step, which should've been there already:

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,7 +6,7 @@ on:
       - '!*'
     branches:
       - '*'
-  pull_request:
+  pull_request_target:
     tags:
       - '!*'
     branches:
@@ -38,5 +38,6 @@ jobs:
         env:
           CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
         with:
+          coverageCommand: npm test
           coverageLocations: |
             ${{github.workspace}}/coverage/lcov.info:lcov


.github/workflows/test.yaml
```diff
@@ -38,5 +38,6 @@ jobs:
         env:
           CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
         with:
+          coverageCommand: npm test
```
robertohuertasm, line 41: Not sure if this is really needed as the `npm test` command has been already executed in the previous step and it may be redundant.

.github/workflows/test.yaml
```diff
@@ -6,7 +6,7 @@ on:
       - '!*'
     branches:
       - '*'
-  pull_request:
+  pull_request_target:
```
robertohuertasm, line 9: According to this [GitHub article](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) this could be potentially unsafe and it's encouraged to go with the approach of using the `workflow-run` trigger.

robertohuertasm (changes requested):
Thanks for pointing out the issue of why the cc credentials were not working from other branches but according to what I've read I think it would be safer to separate the CC coverage step into its own pipeline triggered on `workflow-run`.

Snazzah:
Yeah, would be better to use `workflow_run`, but not sure how it would work out with a matrix along with it. I'm guessing if that entire workflow passes, then the workflow_run goes through, in that case you can specify the version in the coverage workflow and don't have to comment out the other versions.

I'll close this PR since you already put in some commits towards this.

robertohuertasm:
Thanks @Snazzah. Yes, the approach is:
- As the only node version that generates code coverage is `12.4.0`, we'll upload that information to the workflow artifacts
- Wait for the `Tests` workflow to pass
- The `workflow_run` will be triggered
- We must checkout the sha that triggered the event (the one from the PR) for the cc reporter to work properly
- We upload the information to code climate

After some trial & error I made it work. I didn't know about all these parts of github actions so thank you for leading me into this. Feel free, to take a look at the code. I'll accept any improvement suggestions!

PR #3117, opened 2022-08-17
feat: add icon for ".tree" files

PaulRBerg:
**Changes proposed:**

- [x] Add icon for [*.tree](https://github.com/structure-codes/vscode-tree-language) files


src/iconsManifest/supportedExtensions.ts
```diff
@@ -4368,6 +4368,12 @@ export const extensions: IFileCollection = {
       filename: true,
       format: FileFormat.svg,
     },
+    {
+      icon: 'tree',
+      extensions: ['.tree'],
+      filename: true,
```
robertohuertasm, line 4374: @paulrberg thanks for this. My guess is that you didn't want this to be a filename extension. If you want to support `*.tree` files, remove the `filename: true` line and the dot from the `.tree` extension.

robertohuertasm (changes requested):

src/iconsManifest/supportedExtensions.ts
paulrberg: Happy to help!

And thanks for your feedback. You're right, I wanted to support `*.tree` files. Fixed in the latest commit.

robertohuertasm: Thanks @paulrberg, could you rebase this PR?

paulrberg: Sure, just did it.

src/iconsManifest/supportedExtensions.ts
```diff
@@ -4391,11 +4391,8 @@ export const extensions: IFileCollection = {
       format: FileFormat.svg,
     },
     {
-      icon: 'truffle',
```
robertohuertasm, line 4394: It seems you removed the `truffle` icon?

paulrberg: Sorry for that! Fixed.

robertohuertasm (changes requested):
The icon name is wrong. It should be `file_type_tree.svg`. Aside from that, and considering that the icon is a black line, that won't work for dark themes. So, either we provide a light and dark alternative or you should consider the color to a color that could work for both dart and light themes.

Finally, I've tried to open the icon in inkscape and it's not shown. Not sure what kind of software you used for its creation. Illustrator?

paulrberg:
I used Figma to create the icon. I'm sorry, I'm not a graphic designer and I don't really know the best practices.

To be frank, I don't have the bandwidth to put any more effort into this. Happy to close the PR if the existing icon isn't good enough.

robertohuertasm:
@paulrberg, leave it as it is so someone can pick your work and make the changes this requires. Thanks!

paulrberg:
I thank you for merging the PR!

robertohuertasm:
Why did you close it?

paulrberg:
Shoot, I deleted my repo.

I re-created it now, and re-opened the PR.

PR #3122, opened 2022-08-18
feat(icon): Add Truffle icon

cliffoo:
Add Truffle icon for `truffle-config.[jt]s`

In the [contributing docs](https://github.com/vscode-icons/vscode-icons/wiki/Contributing) it mentions using `#d2d2d2` instead of `#fff`, but it doesn't work well with the Truffle logo. We've adjusted it to a shade that still gives good contrast on light theme. Also the white parts are relatively small so it shouldn't be an issue.

Cheers!


robertohuertasm (dismissed):

robertohuertasm (approved):

PR #3126, opened 2022-08-24
feat(icon): Add .dal icon for AL Language

### LINKED ISSUE #3125: [Icon Request]: .dal file type in AL Language
### Housekeeping

### Icon Type

- [X] File Icon
- [ ] Folder Icon

### Icon Type

- [X] Extension
- [ ] Filename
- [X] Language ID

### Original Icon

file_type_al_dal

### Icon Information

This icon represents AL Language definition files (`.dal`). With current version of vscode-icons, this icon shows exactly as `.al` files, which may look confusing when developing. Definition (`.dal`) files are third-party files reconstructed on the fly by the AL Language, and should stand out visually.

I've chosen the original `.al` language icon and applied the `binary` icon color to it (because it matches the color of `.app` files produced by AL Language, where the `.dal` files are reconstructed from).

### More Information:

I am creating this issue so I can link it to the pull request I am also creating.

vjekob:
_**Fixes #3125 **_

**Changes proposed:**

- [x] Add

This pull request provides the icon for `.dal` extension when developing in AL Language. It also makes sure that the existing `.al` icon does not apply to `.dal` files.


src/iconsManifest/supportedExtensions.ts
```diff
@@ -82,7 +82,13 @@ export const extensions: IFileCollection = {
     { icon: 'ai2', extensions: ['ai'], format: FileFormat.svg, disabled: true },
     {
       icon: 'al',
-      extensions: [],
+      extensions: ['al'],
```
robertohuertasm, line 85: Please, remove this extension as we want this to be manage by the language id. So every extension creator should be responsible for adding support to the extensions.

src/iconsManifest/supportedExtensions.ts
```diff
@@ -82,7 +82,13 @@ export const extensions: IFileCollection = {
...
+      languages: [languages.al],
+      format: FileFormat.svg,
+    },
+    {
+      icon: 'al_dal',
+      extensions: ['dal'],
```
robertohuertasm, line 91: In this case, as we want to enforce the usage of `al_dal` icon regardless of the language id, remove the `languages` line.

robertohuertasm (changes requested):
There are some changes to be applied before being able to merge it. Thanks for the effort, though!

robertohuertasm (approved):
Fixed the issues

PR #3137, opened 2022-09-09
feat(icon): add wally icon

sasial-dev:
[Wally](https://wally.run) is a package manager for [Roblox](https://roblox.com) inspired by Cargo and npm. It brings the familiar, community-oriented world of sharing code from other communities into the Roblox ecosystem.


robertohuertasm (changes requested):
The margins for the icon are not respected. Use the `rust` icon as base. That may help you. In any case, thanks for this!

robertohuertasm (approved):

PR #3171, opened 2022-10-29
feat(icon): add static to css folder icon

abhayhk2001:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Folders named static are usually used for storing CSS files of project.


sasial-dev (changes requested):
You've run some sort of formatter which has changed all of the `'` in the file to `"`. As well as not passing the lint test, it will be difficult to look through your changes due to a much larger diff...

abhayhk2001:
Sorry dint notice i will create another pull request

PR #3172, opened 2022-10-30
feat(icon): add static to css folder icon

abhayhk2001:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Folders named static are usually used for storing CSS files of project.


robertohuertasm:
HI @abhayhk2001, unfortunately I will have to close your PR. `static` is not only used for `css`. I've seen `static` folders being used for `html`, `images` and all other kind of assets.

I'd suggest you to use the [customization](https://github.com/vscode-icons/vscode-icons/wiki/Customization) feature.

PR #3180, opened 2022-11-17
feat(angular): add routes.ts as extension for routing icon

csvn:
_**Fixes #IssueNumber**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Angular 15 was just released, and support has been recently introduced where it's possible to use an array of routes directly instead of a module when lazily loading route children.

Though when using `my-feature.routes.ts`, we get the regular TS icon, compared to the Angular icon for `my-feature.routing.ts`. The former extension is used both in the Angular docs and in the release post. Sources:

- https://blog.angular.io/angular-v15-is-now-available-df7be7f2f4c8#ecb8
- https://angular.io/api/router/LoadChildrenCallback#description

Would it be possible to just add this `routing.ts` <-> `routes.ts` "synonym", to share the same icon for these?


src/iconsManifest/supportedExtensions.ts
```diff
@@ -253,6 +253,24 @@ export const extensions: IFileCollection = {
       format: FileFormat.svg,
       disabled: true,
     },
+    {
+      icon: 'ng_routing_dart',
+      extensions: ['routes.dart'],
```
robertohuertasm, line 258: Is there a reason for this duplication of entries? If I understood this right, just adding `routes.dart`, `routes.js`, etc, to the existing entries would be enough.

Thanks for your contribution!

robertohuertasm (changes requested):

src/iconsManifest/supportedExtensions.ts
csvn: Thanks for pointing that out! I think I got confused since there were multiple entries that used e.g. `ng_routing_ts` as icon.  I didn't think of the `filename` property being the reason they might have been kept separate.

```ts
    {
      icon: 'ng_routing_ts',
      extensions: ['routes.ts'],
      format: FileFormat.svg,
      disabled: true,
    },
    // ...
    {
      icon: 'ng_routing_ts',
      extensions: ['app-routing.module.ts'],
      filename: true,
      format: FileFormat.svg,
      disabled: true,
    },
```

I've updated the PR with your suggestion, to just add `routes.(ts|js|dart)` to existing `extensions` arrays for existing entries. Let me know if I missed anything else 🙂

PR #3204, opened 2023-01-03
feat: Add `.mjs` to Gatsby

LekoArts:
Hi!

As of Gatsby 5.3 https://www.gatsbyjs.com/docs/reference/release-notes/v5.3/ there's now `gatsby-config.mjs` and `gatsby-node.mjs`. So I'm updating the current map.

In the future we'll most likely also support `.mts` but when that is the case I'll open a separate PR.

Thanks!

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


robertohuertasm (approved):

PR #3205, opened 2023-01-05
add .bazelversion & .bazelignore

helbing:
_**Fixes #IssueNumber**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm:
I'm wondering if it wouldn't be better to create specific variants of the icon for that. Maybe a different color would suffice.

PR #3221, opened 2023-01-23
feat(tree): support for .tree extension

robertohuertasm:
Fixes #3117

_**Fixes #3117**_

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


Hacksore:
Hey @robertohuertasm thanks for putting this PR up, great work!

The icon does slightly differ from the icon we created for the [tree extension](https://marketplace.visualstudio.com/items?itemName=CTC.vscode-tree-extension).

Should we either make the vscode-icons icon match or should I update the extensions icon to your design?

cc @alex-oser

robertohuertasm:
Ideally, we should be using an icon which is identical to the one you have. Open an issue or submit a PR with the change if you want and we'll take it from there.

Issue #2776, opened 2021-05-19
[FR] Change `.jsx` and `.tsx` icons to framework-agnostic icons


wenfangdu:
Change `.jsx` and `.tsx` icons to framework-agnostic icons, currently, both are bound to react, however many other frameworks like Vue, etc. use `.jsx`, `.tsx` too. When writing jsx/tsx in these frameworks while seeing a react icon feels really weird.

va3y:
I also would really want to make this happen. 

JSX/TSX is it's own language [spec](https://facebook.github.io/jsx/). React is the most popular framework that uses it, but there are other javascript libraries that utilise JSX syntax. I think changing these icons to framework-agnostic could generally a bit negate this misunderstanding. 

My idea is to go for the icon similar to this: 
<img src="https://raw.githubusercontent.com/Protectator/jsx-tsx-logos/10c730252cc053cfe5be76d588426c26a7a51807/jsx-logo.svg"  width="100" height="100">

Perhaps, without the background, like for the normal .js and .ts files. Actually, I think I'll sketch-up some svgs and make a PR in a few days, if that's ok.

PR #3240, opened 2023-02-23
feat(jsx): update jsx to framework-agnostic logo

va3y:
_**Addresses #2776**_

This PR replaces React icon for framework-agnostic JSX/TSX text. The used color is the same as for normal js/ts icons. 

Many people use React and JSX, so I'll appreciate feedback from the community and maintainers about the idea. 

What also bugs me is that vscode internally treats jsx as `javascriptreact`, and it can be seen in a language-selection window: 

This maybe will needed to be addressed in the future, but I think replacing default icon is already a good step. This PR doesn't replace default namings in `INativeLanguageCollection`

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


thetarnav:
For context, the filename for .jsx and .tsx was recently changed in vscode:
issue: https://github.com/microsoft/vscode/issues/138285
PR: https://github.com/microsoft/vscode/pull/168368

Hacksore:
Including an example of all files togther to compare, I like this!

image

hiteacheryouare:
Will there at least be a configuration option to keep it as the current react icon?

JimiC:
@vscode-icons/development This makes total sense to me, since `JSX` and `TSX` is being used outside `React` itself. What we can do, to keep backwards compatibility, is to include this into the icon `selection` mechanism (you know, the old/new icons selector).

JamesonDiamond1 (approved):
@JamesonDiamond1

PR #3260, opened 2023-03-30
feat(icon): add Snaplet icon

jgoux:
Adds icons for [Snaplet](https://www.snaplet.dev/) configuration files and folder

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm:
Hi @jgoux, thanks for this! I've got a couple of questions though. 

1. Is the circle around the cat part of the official logo?
2. Given that the icon has already a circle surrounding the white cat, I'm not sure if the light icon is really needed.

almariestander:
Hi @robertohuertasm & @jgoux. Yes, the Iris circle is part of the official logo and the cat is white. Your documents recommended however that we use a light grey instead of white. This works well in dark mode but it doesn't look good against a light background. That is why we made a grey cat version for dark mode and a white cat version for light mode. If we decide to use just one version, I would recommend we use the white cat instead of the grey cat for both.

robertohuertasm:
Hi @almariestander thanks for the clarification. Normally we use grey for light themes support, to stop some people to provide white icons that won't work in light themes. This is not clearly the case, so I guess we can use the white cat for all the themes provided that it works well for both kind of themes.

robertohuertasm:
btw, really cool branding! 😊

jgoux:
@robertohuertasm I updated the PR to only use the white variant for both themes. I tested in vscode and it works fine for both:

image

image

robertohuertasm (approved):

PR #3278, opened 2023-04-17
feat(apps): use the same icon as for `app`

afranche:
If we work in a mono-repo such as encouraged by [Nx](https://nx.dev/) or [Turborepo](https://turbo.build/repo), one may own an `apps` folder to gather its packages under. Since we provide a different icon for `app` but not `apps`, I thought we might as well provide it here.

**Changes proposed:**

- [x] Add
  - Provide the same icon between `app` and `apps`
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):

PR #3298, opened 2023-05-16
feat(vitest): support vitest workspaces

remcohaszing:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

This applies the Vitest icon to [Vitest workspace](https://vitest.dev/guide/workspace.html) files.

This supports some more file extensions than are officially documented. (https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/constants.ts#L19-L35)


robertohuertasm (approved):

PR #3331, opened 2023-08-07
Add icon for .tsbuildinfo files

### LINKED ISSUE #3009: [Icon Request] Support for tsconfig.tsbuildinfo
  - Type: `file`
  - Icon Name: `filename`
  - Sample original Icon: This should probably be a variant of the existing TypeScript icon.

- More info:

  The TypeScript command `tsc --build` generates this file. It’s generally VCS ignored, but it would still be nice to have an icon for it.

remcohaszing:
I used the `tsconfig` icon, but made the `TS` characters green like the `typescriptdef` icon to indicate this is a generated file.

In the issue I stated the file was called `tsconfig.tsbuildinfo`. This is typically true, because often people use `tsconfig.json`, but they don’t have to. TypeScript will use the base name of the configuration file and use the `.tsbuildinfo` extension.

_**Fixes #3009**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):

PR #3338, opened 2023-08-11
Add some more files to be identified as docker-compose

LinusDierheimer:
Those are commonly used in a lot of projects i work on. Figured others may benefit from that too.


JimiC:
@robertohuertasm In my opinion we should stop accepting this type of PR.

This can easily be supported by doing:
```
    {
      "icon": "docker",
      "extensions": [
        "docker-compose.dev.yml",
        ... # add here any other filenames
      ],
      "filename": true,
      "format": "svg"
    }
  ]
```

Ain't that why we added that functionality in the first place?
We should even remove the rest that are already defined and keep only `docker-compose` as originally intented.
Thoughts?

LinusDierheimer:
I made this pull request, because of this line in your [wiki](https://github.com/vscode-icons/vscode-icons/wiki/Customization):
> Although you now have this power, we're encouraging everyone to raise an issue in the issues section of the GitHub's repo in case you find any of your customizations valuable to the rest of the community so we can implement them out of the box.

I totally agree that adding more and more names is neither a pretty solution, nor really future proof. For example just have a look at the [webpack icon](https://github.com/vscode-icons/vscode-icons/wiki/ListOfFiles#webpack).

However, i also think that an icon extensions should provide icons for as many files as possible for convinience reasons. Having to add a lot of custom configurations per project / vscode installation is not a good experience. And until vscode fixes [this issue](https://github.com/microsoft/vscode/issues/12493), adding more names seems to be the only way.

JimiC:
I believe the pressure should be towards `vscode` devs themselves and not the extension authors.

PR #3369, opened 2023-11-01
feat(icon): add bruno icon

oseughu:
**Changes proposed:**

- [x] Add

Bruno is a Fast and Git-Friendly Opensource API client, aimed at revolutionizing the status quo represented by Postman, Insomnia and similar tools out there. You can read more about it [here](https://usebruno.com)

It allows you edit your requests in markup language identified with the ".bru" file extension. This can be done in an editor such as VsCode.

I didn't find the icon in vscode-icons repo so I thought it would be nice to contribute to it here.

Thank you for your great work. LGTM 🚀


src/iconsManifest/supportedExtensions.ts
```diff
@@ -788,6 +788,12 @@ export const extensions: IFileCollection = {
       filename: true,
       format: FileFormat.svg,
     },
+    {
+      icon: 'bruno',
+      extensions: ['bru'],
```
robertohuertasm, line 793: The extension is not needed given that you have provided a language id. This means that the Bruno extension will expose that id and set the proper extensions (I checked it out and that's already in place).

So basically, you have to decide what to do, either use language id or declare the extension.

robertohuertasm (changes requested):
The icon needs rework. Icons shouldn't be bigger than 2Kb. Also, take a look [here ](https://github.com/vscode-icons/vscode-icons/wiki/Contributing) for the icon guidelines. Thanks for your contribution.

src/iconsManifest/supportedExtensions.ts
remcohaszing: This is true, but leaving out `extension` means the icon would only be applied if the Bruno extension is installed. Leaving out `languages` means the icon is not applied to the `bruno` language in the language selector

robertohuertasm: Yes, that's right. We have been historically not adding the extension if there's a language id already supporting it. This is basically because the extension has priority over the language id. So if the extension owner then decides that a particular extension is no supported anymore but we still have it,  we will have to actively remove it. The idea is that we wouldn't have to pay the price of maintaining the right extensions if possible.

robertohuertasm: Also, we usually default to the language id in our decission process when that alternative exists

robertohuertasm: It seems that the icon is not easily optimizable. Let's keep it as it is for the moment.

robertohuertasm (approved):

PR #3403, opened 2023-12-13
feat(icon): add vento icon

cameronhimself:
**Changes proposed:**

- [x] Add

https://vento.js.org/

Templating language that is the default for [lume](https://lume.land/).


robertohuertasm (approved):

PR #3408, opened 2023-12-30
feat(icon): support the go.mod and go.work languages

remcohaszing:
This adds support for the go.mod and go.work languages registered by https://marketplace.visualstudio.com/items?itemName=golang.Go

The file names were already supported. The icon is now also visible in the language selector.

There is also the related go.sum language, but the matching filename is sometimes associated with the go.mod icon, and sometimes with the go.work icon.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):

PR #3410, opened 2024-01-04
feat(icon): add miva icons

mghdotdev:
Adds icons for Miva Script and Miva Template Language.

https://marketplace.visualstudio.com/items?itemName=mhegler.vscode-miva-ide

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


src/iconsManifest/supportedExtensions.ts
```diff
@@ -2942,6 +2942,40 @@ export const extensions: IFileCollection = {
...
+      ],
+      languages: [languages.mvtcss],
+      format: FileFormat.svg
+    },
+    {
+      icon: 'mvtjs',
+      extensions: [],
+      extensionsGlob: [
+        'mvt',
+        '*.js.mvt',
```
robertohuertasm, line 2973: Unfortunately this is not supported

src/iconsManifest/supportedExtensions.ts
```diff
@@ -2942,6 +2942,40 @@ export const extensions: IFileCollection = {
...
+    },
+    {
+      icon: 'mvtcss',
+      extensions: [],
+      extensionsGlob: [
+        'mvt',
+        '*.css.mvt',
+        'css/*.mvt'
+      ],
+      languages: [languages.mvtcss],
```
robertohuertasm, line 2965: If you declare a language id, there's no need to use extensions at all. Usually, a language id is then managed via an extension. In this case, the Miva extension seems to be declaring some language Ids and its related extensions. That's all that we need.

robertohuertasm (changes requested):
Review the format and check out the comments. Let me know if you need more guidance. Good icons!

src/iconsManifest/supportedExtensions.ts
mghdotdev: These are the settings we use internally to get the icons showing for mvtjs/css files since they technically share the same `.mvt` extension. The only way we know they should be `mvtjs`/`mvtcss` syntax and associated icon is to check if they are contained within the `js` or `css` folder in our template directory.

```
	{
		"icon": "mivascript",
		"extensions": [
			"mv"
		],
		"languages": [
			{
				"ids": "mv",
				"defaultExtension": "mv"
			}
		],
		"format": "svg"
	},
	{
		"icon": "mvt",
		"extensions": [
			"mvt"
		],
		"languages": [
			{
				"ids": [
					"mvt"
				],
				"defaultExtension": "mvt"
			}
		],
		"filename": true,
		"format": "svg"
	},
	{
		"icon": "mvtjs",
		"extensions": [
			"js/*.mvt"
		],
		"languages": [
			{
				"ids": [
					"mvtjs"
				],
				"defaultExtension": "mvt"
			}
		],
		"format": "svg"
	},
	{
		"icon": "mvtcss",
		"extensions": [
			"css/*.mvt"
		],
		"languages": [
			{
				"ids": [
					"mvtcss"
				],
				"defaultExtension": "mvt"
			}
		],
		"format": "svg"
	}
]
```

What would be the correct translation of these settings to the `supportedExtensions.ts` file version?

robertohuertasm: Given that you have an extension and you have added the language ids, there's no need to define the extensions. Those will used from the miva extension.

robertohuertasm (approved):

robertohuertasm (approved):

PR #3419, opened 2024-01-14
feat(icon): add juice icon

DevVali:
This adds the Juice logo (a template engine for https://github.com/Lemon-Framework).

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


DevVali:
I could not make the files smaller than 8 kB for this.

robertohuertasm:
I couldn't find any extension exposing the `juice` language id. If that's the case, there's no need to add the language id.

robertohuertasm:
Optimized it down to 5.6Kb

robertohuertasm:
Optimized down to 713 bytes

robertohuertasm (approved):
LGTM 🚀 Thanks for your contribution ❤️I optimized the icon and removed the language id.

DevVali:
Awesome, thanks!

DevVali:
> I couldn't find any extension exposing the `juice` language id. If that's the case, there's no need to add the language id.

Is that field for a VSCode extension if I understand correctly?

robertohuertasm:
Let's say that I build a Lemon/Juice extension that would support the framework somehow. I, as the extension creator, would declare a language id in the `package.json` and would be able to define which extensions are supported. By doing this, the vscode-icons extension would show the icon for the extensions that the lemon extension defines. There would be no need to keep maintaining the list of file extensions.

PR #3425, opened 2024-01-21
feat(icon): support the search-result language

remcohaszing:
The `search-result` language is a language that’s builtin to VS Code. It’s used for the advanced code search results. The icon was extracted from the VS Code code base.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm:
@remcohaszing a question about the light icon. It feels like the color variation is minimal. It almost seems the same as the normal one. Was the icon set from VS Code like that already?

remcohaszing:
The icon itself was extracted from https://github.com/microsoft/vscode/blob/main/extensions/theme-seti/icons/seti.woff. The colors come from https://github.com/microsoft/vscode/blob/main/extensions/theme-seti/icons/vs-seti-icon-theme.json#L193-L200. So for VSCode the impact to change a color is less impactful than for `vscode-icons`.

I agree the color difference is very minimal. IMO it would be fine to either merge this as-is or pick one color, where I don’t have a strong preference which one.

robertohuertasm (commented):
Let's get rid of the light one.

robertohuertasm (approved):

PR #3427, opened 2024-01-21
chore: update to package-lock v3

remcohaszing:
Version 2 of `package-lock.json` was a hybrid format which includes the content of both version 1 and 3. Versions 8 and greater of npm support lockfile version 3.


robertohuertasm:
I'm not sure about this change. We're [tied to using node v16.14.2](https://github.com/vscode-icons/vscode-icons/blob/master/package.json#L34) and the `package-lock.json` file that its npm version generates is v2. I'd probably keep it that way.

remcohaszing:
What really matters is the npm version, not the Node.js version. Node.js 16.14.2 ships with npm 8.5.0, which supports `package-lock.json` version 3.

```
remco@vali:~ $ docker run -ti --rm --entrypoint bash node:16.14.2
root@dfa3e12a181d:/# node --version
v16.14.2
root@dfa3e12a181d:/# npm --version
8.5.0
root@dfa3e12a181d:/# mkdir app
root@dfa3e12a181d:/# cd app
root@dfa3e12a181d:/app# echo '{}' > package.json
root@dfa3e12a181d:/app# echo 'lockfile-version = 2' > .npmrc
root@dfa3e12a181d:/app# npm i

up to date, audited 1 package in 176ms

found 0 vulnerabilities
root@dfa3e12a181d:/app# cat package-lock.json 
{
  "name": "app",
  "lockfileVersion": 2,
  "requires": true,
  "packages": {}
}
```

You may have run into issues before with Node.js 14 though. That ships with an npm version that doesn’t support `package-lock.json` v3.

robertohuertasm:
I know, what I mean is that npm 8.5 defaults to v2. At least for what I've been able to test.

remcohaszing:
That is correct. The value used is, in this order:

1. `--lockfile-version` CLI flag
2. `lockfile-version` option in `.npmrc`
3. Existing `lockfileVersion` in `package-lock.json`
4. 2 (npm 8) / 3 (npm 9 and 10)

robertohuertasm:
So that's why I think we should keep it as it is. Is there any benefit aside from the file side?

remcohaszing:
Lockfile v2 means you're essentially using 2 lockfiles that need to be kept in sync. Diffs are more noisy. All versions of npm that ship with non EOL Node.js versions, default to lockfile v3. All npm versions needed by this project support both lockfile versions.

Version 2 works, but there are no downsides to using version 3.

robertohuertasm:
Honestly, I don’t see the need for this change. I prefer to stick to the file version that the npm version we use defaults to.

remcohaszing:
That's ok. I just thought this would be a quick and simple small improvement.

robertohuertasm:
I understand the motivation. In my opinion the benefits of that change are not worth the cons. Let’s say that a contributor needs to recreate the package-lock.json file. He will have to understand that we’re using v3 and then use another npm version (considering he’s using the node version we use). That, or either we need to change the .npmrc file.

In my opinion too much things for a small benefit: minimal diffs and file size.

Minimal in the sense that we don’t usually have to deal much with package-lock.json file changes.

In any case, I really appreciate your invaluable help! Your contributions are amazing! ❤️

JimiC:
I have similar opinion with @robertohuertasm on this. 
Upcoming September, hopefully, where we will update the minimum `vscode` and `node` versions, you will get your wish.

remcohaszing:
Only the npm version matters, not the VSCode or Node.js versions. The workflow that uses Node.js 14 updates npm to a version that’s compatible with lockfile version 3.

JimiC:
https://github.com/vscode-icons/vscode-icons/pull/3432#issuecomment-1952779240

robertohuertasm:
Closing this for now.

PR #3432, opened 2024-01-21
chore: update to Node.js 16.14.2 in wpil workflow

remcohaszing:
Before, Node.js 14 was used. A step would update npm to the latest version. However, npm 10, the latest version of npm at the moment of writing, doesn’t support Node.js 14.

Instead of using such an old version of Node.js, the Node.js version was updated to 16.14.2. This is already EOL, so ideally an even more recent version would be used. The npm shipped with this Node.js version is probably recent enough, so the npm update step was removed.

### DIFF
--- a/.github/workflows/wpil.yml
+++ b/.github/workflows/wpil.yml
@@ -21,7 +21,7 @@ jobs:
           node-version: ${{ env.NODE_VERSION }}
 
       - name: Upgrade npm
-        run: npm i -g npm@latest
+        run: npm i -g npm@9
 
       - name: Cloning repository
         uses: actions/checkout@v3
--- a/.github/workflows/wpil.yml
+++ b/.github/workflows/wpil.yml
@@ -1,7 +1,7 @@
 name: Update Wiki Page Icons List
 
 env:
-  NODE_VERSION: 14.21.3
+  NODE_VERSION: 16.14.2
 
 on:
   push:
@@ -20,9 +20,6 @@ jobs:
         with:
           node-version: ${{ env.NODE_VERSION }}
 
-      - name: Upgrade npm
-        run: npm i -g npm@latest
-
       - name: Cloning repository
         uses: actions/checkout@v3
         with:
@@ -39,4 +36,3 @@ jobs:
         run: npx --yes vscode-icons/wpilgenerator
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
--- a/.github/workflows/wpil.yml
+++ b/.github/workflows/wpil.yml
@@ -1,7 +1,7 @@
 name: Update Wiki Page Icons List
 
 env:
-  NODE_VERSION: 16.14.2
+  NODE_VERSION: 14.21.3
 
 on:
   push:
@@ -20,6 +20,9 @@ jobs:
         with:
           node-version: ${{ env.NODE_VERSION }}
 
+      - name: Upgrade npm
+        run: npm i -g npm@latest
+
       - name: Cloning repository
         uses: actions/checkout@v3
         with:
@@ -36,3 +39,4 @@ jobs:
         run: npx --yes vscode-icons/wpilgenerator
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+


JimiC:
In order to update the workflow we need to first update the projects node engine minimum requirement.
To update the projects node engine we need to first update the `nodegit` min requirement.
And for the time being we are stuck at version `0.27` which only provides compiled binaries based on node v14.
I see some `0.28-alpha` releases but we won't upgrade unless there's a stable release.

Hope I have provided some answers.

remcohaszing:
Ah yes, this does provide some answers. So instead of updating the Node.js version, the npm version should be pinned to 9. I added a new commit to do that.

JimiC:
`node` engine is tied to the version the `vscode` uses.

remcohaszing:
The `node` engine is [`>=16.14.2`](https://github.com/vscode-icons/vscode-icons/blob/master/package.json#L34), but you’re saying `nodegit` requires Node.js 14.

JimiC:
Aren't we talking about `wpil workflow` here? The node version is `14.21.3`

JimiC:
> The `node` engine is [`>=16.14.2`](https://github.com/vscode-icons/vscode-icons/blob/master/package.json#L34), but you’re saying `nodegit` requires Node.js 14.

That `node` engine is tied to the one `vscode` uses.

remcohaszing:
These are the steps of the [`wpil` workflow](https://github.com/vscode-icons/vscode-icons/pull/3432/files):

- :heavy_check_mark: Setup Node.js 14
- :heavy_check_mark: Update to npm 10
- :heavy_check_mark: Clone vscode-icons/vscode-icons
- ❌ npm install dependencies
  - ❌ [Syntax error](https://github.com/vscode-icons/vscode-icons/actions/runs/7520531656/job/20470324940#step:5:11): npm 10 requires Requires newer Node.js
  - If solved:  
    ❌ Engines error: Project requires Node.js 16 or greater according to the [engines](https://github.com/vscode-icons/vscode-icons/blob/master/package.json#L34) field

JimiC:
> These are the steps of the [`wpil` workflow](https://github.com/vscode-icons/vscode-icons/pull/3432/files):
> 
> * ✔️ Setup Node.js 14
> * ✔️ Update to npm 10
> * ✔️ Clone vscode-icons/vscode-icons
> * ❌ npm install dependencies
>   
>   * ❌ [Syntax error](https://github.com/vscode-icons/vscode-icons/actions/runs/7520531656/job/20470324940#step:5:11): npm 10 requires Requires newer Node.js
>   * If solved:
>     ❌ Engines error: Project requires Node.js 16 or greater according to the [engines](https://github.com/vscode-icons/vscode-icons/blob/master/package.json#L34) field

Unsolvable then. Until we can update `nodegit` nothing can be done.

JimiC:
Or does updating to npm@9 solves your issue?

remcohaszing:
Updating to npm 9 or lower (or removing that step entirely) solves the first problem. Removing the `engines` field solves the other. AFAIK the `node` engine field doesn’t mean anything for VSCode extensions, only the `vscode` engine does.

remcohaszing:
Not updating to `npm@latest` fixes one issue at least. Even if the workflow can update to a Node.js 16, at some point, eventually `npm@latest` will remove support for that too.

I see the `npx vscode-icons/wpilgenerator` command hangs with the npm version that comes bundled with Node.js 14. I imagine that’s why npm is updated in the `wpil` workflow. So this must stay, just not use `latest`. Merging this as-is at least resolves that issue.

I don’t understand why you’re so keen on enforcing `engines`. The only thing it does is force contributors and workflows to use Node.js 16 or greater. But taking this as a given requirement, another solution is to disable `engines-strict` either in `.npmrc` or in the `wpil` workflow using a CLI flag.

JimiC:
> The only thing it does is force contributors and workflows to use Node.js 16 or greater.

This. It's vital.

robertohuertasm:
Thank you @remcohaszing for this and your other contributions. I really appreciate them. Since #3452 has been merge, I'm proceeding to close this PR.

PR #3443, opened 2024-02-08
feat(icon): support the raku language

remcohaszing:
The [`raku`](https://raku.org) language is builtin to VSCode. I managed to trim down the original icon significantly, but it’s still pretty big. I don’t know how to trim it down further though.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):
Difficult to optimize this one. Let's leave it like that for now.

PR #3463, opened 2024-03-15
feat(icon): register the vim-help and vim-snippet languages

remcohaszing:
Both the `vim-help` and the `vim-snippet` languages are registered by https://marketplace.visualstudio.com/items?itemName=XadillaX.viml.

Because `vim-help` applies to `.txt` files matching a pattern, I thought it makes sense to treat them the same as `plaintext`. `vim-snippet` files uses extensions that may conflict with others, so I just added the language ID to the `viml` language.

This extension also registers the `vim-markdown` extension, which it shouldn’t (https://github.com/XadillaX/vscode-language-viml/pull/67).

This extension also registers the [`vroom`](https://github.com/google/vroom) language. I could only find a PNG icon for this. It’s just text with some colors, but I don’t know the font.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):
LGTM 🚀

PR #3489, opened 2024-04-25
feat(icon): add icon for rnc files

remcohaszing:
RNC files are used for XML validation. There is an icon available on https://github.com/relaxng, but it’s very detailed and there’s no SVG version available. Instead the same icon is used as for XML, but with the same cyan color as is used in the original icon.

The `rnc` language ID is registered by
https://marketplace.visualstudio.com/items?itemName=redhat.vscode-xml

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):

PR #3526, opened 2024-06-16
feat(icon): add rss icon

remcohaszing:
Both the extension and the icon were taken from https://en.wikipedia.org/wiki/RSS.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):

PR #3531, opened 2024-06-17
feat(icon): add vyper icon

itsacoyote:
This adds the Vyper logo from https://vyperlang.org/

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


src/iconsManifest/supportedExtensions.ts
```diff
@@ -5386,6 +5386,12 @@ export const extensions: IFileCollection = {
       filename: true,
       format: FileFormat.svg,
     },
+    {
+      icon: 'vyper',
+      extensions: [],
```
trocher, line 5391:
```suggestion
      extensions: ['vyi'],
```

itsacoyote: I don't believe this is a valid extension for vyper?  Is there an example of this usage?

itsacoyote: Ah, I found it, looks like it is for interfaces: https://docs.vyperlang.org/en/latest/interfaces.html#standalone-interfaces

robertohuertasm (approved):

PR #3552, opened 2024-07-12
Add liara cloud icon

jarqvi:
We have a cloud that sometimes requires our users to create files named liara.json and .liaraignore.
That's why we added our icon.


robertohuertasm (approved):
Thanks for your contribution ❤️ I had to update the icon size, so it would keep our recommended size.

jarqvi:
> Thanks for your contribution ❤️ I had to update the icon size, so it would keep our recommended size.

ok thanks @robertohuertasm

robertohuertasm (approved):

PR #3578, opened 2024-08-24
Adjustment of yaml/yml file extensions.

moraisbrian:
Adjustment of yaml/yml file extensions. Included the yaml extension in places that only recognized yml and yml in places that only knew yaml.

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


moraisbrian:
Codeclimate failed, I don't know why. All tests passed locally.

PR #3583, opened 2024-08-26
feat(icon): use lighthouse icon for budget(s).json

remcohaszing:
Both `budget.json` and `budgets.json` are used by Lighthouse CI. Although there’s no default value, it’s a common convention to use one of these file names. See
https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/configuration.md

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):

PR #3590, opened 2024-09-13
Add '.yaml' as alternative supported extension for 'gitlab-ci.yml'

mbalc:
_**Fixes #IssueNumber**_

**Changes proposed:**

- [X] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Although [the official GitLab documentation](https://docs.gitlab.com/ee/ci/yaml/) commonly references to this file as 'gitlab-ci-yml', in [the same file you can also see traces](https://docs.gitlab.com/ee/ci/yaml/#includelocal) that it also does allow .yaml to be used.

My close-sourced findings are that:
- the core, entrypoint file has to be named exactly `.gitlab-ci.yml`, so `.yaml` can't be used here
- the core, and any other files used can include other files even if the extension is `.yaml`

So we could assume that anything like`*gitlab-ci.yaml` that is not `.gitlab-ci.yml files could be used to organize partial files

### DIFF
--- a/src/iconsManifest/supportedExtensions.ts
+++ b/src/iconsManifest/supportedExtensions.ts
@@ -2007,7 +2007,7 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'gitlab',
-      extensions: ['.gitlab-ci.yml'],
+      extensions: ['.gitlab-ci.yml', '.gitlab-ci.yaml'],
       filename: true,
       format: FileFormat.svg,
     },


robertohuertasm:
hey @mbalc is this PR still in draft or is it up for review?

remcohaszing:
I’ve been a long time GitLab user and was surprised to see this PR. I also generally prefer the `.yaml` extension over `.yml` if that’s allowed. So I validated whether this is correct by renaming `.gitlab-ci.yml` to `.gitlab-ci.yaml` in a private test repo. GitLab does **not** pick up this file.

GitLab uses the `.gitlab-ci.yml` file to configure CI. This may then reference other YAML files. As described in the linked documentation, those files may have any file name with the `.yml` or `.yaml` extension. So the file name `.gitlab-ci.yaml` is related to GitLab as much as `anything-else.yaml`. IMO it would be misleading to show the GitLab icon for a file that has no special meaning in GitLab, especially if it’s so similar to a file that does.

Another hint is that the GitLab UI itself shows a GitLab icon for `.gitlab-ci.yml`, but not `.gitlab-ci.yaml`.

robertohuertasm:
Thank you @remcohaszing for your investigation! 🙇🏻 Taking this into account, I'm proceeding to close this PR. Thanks @mbalc for this anyway.

PR #3593, opened 2024-09-19
feat(icon): add language-configuration icon

remcohaszing:
Language configuration files allow VSCode extensions to configure languages. See https://code.visualstudio.com/api/language-extensions/language-configuration-guide

These file names may prefix `language-configuration.json` with anything, but vscode-icons doesn’t support globs. Official extensions that come with VSCode use `.` as a separator, so I think it’s good to support that convention for the icon as well.

The language configuration format is part of Monaco editor, which is an integral part of VSCode. The icon was taken from the toolbar on https://microsoft.github.io/monaco-editor/.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):

PR #3619, opened 2024-11-22
feat(icon): add icon for fly.toml

remcohaszing:
The `fly.toml` file is a configuration file for https://fly.io. The icon was taken from their brand page on https://fly.io/docs/about/brand/ and adheres to their guidelines. Unfortunately it does somewhat exceeds the file size limit.

Their own website uses a variation of the icon as a favicon with just the balloon and a purple background. We might consider that instead.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (commented):
I agree. We should be using the favicon version. This new icon is about 3kb. Maybe the favicon version would help to reduce its size too.

robertohuertasm (approved):
Let's merge this for now and create a new PR with the favicon.

PR #3625, opened 2024-11-26
feat: capnp file type extension

### LINKED ISSUE #3624: [Icon Request]: Cap'n Proto
### Housekeeping

### Icon Type

- [X] File Icon
- [ ] Folder Icon

### Icon Type

- [X] Extension
- [ ] Filename
- [ ] Language ID

### Original Icon

capnp

### Icon Information

Currently there is no official SVG icon and therefore it has been derived from the official PNG logo from https://github.com/capnproto/capnproto/blob/afe41c0ce712f747e7ae3a61704d543811fbfff8/doc/images/logo.png
Estension: `.capnp`

### More Information:

Icon issue request https://github.com/capnproto/capnproto/issues/2185

MatthewS8:
_**Fixes #IssueNumber**_
Fix #3624 

**Changes proposed:**

- [X] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm:
Hi @MatthewS8, thank you for your contribution but the icon would need some rework. We are aiming for 2kb icons and this icon is around 23Kb. 

Optimization for this icon may be hard, so I suggest to minimize the text. For instance, I think we could ignore the black text, and maybe think of only using something like C'n P.

MatthewS8:
I managed to reduce the original file size, unfortunately I had to simplify the paths.
I have two options: 
 - _CAP'N P_, like the file extension, that is now of 6kb
 - _C'N P_ like you suggested, that is 2kb
Which is the better option?
capnp
cnp

src/iconsManifest/supportedExtensions.ts
```diff
@@ -906,6 +906,11 @@ export const extensions: IFileCollection = {
       filenamesGlob: ['capacitor.config'],
       format: FileFormat.svg,
     },
+    {
+      icon: 'capnp',
+      extensions: ['capnp'],
```
remcohaszing, line 911: I see some VSCode extensions register the `capnp` or `capnproto` language ID. It would be better to use those instead of the file extension, so the icon shows in the language picker too.

Personally I don’t mind having both `extensions` and `languages` (that adds support for people who don’t have the accompanying VSCode extension installed), but IIRC I’ve seen in other code reviews that it’s preferred to specify just the languages.

carlocorradini:
@MatthewS8 Awesome 😎🥳🍜

src/iconsManifest/supportedExtensions.ts
robertohuertasm: The rationale here is that declared extensions have more priority than the ones declared by the extensions via language id. So if we declare, say, extension `.aaa` for language `A` and then for some reason the extension for language `A` changes to `.bbb`, we would have to change it ourselves, meanwhile if we are just relying on language ids, we move that reponsibility to the language id extensions. 

On the other hand, we don't show icons if the extensions supporting those language ids are not installed. That's the trade-off.

robertohuertasm:
I think I prefer the C'nP. It abides by the guidelines and it's also easier to read.

src/iconsManifest/supportedExtensions.ts
MatthewS8: @robertohuertasm So, is it better to add both the language and the extension or?

robertohuertasm: We usually prefer to use the language id if available, not both.

MatthewS8: Hi @robertohuertasm, @remcohaszing I was wondering about the adjustments I made based on your feedback. Is everything in order, or do any extra modifications need to be made? Thank you in advance! 😁

remcohaszing (changes requested):
Almost there! 2 small remaining details.

icons/file_type_capnp.svg
```diff
@@ -0,0 +1 @@
...7-43h-5l-46 5-15 2-14 2h-2v22h1c3-1 11-2 13-1 3 0 3-2 3 47l-1 44c-1 2-3 2-9 3l-6 1h-1v22h5l33-4 29-4h2v-22l-3 1h-8l-5 1-1-3v-34h5l35-6c9-5 17-15 21-26 1-6 2-14 1-21-1-9-4-15-9-21-6-6-13-8-23-8zm1 4 9 1c13 4 19 14 19 30 0 9-2 17-6 23-3 6-11 13-17 15-5 2-10 3-30 5l-8 1h-2v19c0 20 0 21 3 23l11 1 3-1v15l-5 1-30 3-26 2v-14l5-1c6-1 9-2 10-4 2-2 2-15 2-55l-1-39c-2-3-4-3-9-3h-7v-14l3-1a32689 32689 0 0 0 76-7zm-17 16-11 1-7 1v42h3c18-2 23-4 27-10 3-4 4-8 4-15l-1-10c-1-3-4-6-6-7l-9-2zm-1 4 7 1c5 1 8 8 7 17-1 8-4 12-9 15-3 1-10 3-16 3h-2v-35h4l9-1z"/><path fill="#fff" d="M35 244c-17-4-27-16-29-36-1-6 0-16 1-22 3-9 9-19 16-25a54 54 0 0 1 45-9l5 4c3 3 4 3 5 3l1-5 1-4 6-1h4v32h-2l-5 1h-3v-5c-1-5-3-9-6-13-4-3-7-5-13-5-8-1-14 2-19 7-8 8-11 21-10 38 1 12 4 19 9 25 5 4 8 5 14 5 14 1 22-8 25-23l1-5h5l4-1v2c0 2-1 9-3 13-4 10-13 19-25 22-8 2-20 3-27 2zm67-98 1-1 7-6 5-11c0-1-1-1-4 1-4 4-7 3-10 0-2-4-1-8 2-11 3-2 5-3 8-3 4 0 7 2 8 6 3 9-4 21-14 25h-3zm22 85c1-1-1-5 1-4 5-1 10-1 13-4 2-5 1-10 1-16v-51c-2-3-6-3-9-3h-5c0-3-2-9 3-8l40-4 14 24 25 42c2 2 0-6 1-8l-1-49c-2-4-7-2-11-3-2 0-5 1-4-1v-6l42-4 1 7c-3 2-7 1-10 2-6 2-4 9-5 14v68l-28 2-40-70v19c0 14-1 28 1 42 1 5 7 3 11 3 2 0 5-1 4 2v6l-44 4v-4z"/></svg>
```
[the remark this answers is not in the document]
remcohaszing: No worries! Your contribution is appreciated and you’re doing a great job contributing and processing feedback ♥️

MatthewS8:
I have tried to work on that icon again, in this commit https://github.com/vscode-icons/vscode-icons/pull/3625/commits/464eb43fe31bf63cee303e5f5a01f3c681548507 the icon was not good. Lmk what you think about this one now https://github.com/vscode-icons/vscode-icons/pull/3625/commits/4268726520e866de6286a87aeba1bf965c9ebb16

remcohaszing (approved):

robertohuertasm (approved):

MatthewS8:
Awesome! Ty guys 🚀

PR #3627, opened 2024-11-26
feat(icon): support biome_syntax_tree language

remcohaszing:
The Biome extension registers the `biome_syntax_tree` language. This isn’t typically used for files, but exists in the language selector. It now uses the biome icon.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):

PR #3636, opened 2024-12-11
feat: add bun.lock to bun file types

jakeboone02:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

This PR adds a capture for `bun.lock`, the upcoming default lock file format for Bun (available in canary today). See https://x.com/bunjavascript/status/1866879040424120625

PR #3644, opened 2024-12-17
feat(icon): add bun.lock for bun

nguyenyou:
🎉 We now have bun lock as text: https://bun.sh/blog/bun-lock-text-lockfile

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

### DIFF
--- a/src/iconsManifest/supportedExtensions.ts
+++ b/src/iconsManifest/supportedExtensions.ts
@@ -837,7 +837,7 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'bun',
-      extensions: ['bun', 'bun.lockb', 'bun.lock'],
+      extensions: ['.bunfig.toml', 'bunfig.toml', 'bun.lockb', 'bun.lock'],
       filename: true,
       format: FileFormat.svg,
     },
--- a/src/iconsManifest/supportedExtensions.ts
+++ b/src/iconsManifest/supportedExtensions.ts
@@ -837,7 +837,7 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'bun',
-      extensions: ['bun', 'lockb', 'bun.lock'],
+      extensions: ['bun', 'lockb', 'bun.lock', 'lock'],
       format: FileFormat.svg,
     },
     {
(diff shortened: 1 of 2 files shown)


jakeboone02:
I added `'bun.lock'` in #3636 but it didn't seem to work:

image

I think your update may work but you might need to add `filenamesGlob: ['bun']` or something similar as well. I'm not 100% sure how the file name matching algorithm works for this extension.

nguyenyou:
@jakeboone02 
Looks like we need to add ` filename: true`

src/iconsManifest/supportedExtensions.ts
```diff
@@ -837,7 +837,8 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'bun',
-      extensions: ['bun', 'lockb', 'bun.lock'],
+      extensions: ['bun', 'bun.lockb', 'bun.lock'],
```
remcohaszing, line 840: I don’t think there’s a file named `bun`, is there? I do see there are [`bunfig.toml` and `.bundig.toml`](https://bun.sh/docs/runtime/bunfig). You might as well add those while at it.

```suggestion
      extensions: ['.bunfig.toml', 'bunfig.toml', 'bun.lockb', 'bun.lock'],
```

Also Bun for Visual Studio Code adds the [`bun.lockb`](https://github.com/oven-sh/bun/blob/b5b51004e8de80dcd43849e4da1d2623ee3d3008/packages/bun-vscode/package.json#L288) language ID. You could register the icon for that too, so it shows up in the language picker.

remcohaszing (approved):
Thanks! I approved because the change is good as-is, but I refrained from merging for now, because I think it can be improved even further.

src/iconsManifest/supportedExtensions.ts
nguyenyou: I'm not sure about how to config the language ID as you mentioned, please help!

remcohaszing (dismissed):

src/iconsManifest/supportedExtensions.ts
nguyenyou: Done 👍

remcohaszing (approved):

PR #3645, opened 2024-12-17
feat(icon): add Docusaurus config

jakeboone02:
**Changes proposed:**

- [x] Add

- [Home page](https://docusaurus.io/)
- [Icon source](https://github.com/facebook/docusaurus/blob/c5a6c26d94804c91c421904be73c1397da89be6a/website/static/img/docusaurus.svg)
- [Supported extension source](https://github.com/facebook/docusaurus/blob/c5a6c26d94804c91c421904be73c1397da89be6a/packages/docusaurus/src/server/config.ts#L21)


robertohuertasm (changes requested):
I know that this is an icon that would appear probably once in a codebase, but if you could reduce the size a little bit, it would be fantastic.  We aim for 2kb. I take it may not be possible to get that, but at least 4kb would be desirable.

jakeboone02:
Sure, let me see what I can do.

jakeboone02:
@robertohuertasm I got it down to 2.67 KB. I might actually submit a similar version to the Docusaurus team since it's virtually the same as the original but less than half the size.

jakeboone02:
Got it down below 2K with aa91d69eee889d290459a1326967f50c9d9604e4

robertohuertasm:
@jakeboone02 that was really impressive! Well done and thank you for your contribution! ❤️

robertohuertasm (approved):
LGTM 🚀

PR #3648, opened 2024-12-20
feat(icon): add the Weblate icon

remcohaszing:
See https://docs.weblate.org/en/latest/wlc.html#wlc-config

The icon was taken from https://github.com/WeblateOrg/weblate/blob/main/docs/images/logo-text.svg

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):
LGTM 🚀

PR #3654, opened 2024-12-27
Refactor(icon): swap .java and .jar icons. Fixes Issue #3653.

### LINKED ISSUE #3653: Exchange icons for .java and .jar file formats
### Housekeeping

### Information:

From a developer point of view, one is more likely to encounter .java file format (source files) rather than a .jar file format. However the extension uses the properly colored java logo for .jar rather than .java.

A simple exchange of icons between the two would be perfect.

dasTanmoy0096:
_**Fixes #3653**_

**Changes proposed:**
- [X] Fix

From a developer standpoint, .java (java source) files are encountered far often than .jar (java archive) files. So it makes sense for .java files to have the properly colored logos. This PR swaps the icons between the two and makes the necessary changes.


remcohaszing:
I get your point, and I would also prefer to have the colored icon for Java files. I do have my doubts about using the monotone icon for `.jar` files though. `.jar` files are essentially just zip files that have a special meaning to Java.

There are more similar files:
- `.whl` files are just zip files with a special meaning for Python
- `.deb` files are just archive files with a special meaning for Debian

Perhaps we should create icons for these files that are based on the archive icon, but with an additional icon in the lower right corner, akin to certain configuration icons and folder icons.

@robertohuertasm WDYT?

robertohuertasm:
I don't have a strong opinion on how to treat this kind of files, tbh. I can see the point of having special variants of the archive icon, but having completely different icons would work for me, too. 

In any case, I think this PR is good to go. We can later on think of what we want to do with this case that @remcohaszing was exposing, and then create a PR for that if needed.

JimiC:
Since I was the creator of those icons, let me give an insight on the reason for choosing them.
- The monochrome icon was made so to denote the `raw` source code written in `java`.
- The two-color icon was made to denote the compiled (and archived) version.

If you guys can come up with a different icon for the compiled (archived) version that would be great.

remcohaszing (approved):

PR #3658, opened 2025-01-03
refactor(icon): update yaml icon

remcohaszing:
This updates the yaml icon to match yaml branding in other places, such as https://yaml.org/ and https://github.com/yaml

The icon created from the SVG favicon on the official website.

**Changes proposed:**

- [x] Add


remcohaszing:
I don’t know where the old icon comes from. It does show up in some other places too. Maybe they rebranded at some point?

Anyway, I think the new icon works better as an icon as well, since it’s not as wide.

robertohuertasm (approved):
LGTM 🚀

PR #3673, opened 2025-01-24
feat(icon): add the mypy icon

remcohaszing:
The file icon is applied to `.mypy.ini` and `mypy.ini`. See https://mypy.readthedocs.io/en/stable/config_file.html.

The folder icon is applied to the `.mypy_cache` directory. See https://mypy.readthedocs.io/en/stable/config_file.html#confval-cache_dir.

The icon was taken from https://mypy.readthedocs.io/en/stable/

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (approved):
LGTM 🚀

PR #3674, opened 2025-01-24
feat(icon): add the pytest icon

remcohaszing:
The file icon is applied to `pytest.ini` and `.pytest.ini`. See https://docs.pytest.org/en/stable/reference/customize.html.

The folder icon is applied to `.pytest_cache`. See https://docs.pytest.org/en/stable/reference/reference.html#confval-cache_dir.

Previously the python folder icon was used for `.pytest_cache`.

The icon was taken from
https://github.com/pytest-dev/pytest/blob/main/doc/en/img/pytest_logo_curves.svg.

**Changes proposed:**

- [x] Add


robertohuertasm (approved):
LGTM 🚀

PR #3679, opened 2025-01-25
chore: updated the Angular icon

JeevanMahesha:
Replaced the old Angular logo with the new design.

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [X] Fix
- [ ] Prepare


JeevanMahesha:
@remcohaszing can you please review this PR

robertohuertasm (changes requested):
Thank you for this. I think you can make the icon bigger. You can expand it. Remember that we aim for a 2px padding: https://github.com/vscode-icons/vscode-icons/wiki/Contributing.

Also, the icon should be 2kb tops.

robertohuertasm:
I've updated the icon according to our guidelines and reduced the size to 2kb.

robertohuertasm (approved):

remcohaszing:
I think the new icon looks good, but we do have quite a lot of special Angular icons and special treatment for Angular projects. Other icons may need adjustment as well.

robertohuertasm:
We may. But as this one was used only for the config files I thought it was ok to merge it, even if we still keep the old custom angular versions for the angular files.

joeskeen:
Is there any plan to update the other Angular icons? Should I open an issue for that?

robertohuertasm:
@joeskeen no plans so far. Not sure how the new icon would work for that. We had color variatons for each type of Angular file. You can give it a try if you want.

PR #3681, opened 2025-01-26
chore: refactor linting setup

remcohaszing:
Prettier now checks all supported files in the project instead of a glob.

Prettier respects `.gitignore`, so `package*.json` was removed from `.prettierignore`.

ESLint now uses configuration overrides to specify additional file extensions instead of a CLI flag.

Linting no longer happens before building, so we can run tests and debug, even if there are linting errors.

The GitHub actions test workflow now specifies eslint and prettier jobs. These jobs are run once, instead of for every combination in our test matrix. This means linting issues will only show up once in pull requests instead of for every failed build job.


robertohuertasm (approved):
LGTM 🚀

PR #3693, opened 2025-01-28
feat(icon): add core folder icons and update supported folders

cychitivav:
**Changes proposed:**

- [x] Add


robertohuertasm (changes requested):
@cychitivav thank you for this. The icon needs some rework. The margins are not respected. Also, the color of the open folder is flat and should be different. Look at the other folder icons we have in the repo and check the [wiki guidelines](https://github.com/vscode-icons/vscode-icons/wiki/Contributing) for more info.

remcohaszing:
Where does the icon come from? The name `core` is pretty generic. For example there are quite a lot of npm packages named `@some-scope/core`.

cychitivav:
I have made some color modifications, but I don't understand what you mean by start and end value in the open folder icon. I also checked several icons in the repository, which all have basic fill.

> @cychitivav thank you for this. The icon needs a some rework. The margins are not respected. Also, the color of the open folder is flat and should be different. Look at the other folder icons we have in the repo and check the [wiki guidelines](https://github.com/vscode-icons/vscode-icons/wiki/Contributing) for more info.

JamesonDiamond1 (approved):
@JamesonDiamond

JamesonDiamond1 (approved):
@JamesonDiamond1

JamesonDiamond1 (approved):
@JamesonDiamond1

PR #3703, opened 2025-02-02
chore(nsri): remove integrity check

robertohuertasm:
The integrity check is not really needed for the extension to work, and it may become a [maintainability burden](https://github.com/JimiC/nsri/issues/260#issuecomment-2629110523). From that point of view, the idea is to simplify the extension and get rid of unnecessary dependencies.


remcohaszing (approved):
I think VSCode has its own integrity checks. If so, it seems pretty pointless for an extension to check themselves.

JimiC:
@remcohaszing If `VSCode` also checks the extensions integrity then yes, NSRI is redundant. I haven't followed `VSCode` improvements nor `vscode-icons` closely.

@robertohuertasm If you still feel that NSRI should be removed, after reading my [comment](https://github.com/JimiC/nsri/issues/260#issuecomment-2629315064), then feel free to do so.

remcohaszing:
I was wrong. I just tampered with some installed extensions and VSCode is fine with it. I’m still fine with removing our own integrity check though. I wonder if it affects startup performance too.

robertohuertasm:
> VSCode has its own integrity checks

Only if you change VS Code files, not for extensions. They introduced that because some extensions were chaning their source code, and this generated lots GitHub issues in their repo, and a lot of work on their side trying to spot what the real problem was. Indeed, VS Code was one of the firsts, if not the first, using that technique to bring icons into VS Code. 

Nowadays, I think the technique is still being used by some other extensions such as [Synthwave 84](https://marketplace.visualstudio.com/items?itemName=RobbOwen.synthwave-vscode).

In any case, I think we can remove `nsri`. I don't think it's solving any concern for the extension, and it's yet another dependecy we need to keep track of.

Issue #3698, opened 2025-01-30
New YAML icons are terrible (for Just too)


den-is:
### Housekeeping

- [x] I'm sure this issue is _not_ a duplicate

### Information:

I don't know how these set of icons is built, or how decisions are made here. I in general pay respect to your work!
Just, writing here to let you know about my **personal** opinion (hope some others will support me in this).

Today when I opened my VSCode my first thought was "it's some bug or some new setting was introduced" - when I saw highly contrasted, distracting icons.

New icons completely falling out of """my""" theme color palette (Solarized Light theme).

And as a DevOps, most of the days I work with YAML files.

This is disgusting
<img width="248" alt="Image" src="https://github.com/user-attachments/assets/05d3d47c-5c72-4449-a7e4-112ab077b0a9" />

Yea:
- maybe it's about "me" only
- and I will get used to it if nothing changes
- and I do have other real problems in live
- and I'm not interested in other potential philosophical replies

But:
- I would kindly ask to revert icons for YAML and Justfiles.

den-is:
Poor "yamllint", still has this pale, faint icon :S 
vscode icons should check maybe it has a contrast "black and red" icon too.
<img width="391" alt="Image" src="https://github.com/user-attachments/assets/bb7cf195-03cd-4b4b-9187-643727e91d1f" />

remcohaszing:
Hi Denis, it’s ok to dislike or disagree with changes, but be nice about it. Avoid words such as _“terrible”_ or _“disgusting”_, and don’t make it personal. Also 👎 reactions on pull requests are generally not helpful.

It occurs more often that people don’t like UI updates, including icon changes, especially if it changes something you’re used to seeing so often.

See #3658 for rationale on why the YAML icon was changed. We provide a customization mechanism for alternative icons. We welcome a pull request to add back the previous `yaml` icon as the `yaml2` icon. See #3652 for an example.

The `just` language was previously not supported at all. I imagine you were just seeing the default unknown file type icon. You can find how to customize icons in our [Custom Icons](https://github.com/vscode-icons/vscode-icons/wiki/Custom) guide.

I do not know where the original YAML or `yamllint` icons come from. I would love to have a reference.

den-is:
Hi @remcohaszing
I will choose on my own which words to avoid, you can ignore them if you don't like any of them. Not saying anything about actions of setting reactions under PRs.

For Justfile it was 🤖 (used in the original [repo](https://github.com/casey/just)) 
[...]

PR #3706, opened 2025-02-02
feat(yaml): support for official icon toggling

robertohuertasm:
Adds a new command to toggle between the official and the non-official YAML icons.

re #3698


test/app/commands.test.ts
```diff
@@ -325,6 +325,25 @@ describe('ExtensionManager: commands tests', function () {
       });
     });
 
+    context(`the toggle yaml preset command`, function () {
+      it(`toggles the yaml preset`, function () {
+        const toggleYamlPresetCommand =
+          // @ts-ignore
```
remcohaszing, line 331: IMO `@ts-ignore` should never be used. Use `@ts-expect-error` instead. That will fail if the comment is redundant.

robertohuertasm: Agreed. I was just mimicking what we already had before to make this work. I think we should have another PR focusing on how we are using @ts-ignore in the extension and see if we can mitigate its usage.

remcohaszing (approved):

test/app/commands.test.ts
JimiC: @ts-ignore was used because those functions are called via reflection and back in the day there was no other way to make ts lint silent.

PR #3720, opened 2025-02-20
feat(icon): add Lefthook icon

romario-stankovic:
_**Add Lefthook icon**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

The icon was taken from https://evilmartians.com/opensource/lefthook


remcohaszing (approved):
Thanks! Looks great!

PR #3747, opened 2025-03-28
chore(version): update version and audit fix

robertohuertasm:
bump to v12.12.0


moazzem-pixel (commented):
I need support

robertohuertasm:
@moazzem-pixel not sure if these are legit comments... in any case, if you're not more specific I will block your user as it seems you're just spamming.

moazzem-pixel (approved):

PR #3785, opened 2025-05-27
Update Ember and add Glimmer

gossi:
**Changes proposed:**

- [x] Add
- [x] Fix

I've updated the ember icon as per branding guidelines: https://emberjs.com/logos/

At the same time I added the logo for glimmer (see link above), which is used for ember's single file component format, which goes by the extensions `.gts` and `.gjs`.

The idea is to use these icons over in https://github.com/yuyinws/vitepress-plugin-group-icons to have support for ember in vitepress rsp sli.dev.

Thanks
gossi


gossi:
Thank you, that link was very helpful (I looked for `CONTRIB*.yml` and contrib section in `README.md`).

I adjusted the icons and run them through svgo.

I mapped it to the language ids - I hope I connected the dots correctly, anyway please help me in getting that right.

Thanks for assisstance.

remcohaszing (approved):

PR #3788, opened 2025-06-01
Add nsjs icon

otoneko1102:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


robertohuertasm (changes requested):
Hi @otoneko1102, thank you for this. Can you fix the icon to not use text embedded in the SVG? We should be able to see it in the Icon Background Test, which as you can see it's showing nothing at the moment. Thanks!

otoneko1102:
Hi @robertohuertasm , thank you for your reply. I fixed the icon to not use text embedded in the SVG! Thanks!

robertohuertasm (changes requested):
Please, check out the guidelines. You can take any other merged icon and start from there. We ask for a 2px space around the icon. Also, the quality of the rendered text is not good enough. What font are you using?

robertohuertasm:
@otoneko1102 let me see if I can help.

robertohuertasm:
@otoneko1102 I just saw that this is a recent library from npm. We only add icons to the main extension if those are from recognized projects. I'd suggest you to start by using the [Custom Icon ](https://github.com/vscode-icons/vscode-icons/wiki/Custom) feature, and then, once your library starts to get traction. We will be able to merge this at some point. 

For now, I'm going to close this PR. Thank you very much for your contribution, though. And all the best with noshift.js

remcohaszing:
@otoneko1102 since you also maintain a VSCode extension for your language, you should know your extension can also register an icon for your language ID. That way users don’t even need vscode-icons to get an icon for nsjs files.

PR #3789, opened 2025-06-01
feat(icon): add lqf icon

smit4k:
Add lqf (lightweight quick format) which is a configuration language

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


robertohuertasm (changes requested):
Hi @smit4k, thank you for this. Unfortunately, we cannot merge this as is. The icons need to follow our guidelines. You can check the details in the Wiki, but there's a 2px space missing around the icon.

Aside from that, you'll need to declare the extensions associated to these icons. Check out any other merged PR contributing file icons to see how it works.

Cheers!

PR #3813, opened 2025-08-02
fix(ci): coverage action upgrade

robertohuertasm:


robertohuertasm:
For the record: https://status.codeclimate.com indicates that they're good... but let's wait some days before removing the workflows just in case is temporary.

PR #3816, opened 2025-08-02
Update the Angular logo for file types

joeskeen:
_**References #3679**_

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

See https://github.com/vscode-icons/vscode-icons/pull/3679#issuecomment-3146701222

Used GitHub Copilot to adapt the icons to work with the new logo design.

I've had to wrestle with the AI to make sure all the other visual elements remain the same, but using a new Angular logo with the existing icon's color scheme. IMHO they are all looking awesome! I then went through a second time, optimizing them so that they can be under 2KB.

There are four of the files that are over 2KB: `component_sass`, `guard_dart`, `module_dart`, `pipe_dart`. As they include large third-party logos on top of them, I don't see any way to reduce the file size further without hurting fidelity. When I checked the repository before my changes both the `component_less` and `component_sass` logos were already over 2KB.

I've triple checked each of them - I am pretty sure they are all right. But please review and make sure I didn't make any mistakes.


joeskeen:
@robertohuertasm I believe you would be the best person to review this.

robertohuertasm (approved):

PR #3821, opened 2025-08-09
feat: add support for the `.astro` folder

piemot:
**Changes proposed:**

- [x] Add

Adds support for the `.astro` folder, used during development similarly to how `.next` and `.nuxt` are used in their respective frameworks.


remcohaszing (approved):
Awesome! Thanks!

PR #3873, opened 2025-10-21
Add Astro Config `.cts` + `.mts` Support

joealden:
**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

As these file extensions don't seem to be explicitly mentioned in the docs, I've tested both of these and they work.


robertohuertasm (approved):

PR #3888, opened 2025-11-10
feat(icon): mdc file type

zdodson21:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

File icons for cursor rule `.mdc` file type.

### DIFF
--- a/src/iconsManifest/supportedExtensions.ts
+++ b/src/iconsManifest/supportedExtensions.ts
@@ -3365,7 +3365,7 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'mdc',
-      extensions: ['mdc'],
+      extensions: [],
       languages: [languages.mdc],
       light: true,
       format: FileFormat.svg,


robertohuertasm:
This is a tricky one... I don’t think the Cursor team is going to change this just to resolve the extension conflict. This file extension should probably be handled in a similar way to how we deal with Angular or NestJS icons: pick a default and provide a setting to toggle the icon.

For example, we could go with the Cursor icon by default, and add a setting to enable the Nuxt one if needed, so the user can choose.

Personally, I’d lean towards Cursor, as their user base is steadily growing (I have some reliable metrics to support this), and then support Nuxt via a setting if necessary.

In my opinion, it’s better to support Cursor users, even at the cost of occasionally showing a Cursor icon for Nuxt users who happen to use that extension.

WDYT?

robertohuertasm:
> I don’t know for sure if they actually use the mdc language ID. 

You're right. I've just checked and there's no language ID in Cursor. The language id used in Cursor for those files is `markdown`. So that's the icon we are showing today.

image

If we set this as a regular file extension the Cursor icon would be shown regardless of any other consideration, which would not be good.

For now, users could do something like this if they want to see the Cursor icons on those files.

```json
    {
      "icon": "cursorrules",
      "format": "svg",
      "extensions": ["mdc"],
      "filename": false
    }
  ]
```

If we cannot rely on the language Id to set the icon, the only solution is to use the before mentioned solution or add another setting so that users can toggle the icon that is being shown for that file extension.

remcohaszing:
I think we should close this PR in favor of https://github.com/microsoft/vscode/pull/276518. I think having a markdown icon for Cursor markdown files based on the language ID is an acceptable trade-off, while keeping the option open to support Nuxt mdx files.

PR #3890, opened 2025-11-10
feat(icon): cursor folder

zdodson21:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Adds support for the .cursor folder, which can be auto generated by the Angular CLI.


remcohaszing:
> I went with the monotone icon since that is what is on the official website currently, but if preferred I can change the folder icon to use the multi-color icon instead.

I personally don’t really mind which icon is used, but I think the file icon and folder icon should be consistent.

remcohaszing (approved):

PR #3891, opened 2025-11-10
feat(icon): gemini folder

zdodson21:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Adds gemini folder icon support for `.gemini` folder which can be auto generated with Angular CLI.

Note: I understand if this is rejected due to file sizes, but I did use the Gemini icon from this code base which is where much of the file size seems to come from (~7.8kb for Gemini icon alone).


remcohaszing:
Thanks!

I think our folders are usually a monotone color. No reason to diverge from that now.

The Gemini icon doesn’t look too complex. I think we should be able to decrease its size significantly :thinking: I’m also not a fan of the transparency. Let us know if you need any help.

remcohaszing:
I’m ok with it if you are.

PR #3892, opened 2025-11-11
feat(icon): claude folder icon

zdodson21:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Adds claude folder icon, `.claude` folder can be auto generated with Angular CLI. File sizes slightly over preferred limit due to I assume complexity of Claude icon (Claude icon itself is 1.3 kb)


robertohuertasm:
@zdodson21 thank you for this! ❤️ I have some doubts about the contrast between the Claude icon and the folder. I'm afraid it may not be enough. At least on my end, the color of the icon and the color of the folder is too similar, and I'm having difficulties to distinguish it. @remcohaszing wdyt?

remcohaszing (approved):

PR #3906, opened 2025-11-23
fix: add support for TS files for capacitor config icon

zdodson21:
**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

Capacitor icon would only appear for `capacitor.config.json` file. This adds support for the [supported](https://capacitorjs.com/docs/config) `capacitor.config.ts` file.


robertohuertasm (approved):

PR #3926, opened 2025-12-06
feat(icon): gimp file icon

zdodson21:
_**Fixes #IssueNumber**_

**Changes proposed:**

- [x] Add

Saw icons were available for some Adobe products, figure it would be good to add support for the GIMP icon too.


robertohuertasm (approved):
LGTM 🚀 Thanks for your contribution ❤️ Although the icon doubles our limit, I don't think we can optimized it that much.

PR #3941, opened 2025-12-25
feat(pixi): add pixi files icon

benzlokzik:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Add support for `pixi.lock` and `pixi.toml` files which can be auto-generated with pixi package manager.


zdodson21:
file_type_pixi

Here is an SVG with a reduced file size and the required margins. The file size for this SVG still does not meet the preferred [file size requirements](https://github.com/vscode-icons/vscode-icons/wiki/Contributing) however, so more optimizing is probably needed.

robertohuertasm:
Closing this one in favor of  #3995

PR #3959, opened 2026-01-20
Add 'litcoffee' extension to CoffeeScript

otnc:
Added 'litcoffee' as a known extension for CoffeeScript.

CoffeeScript now supports the `.litcoffee` extension from v1.5.0 or higher.

**Changes proposed:**

- [x] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


src/iconsManifest/languages.ts
```diff
@@ -86,7 +86,13 @@ export const languages = {
...
   cobol: { ids: 'cobol', knownExtensions: ['cbl'] },
   codeql: { ids: 'ql', knownExtensions: ['ql'] },
-  coffeescript: { ids: 'coffeescript', knownExtensions: ['coffee'] },
+  coffeescript: {
+    ids: 'coffeescript',
+    knownExtensions: [
+      'coffee',
+      'litcoffee'
+    ],
+  },
```
remcohaszing: `knownExtensions` isn’t actually used. You should add it to `src/iconsManifest/supportedExtensions.ts` instead.

remcohaszing (changes requested):
I wonder if it would be better to use a custom icon instead that combines the coffeescript and markdown icons.

Also, should we provide an upstream PR to make VSCode treat `.litcoffee` files as markdown by default?

remcohaszing:
I’m closing this, because:

1. VSCode now treats `.litcoffee` files as markdown, meaning vscode-icons assigns it the `markdown` icon.
2. Icon associations are declared in `src/iconsManifest/supportedExtensions.ts`, not `src/iconsManifest/languages.ts`.

Feel free to open a new issue or PR if you think this isn’t the correct behaviour.

PR #3962, opened 2026-01-24
feat: optimise and update gemini icons

piemot:
**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare

Builds on #3891.

Updates the Gemini file icon:
- Removes transparency
- Removes usage of poorly-supported SVG features
- Reduces file size
	Screenshot 2026-01-24 at 12 35 13 PM

Update the Gemini folder icons:
- Removes transparency
- Updates folder colour
- Reduces file size


robertohuertasm (commented):
Thank you for this. The icon doesn't have the appropriate size. It should be 32px and it's currently 64. Checkout any other icon to compare.

PR #3970, opened 2026-02-06
feat(icon): add the skill icon

remcohaszing:
This is a builtin language ID. The icon was taken from https://github.com/agentskills/agentskills/blob/main/docs/favicon.svg.

**Changes proposed:**

- [x] Add

### DIFF
--- a/src/iconsManifest/languages.ts
+++ b/src/iconsManifest/languages.ts
@@ -560,7 +560,7 @@ export const languages = {
   sln: { ids: 'sln', knownExtensions: ['sln'] },
   silverstripe: { ids: 'silverstripe', knownExtensions: ['ss'] },
   sino: { ids: 'sino', knownExtensions: ['sn'] },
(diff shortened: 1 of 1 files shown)


robertohuertasm (approved):
LGTM 🚀

src/iconsManifest/languages.ts
```diff
@@ -560,6 +560,7 @@ export const languages = {
   sln: { ids: 'sln', knownExtensions: ['sln'] },
   silverstripe: { ids: 'silverstripe', knownExtensions: ['ss'] },
   sino: { ids: 'sino', knownExtensions: ['sn'] },
+  skill: { ids: 'skill', knownExtensions: ['md'] },
```
robertohuertasm: I'd change this and instead of using `knownExtensions` -> `md`, I would use `knownFilenames`  -> `skill.md, SKILL.md`. 

This is what we're using to build the zed version. So if we use extension .md. It will be used for all the md extensions.

PR #4000, opened 2026-03-18
feat(icon): add Bison and Flex file icons

theodevelop:
Add file icons for GNU Bison and Flex/RE-flex files.

File associations:
- Bison: .y, .yy, .ypp, .bison
- Flex: .l, .ll, .lex, .flex

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


remcohaszing (changes requested):

robertohuertasm:
@theodevelop thanks for the PR. Generally, it's better to add 2 PRs instead of 1 with 2 icons. Now we have the problem that we could merge the flex icon but we can't do the same for the Bison icon, since it's 5Kb and we're aiming for 2Kb tops.

Please, split the PRs so we can merge at least the Flex one while we figure out how to optimize the Bison icon. 

Thanks!

PR #4003, opened 2026-03-21
feat: added new icon neuralang

Abishake01:
_**Fixes #IssueNumber**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


remcohaszing:
Hi @Abishake01 :wave: 

Could you elaborate what the icon is for? What is neuralang? And where does the icon come from? Do you have an original version?

Abishake01:
> You should never tag people unless you already have ties with the project and are very sure those people will appreciate being notified. Maintainers will be notified according to their own settings. The people you tagged are not even related to this project.
> 
> Also, do you have any links? Where does the icon come from? It looks low quality and is huge in byte size.
> 
> I don’t see any Neuralang related VSCode extensions. I do see other languages with conflicting file extensions. Because of this, I don’t think we can allow extensions for the `.nl` file extension. We can only allow it for the language ID.
> 
> If you are building your own VSCode extension for your language, know that you can add an icon for your language yourself as well. This will work even if your users don’t use vscode-icons.

Thank you so much, Yeah thats correct i build for my own, and i used it in my local extension... Thanks for Replying

robertohuertasm:
@Abishake01 thank you for your contribution. Unfortunately, the icon size is way to big for us. Aside from that, I would consider going for what @remcohaszing suggested in his previous messege:

>If you are building your own VSCode extension for your language, know that you can add an icon for your language yourself as well. This will work even if your users don’t use vscode-icons.

Considering that this is for a personal project that doesn't have massive adoption yet, and you have already an extension and can ship the icon with it, I'm closing the PR. Thanks!

PR #4006, opened 2026-03-23
feat(icon): support .biome.json and .biome.jsonc

mkvlrn:
Support `.biome.json` and `.biome.jsonc` files for the Biome icon, since they're also discoverable by the tool. [Docs here](https://biomejs.dev/guides/configure-biome/#configuration-file-resolution).

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


remcohaszing (approved):
Thanks!

PR #4030, opened 2026-04-27
feat(icon): add support for jsr.jsonc

remcohaszing:
We only had support for `jsr.json`. However, this file may also use the `jsonc` extension.

See https://jsr.io/docs/package-configuration#jsrjson-file

**Changes proposed:**

- [x] Add


robertohuertasm (approved):
LGTM 🚀

PR #4038, opened 2026-05-03
feat(icon): add LTE (Luany Template Engine) file icon

Ngola-Programador-Full-Stack:
Changes proposed:

- [x] Add

Adds file icon support for `.lte` files used by the Luany Template Engine (LTE),
part of the Luany PHP MVC Framework ecosystem (https://github.com/luany-ecosystem).

### Preview

image

image


package-lock.json
remcohaszing, line 1: Please revert the lockfile. Lockfile changes by external contributors are a security risk.

src/iconsManifest/supportedExtensions.ts
```diff
@@ -3363,6 +3363,11 @@ export const extensions: IFileCollection = {
       languages: [languages.lolcode],
       format: FileFormat.svg,
     },
+    {
+      icon: 'lte',
+      extensions: ['lte'],
```
remcohaszing: [It looks like these files use the `lte` language ID.](https://github.com/luany-ecosystem/luany-lte-vscode/blob/main/package.json#L45). Please register the icon using the language ID instead of a file extension. This way the icon will also be used for the language picker and other file patterns if user settings specify this.

icons/file_type_lte.svg
remcohaszing, line 1: This icon is over 300kB. We try to limit icon sizes to a maximum of 2kB. See https://github.com/vscode-icons/vscode-icons/wiki/Contributing for more icon guidelines.

It looks like this icon was converted from PNG to SVG. There’s probably a better way to get this done.

remcohaszing (changes requested):
Thank you for contributing!

I have some notes that need to be adressed before we can merge this.

Ngola-Programador-Full-Stack:
Hi @remcohaszing, thanks again for the review — here's a summary after a few iterations to get the icon right:

package-lock.json — reverted, matches upstream/master.
Registration — uses languages: [languages.lte] (matching the lte language ID from [luany-lte-vscode](https://github.com/luany-ecosystem/luany-lte-vscode/blob/main/package.json#L45)), correct alphabetical position.
Icon — rebuilt from scratch as a proper hand-built vector: solid circle using the actual Luany brand gradient (#633277 → #f06823, sampled directly from the brand logo), with "LTE" lettering. Final size: 568 bytes (commit 21a1130).

npm test passes locally (707 passing). Verified visually in a real Extension Development Host session across both light and dark themes.
Let me know if anything else needs adjusting.

PR #4053, opened 2026-05-16
Add support for dotenv environment variants and Azure Pipelines filename patterns

mskp:
## Description

Adds support for common dotenv environment variants and Azure Pipelines filename patterns.

### Dotenv variants

Added dedicated icons and filename mappings for:

* `.env.local`
* `.env.dev`
* `.env.development`
* `.env.test`
* `.env.qa`
* `.env.staging`
* `.env.uat`
* `.env.prod`
* `.env.production`

### Azure Pipelines variants

Added support for Azure Pipelines filename variants such as:

* `azure-pipelines-backend.yml`
* `azure-pipelines-build.yml`
* `azure-pipelines-cd.yml`
* `azure-pipelines-ci.yml`
* `azure-pipelines-dev.yml`
* `azure-pipelines-development.yml`
* `azure-pipelines-frontend.yml`
* `azure-pipelines-prod.yml`
* `azure-pipelines-production.yml`
* `azure-pipelines-qa.yml`
* `azure-pipelines-release.yml`
* `azure-pipelines-stage.yml`
* `azure-pipelines-staging.yml`
* `azure-pipelines-test.yml`
* `azure-pipelines-testing.yml`
* `azure-pipelines-uat.yml`

## Changes

* Added dotenv environment SVG icons
* Added dotenv filename mappings
* Added Azure Pipelines filename variant mappings
* Verified in Extension Development Host
* All tests passing locally (`707 passing`)

### Dotenv variants

image

### Azure Pipelines variants

image

## Checklist

* [x] Build passes
* [x] Tests pass
* [x] Icons render correctly in VS Code
* [x] Naming follows existing repository conventions


remcohaszing (changes requested):
This PR does too many things at once. It’s better to split it up, so each item can be considered separately.

Re additional Azure pipeline patterns: I’m not familiar with these. I have no strong opinions on this.

Re dotenv filenames: I’m leaning towards rejecting this. Why would we do this for dotenv, but not every other file pattern that might target multiple environments?

Re icon designs: All icons should be centered and minified.

mskp:
> This PR does too many things at once. It’s better to split it up, so each item can be considered separately.
> 
> Re additional Azure pipeline patterns: I’m not familiar with these. I have no strong opinions on this.
> 
> Re dotenv filenames: I’m leaning towards rejecting this. Why would we do this for dotenv, but not every other file pattern that might target multiple environments?
> 
> Re icon designs: All icons should be centered and minified.

Thanks for the review.

Fair point on the PR scope. The Azure Pipelines support was the original goal, and the dotenv variants got bundled in since they followed a similar matching approach. I can split those into separate PRs.

For the icons, the dotenv variants were based directly on the existing `.env` icon already present in the repo, with only the small environment overlays added on top. Me and a few friends have also been using these icons locally for quite some time, and they have been pretty helpful in practice.

Regarding the dotenv filename support, my reasoning was mainly that `.env` is one of the very few file patterns that is commonly used across almost every ecosystem with environment-specific variants like `.env.production`, `.env.local`, `.env.staging`, etc., rather than being something niche or tool-specific.

Also, since you mentioned Azure Pipelines is not something you are very familiar with, could you confirm whether those filename variants would still be considered for merge/review? If not, I would rather close the PR instead of spending more time splitting/refining it unnecessarily.

Similarly, if the dotenv variants are considered unnecessary, I can avoid opening a separate PR for those as well.

remcohaszing:
@robertohuertasm WDYT?

robertohuertasm:
I like the idea of supporting the .dotenv variants, as I tend to use some of these patterns myself, and I imagine others would benefit from them as well. That part feels fairly standard and low-risk.

Regarding the Azure Pipelines patterns, I see this as more of a borderline case. Unlike the dotenv variants, this starts to introduce the risk of a pattern explosion, where more and more filename variations get added over time, making it harder to draw a clear line and potentially introducing more opinionated choices.

Hopefully, when this PR (https://github.com/microsoft/vscode/pull/311376) finally lands, that kind of expansion won’t be necessary anymore. In this case, it feels a bit different from the dotenv situation, as it assigns a single icon to multiple filename variants, some of which may not be universally used.

Regarding the Azure Pipelines patterns, hopefully when this [PR](https://github.com/microsoft/vscode/pull/311376) finally lands, that kind of name explosion won’t be necessary anymore. In this case, I think it’s a bit different from the dotenv situation, as it assigns a single icon to multiple filename variants, and some of those prefixes can feel quite opinionated.

That said, I wouldn’t reject the PR outright, as this falls into a gray area, and there is still value in supporting some common scenarios. I would just be cautious about how far this is taken to avoid unnecessary growth in the number of patterns.

Finally, I think Remco's concerns about the PR doing too many things is fair. And also, the icons should follow our guidelines and be properly centered and minified.

PR #4062, opened 2026-05-27
fix(icon): add support for .jsonc, .ts, .mts, .cts, and .toml for cspell

NaufalK25:
_**Fixes #N/A**_

**Changes proposed:**

- [x] Fix

Add support for missing extension for [CSpell](https://cspell.org/). Here is the official docs for the config file: https://cspell.org/docs/Configuration and this PR also partially related to #3991


src/iconsManifest/supportedExtensions.ts
```diff
@@ -1428,7 +1428,19 @@ export const extensions: IFileCollection = {
...
+        'js',
+        'mjs',
+        'cjs',
+        'ts',
+        'mts',
+        'cts',
+        'yaml',
+        'yml',
+        'toml',
+      ],
```
remcohaszing, line 1451: Not all of these are supported. For example, `cspell.toml` isn’t.

To support all patterns, we need to switch from the `filenamesGlob` / `extensionsGlob` syntax to listing all filenames in `extensions`.

remcohaszing (changes requested):

src/iconsManifest/supportedExtensions.ts
NaufalK25: Should I separate it to 2 different entry, one with cspell.config and one without it?

remcohaszing: Sure, that works 👍

remcohaszing (approved):
Thanks!

PR #4069, opened 2026-05-29
chore(eslint): update to ESLint 10

remcohaszing:
This updates ESLint and all plugins to the latest version. I tried to keep the configuration in the same spirit as the original one.

ESLint formatting related rules are deprecated. I removed the ones that are handled by Prettier.

The ESLint community mostly recommends against `eslint-plugin-prettier`. I removed it. Use Prettier directly instead.

`eslint-plugin-config` disables rules that have been deprecated. This is now redundant.

I used a file pattern to disable commonly disabled rules in test files. As a result, I could remove a bunch of `eslint-disable` comments. I also made ESLint report unused `eslint-disable` directives.

The updated configuration caught a few small issues that I fixed.


robertohuertasm (approved):
LGTM 🚀

PR #4070, opened 2026-05-30
fix(icon): add .cts and .mts support to jest config

NaufalK25:
_**Fixes part of #3991**_

**Changes proposed:**

- [x] Fix

Add support for `.cts` and `.mts` to [Jest](https://jestjs.io/). [Here](https://jestjs.io/docs/configuration) are the docs that reference the config file.


remcohaszing (approved):
Thank you!

This is great to have in vscode-icons. However, as a library maintainer I want to point out you should never use Jest. If you use it, please upgrade to Vitest.

NaufalK25:
Thank you for the information! I appreciate it.

PR #4073, opened 2026-05-31
feat(icon): add vale icon

remcohaszing:
The icon and file name are both taken from
https://vale.sh/docs/vale-ini.

**Changes proposed:**

- [x] Add


robertohuertasm (approved):
LGTM 🚀

PR #4075, opened 2026-06-01
feat(icon): add gitignore language ID to git

NaufalK25:
_**Fixes #N/A**_

**Changes proposed:**

- [x] Add

Add support for `gitignore` language ID based from this VS Code extension: https://marketplace.visualstudio.com/items?itemName=codezombiech.gitignore

image


remcohaszing (approved):
This change is fine, but I think it would be even better for that extension to remove the `gitignore` language. It’s seems pointless. VSCode has the `ignore` language ID builtin. I suspect the extension might predate the `ignore` language ID.

NaufalK25:
Yeah, you're right. It probably exists before VS Code's builtin `ignore` language ID.

PR #4083, opened 2026-06-04
feat(icon): add metro icon

NaufalK25:
_**Fixes #N/A**_

**Changes proposed:**

- [x] Add

Add support for [Metro](https://metrobundler.dev/) config file. Here are the [docs](https://metrobundler.dev/docs/configuration/) and [filename reference](https://github.com/facebook/metro/blob/main/packages/metro-config/src/loadConfig.js#L51-L52). They also used to support [.yaml and .yml](https://github.com/facebook/metro/blob/main/packages/metro-config/src/loadConfig.js#L65) but flagged as deprecated, should I also add that?
The icon is taken from [here](https://github.com/facebook/metro/tree/main/website/static/img).
image


src/iconsManifest/supportedExtensions.ts
```diff
@@ -3604,6 +3604,14 @@ export const extensions: IFileCollection = {
       languages: [languages.metal],
       format: FileFormat.svg,
     },
+    {
+      icon: 'metro',
+      filenamesGlob: ['metro.config'],
```
remcohaszing: It looks like `.config/metro.[ext]` is also supported. So we should add `'metro'` here.

remcohaszing (changes requested):
Thanks!

It looks like Metro supports loading config files with the `.ejs` / `.yaml` / `.yml` extensions, but it doesn’t load a file named `metro.config.ejs` etc by default. So it’s not a special case that we need to support.

src/iconsManifest/supportedExtensions.ts
NaufalK25: Done

remcohaszing (approved):

Issue #2180, opened 2019-07-30
[Icon Request] Add (meson) wrap


ZanderBrown:
- [X] I'm sure this issue is not a _duplicate_.
- [X] I'm submitting an icon request:
  - Type: `filename`
  - Icon Name: `meson`/`meson-wrap`
  - Sample original Icon: ![logo]()
  - Extensions: `subprojects/*.wrap`
  - Language ids: ``

- Language Extensions: https://marketplace.visualstudio.com/items?itemName=asabil.meson
- More info: http://mesonbuild.com/Wrap-dependency-system-manual.html

Should probably just be an alias for meson

JimiC:
Not sure that the `wrap` file extension is `Meson` specific.

ZanderBrown:
Possibly not but for meson they are always located in a `subprojects` folder so perhaps that's specific enough?

JimiC:
We cannot bind the parent folder name. `vscode` has no such capability.

ZanderBrown:
That's unfortunate

A quick search on the various "extension info" websites only seems to pull up some obsure imaging software for `*.wrap`

PR #4085, opened 2026-06-05
feat(icon): add meson options to meson

NaufalK25:
_**Fixes #N/A**_

**Changes proposed:**

- [x] Add

Add support for `meson.options` and `meson_options.txt`, a build options file for [Meson](https://mesonbuild.com/). Here is the filename reference: https://mesonbuild.com/Build-options.html

I also wanted to add the `.wrap` file extension related to #2180, and I found the [reference](https://mesonbuild.com/Wrap-dependency-system-manual.html) in the Meson Wrap dependency documentation. However, after reading the discussion there, I feel that I shouldn't.

### DIFF
--- a/src/iconsManifest/supportedExtensions.ts
+++ b/src/iconsManifest/supportedExtensions.ts
@@ -3599,6 +3599,12 @@ export const extensions: IFileCollection = {
       format: FileFormat.svg,
     },
     {
+      icon: 'meson',
+      extensions: ['meson.options', 'meson_options.txt'],
+      filename: true,
+      format: FileFormat.svg,
+    },
+    {
       icon: 'metal',
       extensions: [],
       languages: [languages.metal],


remcohaszing (changes requested):
The meson icon is already used for the `meson` language ID, which is also assigned to `meson.options` and `meson_options.txt` in https://github.com/mesonbuild/vscode-meson/blob/main/package.json#L480.

I’m inclined to reject this, but I will leave the final judgement to @robertohuertasm.

NaufalK25:
Thanks for the clarification.

That makes sense. I saw that `meson.options` and `meson_options.txt` are associated with the existing Meson language ID in the VS Code Meson extension, but I wasn't aware that the Meson icon was already being applied through that mapping.

In that case, I'm fine with closing this PR if the files are already covered. Thanks for taking a look.

PR #4086, opened 2026-06-06
feat(icon): add powerbi icon

### LINKED ISSUE #3306: [Icon Request]: Power BI as ".pbix"
### Housekeeping

### Icon Type

- [X] File Icon
- [ ] Folder Icon

### Icon Type

- [X] Extension
- [ ] Filename
- [ ] Language ID

### Original Icon

Power-BI

### Icon Information

According to the Microsoft github acct this is the current Power BI icon file. See https://github.com/microsoft/PowerBI-Icons

### More Information:

_No response_

NaufalK25:
_**Fixes #3306**_

**Changes proposed:**

- [x] Add

Add [Power BI](https://www.microsoft.com/en-us/power-platform/products/power-bi/) support through this VS Code extension: https://marketplace.visualstudio.com/items?itemName=GerhardBrueckl.powerbi-vscode. The icon is taken from [here](https://cdn-dynmedia-1.microsoft.com/is/content/microsoftcorp/1068058-Icon-PowerBI?resMode=sharp2&op_usm=1.5,0.65,15,0&wid=96&hei=96&qlt=100&fit=constrain).


remcohaszing (commented):
If I look at the linked VSCode extension, it already provides icons for these language IDs. I do think out icon looks better, but that VSCode extension uses different icons for these language IDs.

I’m not saying providing this icon is a bad idea, but I do have my doubts.

NaufalK25:
I think it's okay to add. There are already some icons in the project that differ from what the language VS Code extension provides. One example I know of is Mermaid, where we already provide our own version rather than using the one they provide, and there might be more examples as well.

That said, I can see the concern about consistency with the linked VS Code extension. I'm fine either way and don't feel strongly about it, so I'll leave the final decision to you.

robertohuertasm:
One of the benefits of adding an icon on our side is that we can support this for Zed users. I'd be ok adding it.

remcohaszing:
My concern isn’t really the icon itself or the override. The icon looks great. My concern is that we now use one icon to represent two file types that have distinct icons in the original VSCode extension.

Of course one strategy would be to merge and see if users complain.

NaufalK25:
I see, what if I remove the `.pbinb`/`powerbi-api` related code and only keep `.pbix`/`powerbi-desktop`? That would line up with the original issue scope and avoid the one-icon-for-two-distinct-types overlap you flagged. Happy to push that update if it sounds good.

robertohuertasm:
@NaufalK25 leave the `knownExtensions` as they are, as this can be used for Zed and doesn't affect at all the VS Code version of the extension. Then just remove the `power-api` id. @remcohaszing would you be ok with that?

remcohaszing:
I’m just in doubt. Based on my observations, I’m not sure that the proper approach is. I’m not against the change as-is or with some specific language IDs removed. I’m not familiar with the Power BI ecosystem at all.

I’m leaning more and more towards merging this as-is and seeing if anyone complains.

robertohuertasm:
Ok, so let's keep it simple and let's merge this.

robertohuertasm (approved):

PR #4056, opened 2026-05-18
feat(icon): add postgresql icon

NaufalK25:
_**Fixes #3037**_

**Changes proposed:**

- [x] Add

Add support for [PostgreSQL](https://www.postgresql.org/) by adding `.pgsql`, `.psql`, and `.sql` file extension and `postgres` language ID based on this VS Code extension: https://marketplace.visualstudio.com/items?itemName=ckolkman.vscode-postgres

image

Issue #3037, opened 2022-06-01
[Icon Request]: PostgresQL doesn't have a file icon.


y-okamoto-1113:
### Housekeeping

- [X] I'm sure this issue is _not_ a duplicate

### Icon Type

- [X] File Icon
- [ ] Folder Icon

### Icon Type

- [X] Extension
- [ ] Filename
- [ ] Language ID

### Original Icon

i don't know which icon is original.
maybe, this is original 🤔
https://upload.wikimedia.org/wikipedia/commons/2/29/Postgresql_elephant.svg

### Icon Information

svg

### More Information:

Hi, I noticed that when we choose the PostgresQL mode, there is no more icon displayed.
# if i choose `sql`, there is a database icon displayed.
![スクリーンショット 2022-06-01 16 55 49](https://user-images.githubusercontent.com/40970030/171355910-baee77ea-defb-4f8e-a64a-94b85947f4a2.png)

# if i choose `postgresql`, there is no icon displayed.
![スクリーンショット 2022-06-01 16 55 18](https://user-images.githubusercontent.com/40970030/171355891-04326a60-cdf3-44af-983c-af72b000657d.png)

sasial-dev:
Would you be happy with the postgresql language having the database icon?

y-okamoto-1113:
@sasial-dev 
yes, please.

sasial-dev:
Alright. I'll use the postgres icon though, as to my knowledge only postgres uses the extension

y-okamoto-1113:
@sasial-dev 
thank you so much! 😄

sasial-dev:
Can you send a link to which postgresql VSC extension you are using?

Slowhigh:
Still in the progress?

PR #4093, opened 2026-06-11
fix(icon): remove duplicate postgresql icon

NaufalK25:
***Re #3037***

**Changes proposed:**

* [x] Fix

While looking for icons that contain more than one line of code, I found that [file_type_pgsql.svg](https://github.com/vscode-icons/vscode-icons/blob/master/icons/file_type_pgsql.svg) already exists. I apologize for overlooking this and adding a duplicate [file_type_postgres.svg](https://github.com/vscode-icons/vscode-icons/blob/master/icons/file_type_postgres.svg) icon (#4056). Instead, the correct approach is to add the `postgres` language ID to the existing `language.ts` entry.


remcohaszing (approved):
Oh oops! Thanks for correcting this!

PR #4096, opened 2026-06-11
chore(typescript): update to TypeScript 6

remcohaszing:
TypeScript 6 no longer supports the `module` option `commonjs`. It was
replaced with `node16` (or higher). The correct `moduleResolution` is
implied.

`strict` is now enabled by default, but we didn’t use that. So I
explicitly disabled `noImplicitAny`, `strictFunctionTypes`, and
`strictNullChecks`. It might be worthwhile to remove these options, but
that that can be done separately.

Some slight code changes were made to adhere to the new options.


KingDarBoja:
@remcohaszing Pushed a small fix that should mark all tests as green 🚀

KingDarBoja (approved):

PR #4097, opened 2026-06-11
chore(npm): disallow lifecycle scripts

remcohaszing:
Starting with npm 11, you can disallow dependencies from running lifecycle scripts. As of npm 12, the default behaviour is to fail installation if a script wasn’t configured explicitly.

This change disables all lifecycle scripts, but tells npm it’s ok to continue installation.


robertohuertasm (approved):
LGTM 🚀

PR #4100, opened 2026-06-12
feat(icon): add .pdbrc icon

remcohaszing:
Pdb is the builtin Python debugger. It reads input from the `.pdbrc` file. https://docs.python.org/3/library/pdb.html.

Pdb doesn’t have its own icon. So I duplicated the Python icon, and copied the cogwheel from the tsconfig icon into it.

It could be interesting to purpose this icon for a couple more files. For example, Sphinx uses `conf.py` and Flask uses `config.py`. These are just Python files with a pretty generic name, so I don’t think it’s appropriate to use the Sphinx or Flask logos for these. But the `pythonconfig` icon is subtle.

**Changes proposed:**

- [x] Add


robertohuertasm (dismissed):
LGTM 🚀

remcohaszing:
@robertohuertasm  Do you also have thoughts on this?

> It could be interesting to purpose this icon for a couple more files. For example, Sphinx uses `conf.py` and Flask uses `config.py`. These are just Python files with a pretty generic name, so I don’t think it’s appropriate to use the Sphinx or Flask logos for these. But the `pythonconfig` icon is subtle.

robertohuertasm:
@remcohaszing I'm not a Python dev myself. If you think it makes sense, I'll second your opinion here.

remcohaszing:
I added a couple of well-known filename patterns. Feel free to merge.

robertohuertasm (approved):
LGTM 🚀

PR #4104, opened 2026-06-15
feat(icon): add databricks icons

### LINKED ISSUE #4094: [Icon Request]: Databricks
### Housekeeping

### Icon Type

- [x] File Icon
- [x] Folder Icon

### Icon Type

- [ ] Extension
- [x] Filename
- [ ] Language ID

### Original Icon

Image

### Icon Information

The Databricks extension for VS Code generates a `.databricks` directory for storing configuration settings.Declarative Automation Bundles (Databricks Asset Bundles) rely on a `databricks.yml` file, which could also utilize the icon.

### More Information:

[Databricks extension for Visual Studio Code](https://docs.databricks.com/aws/en/dev-tools/vscode-ext/)
[Declarative Automation Bundles configuration](https://docs.databricks.com/aws/en/dev-tools/bundles/settings)

DST-Kasanicky:
- Add file icon for databricks.yml / databricks.yaml (Databricks Asset Bundle configuration files)
- Add folder icon for databricks / .databricks (Databricks VS Code extension configuration directory)

**Fixes #4094**

**Changes proposed:**

- [x] Add


src/iconsManifest/supportedExtensions.ts
```diff
@@ -1587,6 +1587,12 @@ export const extensions: IFileCollection = {
       languages: [],
       format: FileFormat.svg,
     },
+    {
+      icon: 'databricks',
+      extensions: ['databricks.yml', 'databricks.yaml'],
```
remcohaszing, line 1592: I only see a reference to `databricks.yml`, not `databricks.yaml`. Are you sure that file is also supported?

```suggestion
      extensions: ['databricks.yml'],
```

src/iconsManifest/supportedFolders.ts
```diff
@@ -212,6 +212,11 @@ export const extensions: IFolderCollection = {
       extensions: ['dart', '.dart_tool'],
       format: FileFormat.svg,
     },
+    {
+      icon: 'databricks',
+      extensions: ['databricks', '.databricks'],
```
remcohaszing: In the issue you stated that the `.databricks` directory has a special meaning, but you didn’t mention `databricks`. Should that be supported?

```suggestion
      extensions: ['.databricks'],
```

remcohaszing (changes requested):
Thanks! This looks great! I just have a couple of questions.

src/iconsManifest/supportedExtensions.ts
DST-Kasanicky: The doc only mentions yml, but the plugin recognizes both endings since they are equivalent.

src/iconsManifest/supportedFolders.ts
DST-Kasanicky: yeah probably unnecessary

src/iconsManifest/supportedExtensions.ts
remcohaszing: Is this based on an assumption or did you try it? This might be true for Databricks, but it’s not true for all YAML files in general.

DST-Kasanicky: For the Databricks plugin, I tested it and both work.

Historically, .yml and .yaml should always be the same.

If we want to stick strictly to the documentation, yml is fine for me too, since it will cover most cases.

remcohaszing: If it works it’s great! I just don’t know, so I wanted to double check. We have a lot of icons for a great variation of tools, so we’re not familiar with every related tool we support. :)

remcohaszing (approved):

PR #4105, opened 2026-06-16
feat(icon): add knex icon

### LINKED ISSUE #3966: [Icon Request]: knex
### Housekeeping

### Icon Type

- [x] File Icon
- [ ] Folder Icon

### Icon Type

- [ ] Extension
- [x] Filename
- [ ] Language ID

### Original Icon

Image

### Icon Information

As mentioned in https://knexjs.org/guide/migrations.html#knexfile-js - `knexfile.js` is the convention (and the default for their CLI to look for) for knex configs. TS extensions also work, so I think we should support:

`knexfile.[js|cjs|mjs|ts|cts|mts]`

### More Information:

_No response_

NaufalK25:
_**Fixes #3966**_

**Changes proposed:**

- [x] Add

Add support for the [Knex](https://knexjs.org/) configuration file, `knexfile`.

The supported file extensions are based on the stub files generated when running `knex init -x [js|ts|coffee|ls|eg]` in the terminal. These extensions are derived from the corresponding [`knexfile-*.stub`](https://github.com/knex/knex/tree/master/lib/migrations/migrate/stub) files.

I couldn't find any references for `cjs`, `mjs`, `cts`, or `mts`, so I did not include them.

The icon is taken from [here](https://github.com/knex/knex/tree/master/docs/src/public) and optimized.


src/iconsManifest/supportedExtensions.ts
```diff
@@ -3173,6 +3173,14 @@ export const extensions: IFileCollection = {
       languages: [languages.kivy],
       format: FileFormat.svg,
     },
+    {
+      icon: 'knex',
+      extensions: [],
+      filenamesGlob: ['knexfile'],
+      extensionsGlob: ['js', 'ts', 'coffee', 'ls', 'eg'],
```
remcohaszing, line 3180: I believe this is the list of supported extensions: https://github.com/knex/knex/blob/3.2.10/lib/migrations/common/MigrationsLoader.js#L3-L11

NaufalK25: I thought it was for their migration file and not for knexfile, I also found this: https://github.com/knex/knex/blob/master/bin/cli.js#L101 that include `mjs`  but throw error when ran with `npx knex init -x mjs` because the `knexfile-mjs.stub` doesn't exists.
```bash
$ npx knex init -x mjs
Error: ENOENT: no such file or directory, open ...\node_modules\knex\lib\migrations\migrate\stub\knexfile-mjs.stub'
```
I also found this one: https://github.com/knex/knex/issues/4819 that tell about the use case of other extension such as `cjs`.

remcohaszing: Ok, we can always add more extensions later if users request it.

remcohaszing (approved):

PR #4114, opened 2026-06-24
feat(icon): add Doxyfile icon

### LINKED ISSUE #3247: [Icon Request]: Support for Doxyfile
### Housekeeping

### Icon Type

- [X] File Icon
- [ ] Folder Icon

### Icon Type

- [ ] Extension
- [X] Filename
- [ ] Language ID

### Original Icon

The original icon is the default icon

### Icon Information

Need some icon for Doxyfile, because there is no one icon for this files

### More Information:

_No response_

NaufalK25:
_**Fixes #3247**_

**Changes proposed:**

- [x] Add

Add support for [`Doxyfile`](https://www.doxygen.nl/manual/config.html), a doxygen configuration file.
image


src/iconsManifest/supportedExtensions.ts
```diff
@@ -1821,8 +1821,9 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'doxygen',
-      extensions: [],
+      extensions: ['Doxyfile'],
```
remcohaszing, line 1824: There’s also the `doxyfile` language ID registered by https://marketplace.visualstudio.com/items?itemName=samubarb.vscode-doxyfile.

In cases like this, when there’s a well-known special file name and a lesser popular extension, IMO we can register both the file name and the language ID.

src/iconsManifest/supportedExtensions.ts
```diff
@@ -1821,8 +1821,9 @@ export const extensions: IFileCollection = {
     },
     {
       icon: 'doxygen',
```
remcohaszing: If we have a special file name like this to configure a language, I often copy the language icon and copy the cogwheel from the tsconfig icon inside it. I think it’s nice to have a different icon for the language and its configuration.

remcohaszing (changes requested):

src/iconsManifest/supportedExtensions.ts
NaufalK25: I see, I wasn't aware that the Doxyfile has its own language ID.

src/iconsManifest/supportedExtensions.ts
NaufalK25: Okay, I'll give that a try.

remcohaszing (approved):

PR #4124, opened 2026-06-26
feat(icon): add dedicated txt icon

### LINKED ISSUE #3849: [Icon Request]: .txt
### Housekeeping

### Icon Type

- [x] File Icon
- [ ] Folder Icon

### Icon Type

- [x] Extension
- [ ] Filename
- [ ] Language ID

### Original Icon

https://github.com/vscode-icons/vscode-icons/blob/master/icons/default_file.svg

### Icon Information

Large blue "TXT" could be nice

### More Information:

_No response_

Nut2026:
_**Fixes #3849**_

**Changes proposed:**
- [x] Add
- [x] Fix

- Add a dedicated icon for `.txt` files as suggested in #3849


Nut2026:
Hmm... The icon contains the word `TXT` - the Background Test didn't display it, but it was visible in the Development Host during testing. I used Inkscape's Calibri font.

remcohaszing (changes requested):
You shouldn’t use Calibri, or any other Microsoft fonts, since they are proprietary. In many cases they are unavailable. But the same can be said for any text in SVG. It may render differently based on the environment.

But also, we already have an icon for plaintext. I prefer the old icon TBH.

Nut2026:
Thanks for taking the time to explain the font reasoning - I appreciate it. Reflecting, I agree that a new text icon would perhaps be redundant, considering the plaintext icon already serves the purpose well. 

Closing this PR now.

PR #4131, opened 2026-06-28
feat(icon): add Rstack icons

### LINKED ISSUE #3883: [Icon Request]: Rstack (rspack, rsbuild etc.)
### Housekeeping

### Icon Type

- [x] File Icon
- [ ] Folder Icon

### Icon Type

- [ ] Extension
- [x] Filename
- [ ] Language ID

### Original Icon

- https://assets.rspack.rs/rspack/rspack-logo.svg
- https://assets.rspack.rs/rsbuild/rsbuild-logo.svg
- https://assets.rspack.rs/rspress/rspress-logo.svg
- https://assets.rspack.rs/rslib/rslib-logo.svg
- https://assets.rspack.rs/rstest/rstest-logo.svg
- https://assets.rspack.rs/rslint/rslint-logo.svg

### Icon Information

See https://github.com/web-infra-dev

Most of the tooling here has their own config files - including:

- `rspack.config.js|cjs|mjs|ts|cts|mts`
- `rsbuild.config.js|cjs|mjs|ts|cts|mts`
- `rspress.config.js|cjs|mjs|ts|cts|mts`
- `rslib.config.js|cjs|mjs|ts|cts|mts`
- `rstest.config.js|cjs|mjs|ts|cts|mts`
- `rslint.json`

They have unique icons, but if we don't want them all added, I could see the rspack one being used for all.

I could also see just adding the logos for the popular/stable tools for now - namely `rspack` and `rsbuild`.

### More Information:

_No response_

amirsabanovic:
_**Fixes #3883**_

**Changes proposed:**

- [X] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

Per suggestions from maintainers, adding only the `rspack` icon at this point.


remcohaszing (changes requested):
Thanks for trying to tackle this issue! I know it’s a tricky one.

We aim to keep our icons under 2kB, but we can make exceptions if there’s no other way. Still, 18kB is huge. We must reduce the file size somewhat.

I suggest to start with just the `rspack` icon, as it’s the simplest and most neutral one. I would even argue it’s fine to use this for all file names suggested in #3883 for a first iteration.

amirsabanovic:
I’ve been able to remove a set of stroke-width paths that have a negligible impact on the icon’s appearance, bringing the file size down to 11 kB. Beyond this, I think I would have to simplify some of the more complicated paths and see if that shaves a kilobyte or two off.

remcohaszing (dismissed):
These icons are still too big according to our guidelines, but this is definitely getting closer to at least me considering we can make an exception here. The icons are simply large and hard to scale down further.

I’m ok with this contribution, but I will leave final judgement up to @robertohuertasm

KingDarBoja:
Hmmm seems like I might need to fix it just a little bit but managed to reduce the file size.

joealden:
@chenjiahan are the icons you previewed in https://bsky.app/profile/chenjiahan.bsky.social/post/3mtginh4rcc2h available as SVGs anywhere? As I'm wondering if those versions could get the sizes more reasonable here (and they should be more legible as file icons anyway).

chenjiahan:
Yes! The SVGs are available in [this PR](https://github.com/rstackjs/rstack-design-resources/pull/91).

I’d recommend the [claw logo](https://github.com/rstackjs/rstack-design-resources/blob/main/rspack/rspack-claw-logo.svg). It has a simple design and a smal size, so it should work well as a file icon.

SoonIter:
Thanks for the work on this! We can continue in #4252, where the icon is now 2.2 KB.

PR #4133, opened 2026-06-29
fix(icon): add missing ts, cts, and mts support for prettier

NaufalK25:
_**Fixes part of #3991**_

**Changes proposed:**

- [x] Fix

Add missing `ts`, `cts`, and `mts` support for [Prettier](https://prettier.io/). 
See https://prettier.io/docs/configuration for more details.


remcohaszing (approved):
Thanks!

PR #4136, opened 2026-06-30
feat(icon): add WGSL language ID to wgsl

### LINKED ISSUE #3322: My vscode-icon cannot display the icon for wgsl
### Housekeeping

### Information:

vscode version: 1.80.1 
vscode version 12.4.0

system version: macos 13.4.1 

As per the question, my vscode cannot display icons with the suffix `*. wgsl`

image
image
image

NaufalK25:
_**Fixes #3322**_

**Changes proposed:**

- [x] Add

Add support for the WGSL language ID provided by the VS Code extension: https://marketplace.visualstudio.com/items?itemName=ybwork.ybwork-wgsl.

I checked all WGSL language support extensions available on the VS Code Marketplace and found that `WGSL` (uppercase) is the only language ID currently missing. I believe this may also be related to #3322.

The only other related language ID I found is `wgsl_template`, provided by https://marketplace.visualstudio.com/items?itemName=fs-eire.wgsl-template. Since it is specifically used for `.wgsl.template` files rather than `.wgsl` files, I don't think it should be included here.

### DIFF
--- a/src/iconsManifest/languages.ts
+++ b/src/iconsManifest/languages.ts
@@ -701,7 +701,7 @@ export const languages = {
   wasp: { ids: 'wasp', knownExtensions: ['wasp'] },
   wenyan: { ids: 'wenyan', knownExtensions: ['wy'] },
   wesl: { ids: 'wesl', knownExtensions: ['wesl'] },
-  wgsl: { ids: 'wgsl', knownExtensions: ['wgsl'] },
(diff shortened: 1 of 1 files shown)


remcohaszing (approved):
Thanks!

Bonus points if you can get that VSCode extension to lower case their language ID. 😉

Regardless, this is good to have.

PR #4139, opened 2026-07-01
fix(icon): add missing .js, .mjs, and .mts support for alchemy

NaufalK25:
_**Fixes #N/A**_

**Changes proposed:**

- [x] Fix

Add missing .js, .mjs, and .mts support for [Alchemy](https://alchemy.run/). 
See: https://github.com/alchemy-run/alchemy/blob/main/alchemy/bin/services/execute-alchemy.ts#L209-L214 for filename reference.


remcohaszing (approved):
Thanks!

PR #4152, opened 2026-07-02
feat(icon): add the ocx icon

michael-herwig:
This uses a new `ocx` icon for files named `ocx.toml` (project manifest) and `ocx.lock` (lockfile), mirroring how the `cargo` icon covers `Cargo.toml` and `Cargo.lock`.

The file names come from ocx, a package manager that distributes prebuilt binaries through OCI registries: https://ocx.sh

**Changes proposed:**

- [x] Add


remcohaszing (approved):
Thanks!

PR #4156, opened 2026-07-05
feat(icon): add ejs language ID

NaufalK25:
_**Fixes #N/A**_

**Changes proposed:**

- [x] Add

Add `ejs` language ID provided by https://marketplace.visualstudio.com/items?itemName=DigitalBrainstem.javascript-ejs-support.

The purpose of this addition is not to display an icon in the file tree, but to show the appropriate icon in the Language Mode picker.

| Before | After |
|--------|-------|
| image |  image |

I also noticed that VS Code natively recognizes `.ejs` files as `html`, as shown here: https://github.com/microsoft/vscode/blob/main/extensions/html/package.json#L31. Therefore, I didn't remove the `extension` from `supportedExtension.ts`, since it is still recognized and handled as `html` by default.


remcohaszing (approved):
Thanks!

PR #4157, opened 2026-07-05
feat(icon): add telegram folder icons

Nut2026:
_**Fixes #N/A**_

**Changes proposed:**
- [x] Add

Add folder support for the messaging platform [Telegram](https://telegram.org/). 

- The icon was taken from [here](https://telegram.org/tour/screenshots). 
- It would be beneficial for visual identification of folders dedicated to Telegram workflows such as bot development.


src/iconsManifest/supportedFolders.ts
```diff
@@ -777,6 +777,11 @@ export const extensions: IFolderCollection = {
       format: FileFormat.svg,
     },
     { icon: 'tauri', extensions: ['src-tauri'], format: FileFormat.svg },
+    {
+      icon: 'telegram',
+      extensions: ['telegram', '.telegram'],
```
remcohaszing: I noticed I also have the `~/.local/share/Telegram` and `~/.local/share/TelegramDesktop` folders. Could you add those too?

The file names are case insensitive for VSCode, but they are case sensitive for Zed.

remcohaszing (dismissed):
Thanks! I just have one nitpicky request.

src/iconsManifest/supportedFolders.ts
Nut2026: Got it! I've added `Telegram` and `TelegramDesktop` as well.

remcohaszing (approved):

PR #4158, opened 2026-07-05
feat(icon): support translation and translations

Nut2026:
_**Fixes #N/A**_

**Changes proposed:**
- [x] Add

- Add `translation` and `translations` to the existing `locale` folder icons mapping.


remcohaszing (approved):
Thanks! IMO this makes sense, but I’ll leave final judgement up to @robertohuertasm.

PR #4166, opened 2026-07-08
fix(icon): support mts to next config

NaufalK25:
_**Fixes part of #3991**_

**Changes proposed:**

- [x] Fix

Add `mts` support to [Next](https://nextjs.org/) config. Here are the reference: https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/constants.ts#L109-L116.

I couldn't find the reference for `cts`, and in the reference I gave, `cjs` is not available either. Should I remove it?


remcohaszing (approved):
Thanks!

PR #4170, opened 2026-07-09
feat(icon): add behat icon

NaufalK25:
_**Fixes #N/A**_

**Changes proposed:**

- [x] Add

Add support for [Behat](https://docs.behat.org/en/latest/). Filenames reference were taken from [confirguration file page](https://docs.behat.org/en/v3.x/user_guide/configuration.html#behat-php) and the icon was taken from https://github.com/Behat/logo.


remcohaszing (approved):
Thanks!

PR #4176, opened 2026-07-11
feat(icon): add nitro icon

NaufalK25:
_**Fixes #N/A**_

**Changes proposed:**

- [x] Add

Add support for [Nitro](https://nitro.build/) config file. Filenames were taken from https://nitro.build/docs/configuration and icon was taken from https://github.com/nitrojs/nitro/tree/main/docs/.docs/public.


src/iconsManifest/supportedExtensions.ts
```diff
@@ -4119,6 +4119,14 @@ export const extensions: IFileCollection = {
       filename: true,
       format: FileFormat.svg,
     },
+    {
+      icon: 'nitro',
+      extensions: ['.nitrorc'],
+      filenamesGlob: ['nitro.config'],
+      extensionsGlob: ['js', 'ts'],
```
remcohaszing, line 4126: Nitro uses [c12](https://github.com/unjs/c12) to load configuration files. Their documentation states a lot more file names are supported, but it’s not clear to me which filenames are supported exactly.

Even if the list is incomplete, this change is an improvement. So I approve and we should include this in the next release. But I’ll wait a bit with merging, so you can investigate the supported filenames if you’re interested.

remcohaszing (dismissed):

src/iconsManifest/supportedExtensions.ts
NaufalK25: Yeah, I believe so as well. The feature list in the repository suggests that more configuration filenames are supported, but I'm not entirely sure which ones are actually recognized in practice, or whether it simply supports all of them through [c12](https://github.com/unjs/c12).

To be on the safe side, I added `.nitrorc`, `nitro.config.js`, and `nitro.config.ts` since those are explicitly documented. If we later confirm support for additional filenames or extensions, we can add them in a follow-up PR. We've taken a similar incremental approach for other tools as well—for example, Prettier, Prisma, and Next.js, where we added the `.mts` variants recently. That was my reasoning for keeping this change conservative.

remcohaszing: I opened https://github.com/nitrojs/nitro/issues/4505. But my original comment remains:

> Even if the list is incomplete, this change is an improvement. So I approve and we should include this in the next release. But I’ll wait a bit with merging, so you can investigate the supported filenames if you’re interested.

NaufalK25: Thanks for opening the issue and for the heads-up! I'll wait and see if we get a definitive answer there before making any further changes.

remcohaszing (approved):

PR #4180, opened 2026-07-17
Adds additional Playwright filenames

DiefBell:
**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


remcohaszing (approved):
These are not really special filenames for Playwright, but I do like them. It’s in a similar space as `tsconfig.build.json` or `webpack.production.js`.

I’ll leave final judgement up to @robertohuertasm :)

PR #4190, opened 2026-08-04
feat: add direnv icon

NaufalK25:
_**Fixes #N/A**_

**Changes proposed:**
- [x] Add [direnv](https://direnv.net/) config file support named [`.envrc`](https://github.com/direnv/direnv/blob/master/internal/cmd/config.go#L181-L183).

Icon was taken from https://github.com/direnv/direnv-logo.

image

**Note:** `.envrc` is already recognized as a `shellscript` file by VS Code, since [shell script language support ships natively/built-in including envrc filename](https://github.com/microsoft/vscode/blob/main/extensions/shellscript/package.json#L62) rather than through an extension. This means `.envrc` already gets the shell icon ([`file_type_shell`](https://github.com/vscode-icons/vscode-icons/tree/master/icons/file_type_shell.svg)).

I'm honestly not 100% sure whether this icon addition is worth adding, given `.envrc` already resolves to a reasonable shell icon out of the box. Wanted to flag this so maintainers can weigh in on whether a dedicated direnv icon adds enough value (e.g. visually distinguishing config from regular scripts) to justify maintaining a separate icon mapping.


remcohaszing (approved):
> I'm honestly not 100% sure whether this icon addition is worth adding, given `.envrc` already resolves to a reasonable shell icon out of the box. 

I’m not familiar with direnv. I don’t see how this is different to for example how we show a dedicated ESLint icon for `eslint.config.js`, even though we already have a nice JavaScript icon.

NaufalK25:
Ah good point, I didn't think about eslint.config.js example. Yeah that makes sense, having separate icon can help to visually tell it apart even if the fallback icon already works. Thanks for reviewing.

PR #4201, opened 2026-08-10
fix(buf): add missing `filename` property

jstoparczyk:
**Changes proposed:**

- [x] Fix

The extensions for Buf config files represent whole filenames so a `filename: true` property is required but was missing


remcohaszing (approved):
Thanks!

PR #4202, opened 2026-08-10
fix(bruno): add missing extension

jstoparczyk:
**Changes proposed:**

- [x] Fix

Add missing extension for Bruno files and `filename: false` property

### DIFF
--- a/src/iconsManifest/languages.ts
+++ b/src/iconsManifest/languages.ts
@@ -61,7 +61,7 @@ export const languages = {
     ids: 'browserslist',
     knownFilenames: ['browserslist', '.browserslistrc'],
   },
-  bruno: { ids: 'bruno', knownExtensions: ['bru'] },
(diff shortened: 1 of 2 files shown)


jstoparczyk:
Not sure why those copilot reviews are being triggered. I thought I had them disabled.

remcohaszing (changes requested):
Icons are mapped by a language ID if possible.

I see the Bruno extension registers the `bru` language ID, but we apply the icon to the `bruno` language ID. The right fix would be to revert these changes, and instead change the `bruno` language ID to `bru` in `src/iconsManifest/languages.ts`.

jstoparczyk:
Hi. Assuming both commits would be reverted, I noticed that changing the language ID in `src/iconsManifest/languages.ts` alone seems to not be sufficient for the icon to load properly and an entry in `extensions` field in `src/iconsManifest/supportedExtensions.ts` is also required.

PR #4220, opened 2026-08-19
feat(icon): add checkov icon

NaufalK25:
_**Fixes #N/A**_

**Changes proposed:**

- [x] Add

Add [Checkov](https://www.checkov.io/) — a static code analysis tool for infrastructure as code (IaC) and also a software composition analysis (SCA) tool for images and open source packages.

Filename references were taken from [`checkov/common/util/config_utils.py#L8-L9`](https://github.com/bridgecrewio/checkov/blob/main/checkov/common/util/config_utils.py#L8-L9).

Icon was taken directly from their website. The site provides a full SVG of their logo, so I used that instead of the favicon (which is only a PNG), modifying it to remove everything except the 'C'.


remcohaszing (approved):
Thanks!

PR #4233, opened 2026-09-04
Add NXD Language support

GTS-AppliedTech:
_**Fixes #IssueNumber**_
Adds support for the NXD programming language

**Changes proposed:**
Add NXD Language definition.
Add .nxd file extension association.
Add NXD SVG file icon.

- [ x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare


remcohaszing:
Thanks for the contribution.

Unfortunately we get too many icon requests for new languages that haven’t been well-established yet. So we can’t support them all.

I see that you maintain your own VSCode extension for this language. Know that you can specify a language icon in your own language contribution. That will work, even for users who don’t use vscode-icons.

PR #4241, opened 2026-09-12
feat(icon): add Binary Ninja database support

psyche314:
_**Fixes #IssueNumber**_

**Changes proposed:**

- [x] Add
- [ ] Delete
- [ ] Fix
- [ ] Prepare

**Review follow-up:**

- Replaced the preliminary hand-drawn redraw with a high-fidelity vector trace of the original Binary Ninja artwork.
- Removed transparent-edge artifacts before tracing so the red, black, and white shapes remain clean at icon size.
- The icon uses a 32x32 viewBox and a 2px outer margin; no base64-encoded raster data remains in the SVG.

**Sources:**

- Icon source: the Binary Ninja application icon from an installed Binary Ninja distribution; [Vector 35's Binary Ninja branding reference](https://github.com/Vector35/binaryninja-api/blob/master/docs/img/logo.png).
- Conversion reference: [VTracer](https://github.com/visioncortex/vtracer), used to trace the source artwork into SVG paths.
- Extension reference: [Binary Ninja User Guide - Saving Files](https://dev-docs.binary.ninja/guide/index.html#saving-files), which documents analysis databases with the `.bndb` extension.


remcohaszing (changes requested):
Please follow the [icon contributing guidelines](https://github.com/vscode-icons/vscode-icons/wiki/Contributing). Icons must be a 32x32 SVG with a 2px margin. You provided a base64 encoded PNG inside an SVG.

Also please provide the source of the icon and the source on which the file extensions are based.

Also don’t delete PR description templates. You can use an LLM as a tool, but in the end it’s you, a human, who is responsible for all changes and communication.

PR #4243, opened 2026-09-12
feat(icon): support Unity IL2CPP metadata files

psyche314:
## Summary

- Associate Unity IL2CPP's `global-metadata.dat` with the existing ShaderLab icon.
- Match the complete filename so unrelated `.dat` files retain their current associations.

## Verification

- `npm test` — 707 passing
- `npm run lint` — passed


remcohaszing (commented):
Please provide the source of the icon and the source on which the file extensions are based.

Also don’t delete PR description templates. You can use an LLM as a tool, but in the end it’s you, a human, who is responsible for all changes and communication.

PR #4250, opened 2026-09-16
🐛 TS Config now includes types and libs for TS 6.0

KingDarBoja:
- ~~Add DOM library to compiler options.~~
- Include node types as well because the default types for TS 6.0 is an empty array.

This should fix the wiki GH action workflow as the latest TS 6.0 update included a small change that is breaking it -> https://devblogs.microsoft.com/typescript/announcing-typescript-6-0-rc/#types-now-defaults-to-%5B%5D

**Changes proposed:**

- [ ] Add
- [ ] Delete
- [x] Fix
- [ ] Prepare


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "lib": ["es2022"],
+    "lib": ["es2022", "DOM"],
```
remcohaszing: We do indeed need to add the `node` types, but we don’t need the `dom` lib.


```suggestion
    "lib": ["es2022"],
```

remcohaszing (changes requested):

remcohaszing (approved):

